PR #97, opened 2015-01-23
don't bomb if apiKey is false

atogata:


ConradIrwin:
Out of interest, how did you find this?

PR #98, opened 2015-01-23
allow console logging to be disabled

atogata:


ConradIrwin:
Thanks for the great pull request :).

Issue #100, opened 2015-02-02
Error when loading bugsnag through commonjs as a bundled file.


PrototypeAlex:
After finding that uBlock outright blocks bugsnag, we've decided to wrap bugsnag up in our external-libraries file in our browserify setup. Being so, means we don't have a script tag where bugsnag is bound to, and that leads to the following error.

```
bugsnag.js:318
      var attrs = node.attributes;
                      ^
TypeError: Cannot read property 'attributes' of undefined
  at getData (/Users/alex/Development/reactor/client/scripts/vendor/bugsnag.js:318:23)
  at getSetting (/Users/alex/Development/reactor/client/scripts/vendor/bugsnag.js:334:22)
```

PrototypeAlex:
Can't test this easily, not without adding a browserify step in the process to load as a script, or just running a test via node directly. Thoughts?

ConradIrwin:
It would be nice to test for this, but I don't really see how you got into the state where thisScript was null (there would have to be absolutely no script tags on the page). If you could set up a simple example, that'd be awesome.

I'm also not averse to adding a browserify/webpack/w.e. step to running the tests.

Closing this because the bug is fixed, but feel free to send tests :).

PrototypeAlex:
You're totally right. My problem is that it was breaking my tests, but I'd completely forgotten that I run my tests using jsdom, so there is no script tag in my situation.

I suppose it's still useful, as this proves that it can be completely decoupled from the DOM, and run as a standalone js module. Though.... I'm not sure how helpful that is for others, but... it's extremely helpful for us and our test suite.

Thanks for merging :smile:

PR #101, opened 2015-02-02
Add a null check to node so we don't try access attributes on undefined

PrototypeAlex:
This addresses the use case of trying to load bugsnag when you don't load it through a script tag, typically through commmonjs.

Fixes: https://github.com/bugsnag/bugsnag-js/issues/100


ConradIrwin:
Thanks!

PR #103, opened 2015-03-31
Add an option for serialization depth

trodrigues:
At my current workplace we often log errors sent by our API which tend to have a structure that quickly gets cut off by this mechanism.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -244,7 +244,7 @@
   // nested object syntax, `nested[keys]=val`, to support heirachical
   // objects. Similar to jQuery's `$.param` method.
   function serialize(obj, prefix, depth) {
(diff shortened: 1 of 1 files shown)


ConradIrwin:
@trodrigues 5 should be more than enough. The limit is there to ensure the client can make the request (it has to fit in a GET url), in addition to saving them and us bandwidth.

If you really need the more nested information, and you're sure it won't cause problems with reporting, you can just use a local copy of the library with this number changed, but I don't want to make it too easy for people to meddle without thinking about it.

Thanks,
Conrad

jacobmarshall:
Fair enough. I remember 'needing' this a while ago, but instead opted for flattened data. As long as there is structure, there isn't much difference anyway.

PR #107, opened 2015-04-15
Fix bugsnag not notifying when notifyReleaseStages is set but releaseSta...

jfmengels:
Fix bugsnag not notifying when notifyReleaseStages is set but releaseStage isn't.

As is written here https://bugsnag.com/docs/notifiers/js under releaseStage, by default, the releaseStage is "production". It is true that when notifying, if releaseStage is not specified, the server will consider it as "production".

The problem I noticed is, when notifyReleaseStages is set but releaseStage isn't, then it won't notify at all. That is contrary to what you'd expect when reading the documentation.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -362,7 +362,7 @@
     eventsRemaining -= 1;
 
     // Check if we should notify for this release stage.
-    var releaseStage = getSetting("releaseStage");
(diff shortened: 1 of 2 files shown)


jessicard:
@jfmengels Great catch! Thanks for taking care of this, it totally makes sense. I'll go ahead and merge this in.

PR #110, opened 2015-06-12
Fixes an issue where null exceptions throws an error.

rickharrison:
You may consider this a userland issue, but in an extreme edge case, I was accidentally calling `notifyException` with an `err` that I thought existed, but it actually did not. This caused my client's code execution to stop with nothing reported to bugsnag. Adding this sanity check would fix my issue.


jessicard:
Hey @rickharrison - thanks for the PR! Appreciate not blowing up when an `err` doesn't exist. :)

Issue #94, opened 2014-12-08
Chrome extension


H2go:
Hello,

I'm using bugsnag-js for monitoring a chrome extension. Errors are sent, but never displayed in Bugsnag It seems that there is a problem because of 'chrome-extension" into stacks error.

Exemple :
```
Error
    at generateStacktrace (chrome-extension://nkleeghhjabalfolcphldebmfabofbia/lib/bugsnag.js:489:13)
    at Object.self.notify (chrome-extension://nkleeghhjabalfolcphldebmfabofbia/lib/bugsnag.js:100:19)
    at _init (chrome-extension://nkleeghhjabalfolcphldebmfabofbia/js/content-script/smartbar.js:442:17)
    at Object.callback (chrome-extension://nkleeghhjabalfolcphldebmfabofbia/js/content-script/smartbar.js:322:13)
    at safeCallbackApply (extensions::sendRequest:21:15)
    at handleResponse (extensions::sendRequest:73:7)
```

As workaround, I do this in bugsnag.js :
```
function stacktraceFromException(exception) {
    return (exception.stack || exception.backtrace || exception.stacktrace).replace(/chrome-extension:/g, "chromeextension:");
  }
```

thomchop:
Is this still the unofficially approved way to enable error reporting for a chrome extension?
I have tried removing 'chrome-extension' from the stack trace, url, context, and projectRoot params, but I'm still not seeing any errors in my project dashboard. I've even forwarded the error notification via my server so I can override the referer header. Are there other params that would cause Bugsnag to ignore the notification?

KaoruDev:
+1 for chrome extension support =)

eanakashima:
We added some docs in #129 that should help people get Chrome extension monitoring set up — if you bump into any other issues, let us know.

blimmer:
For those who end up here and are looking for the current docs, they're now here: https://docs.bugsnag.com/platforms/javascript/faq/#how-can-i-get-error-reports-from-browser-extensions

In case it moves again, here's a screenshot:

![Screen Shot 2019-03-21 at 10 18 34](https://user-images.githubusercontent.com/630449/54767422-b8f03780-4bc2-11e9-88a8-911ed7540a70.png)

PR #129, opened 2016-01-27
Document browser extension error handling

eanakashima:
This adds some documentation on how the Bugsnag backend deals with js errors from browser extensions, and gives some suggestions for how to work around it if you have an unusual use case like monitoring an extension.

Refs https://github.com/bugsnag/bugsnag-js/issues/94


snmaynard:
LGTM

kattrali:
:star2: Shippin' it! Thanks, @eanakashima.

PR #131, opened 2016-02-15
Swap JSHint for ESLint analysis

kattrali:
ESLint checking has clearer error reporting, perhaps because it is AST-based. Examples:

* A single starting line being out of alignment is reported correctly, as opposed to reporting all of the following lines as having errors.
* `eqeqeq`/avoiding type coercion is enforced more intelligently around `null`


snmaynard:
LGTM

PR #140, opened 2016-03-24
Add support for `wrapCallbacks`

ConradIrwin:
Hey Bugsnag!

I'd love to be able to re-use the Bugsnag callback wrapping mechanism without re-implementing it in our code-base. This tiny change exposes a `wrapCallbacks` function that lets people like me intercept all callbacks in the same manner that Bugsnag does.

I want to use it to monitor typing lag in the Superhuman email client. Because there are asynchronous jumps in the code while handling key events, it's not sufficient to just see how long handling a keypress event takes, instead we need to take into account all the time in the background spent synching drafts to the database etc, and see whether any of these contribute to delayed keypresses or long ticks.

Would anyone else find this useful?

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -131,7 +131,7 @@
       }
       if (!_super.bugsnag) {
         var currentScript = getCurrentScript();
-        var callback = Bugsnag.wrapCallbacks ? Bugsnag.wrapCallbacks(_super) : _super;
+        var callback = self.wrapCallbacks ? self.wrapCallbacks(_super) : _super;
 
         _super.bugsnag = function (event) {
           if (options && options.eventHandler) {
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -131,18 +131,21 @@
       }
       if (!_super.bugsnag) {
         var currentScript = getCurrentScript();
+        var callback = Bugsnag.wrapCallbacks ? Bugsnag.wrapCallbacks(_super) : _super;
+
         _super.bugsnag = function (event) {
           if (options && options.eventHandler) {
             lastEvent = event;
           }
           lastScript = currentScript;
 
+
           // We set shouldCatch to false on IE < 10 because catching the error ruins the file/line as reported in window.onerror,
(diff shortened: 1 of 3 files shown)


jacobmarshall:
I can't think of a good use for this, other than your use case, however providing the ability to do such a thing can only be a good thing. I like it!

src/bugsnag.js
```diff
@@ -131,18 +131,21 @@
       }
       if (!_super.bugsnag) {
         var currentScript = getCurrentScript();
+        var callback = Bugsnag.wrapCallbacks ? Bugsnag.wrapCallbacks(_super) : _super;
```
jacobmarshall: I think this should be `self.wrapCallbacks` to pass the linter.

ConradIrwin: Thanks!

kattrali:
Hey, @ConradIrwin! Tracking timing is an interesting use case, though I'm not sure this is a general case for an error handling library. Maybe if we did that explicitly instead of generic callback wrapping? I'm giving it some thought.

ConradIrwin:
Thanks @kattrali! The other functionality that this allows is:
1. something like https://www.npmjs.com/package/continuation-local-storage for the browser.
2. tracking async stack-traces (so you can see what the stack was before the async jump), bluebird has an implementation of that for promises.

If you decide it's out of scope for bugsnag js, that's fine :) I'm running our own copy for now at https://npmjs.org/superhuman-bugsnag-js, it would just be nice to not have to remember to pull the changes from this repo all the time.

eanakashima:
Thanks Conrad. It sounds like there is not a great deal of demand for this at the moment, so I think we'll hold off for now. We wouldn't be opposed reconsidering it in the future though, if there's more community interest or we start seeing more common cases where it would be useful.

ConradIrwin:
No worries! I'll keep using my own fork for now :).

Looking forward to v2, I'll let you know how automatic bread-crumbing works for us.

eanakashima:
> I'll let you know how automatic bread-crumbing works for us.

Please do! 👍

PR #143, opened 2016-05-02
Add support for events with DOMHighResTimeStamp

### LINKED ISSUE #130: High-resolution event timestamp breaks eventToMetaData
[Chrome](https://www.chromestatus.com/features/5523910145605632) is switching the Event.timeStamp from [DOMTimeStamp][timestamp-mdn] to a [DOMHighResTimeStamp][highres-timestamp-mdn] in 49th release ([Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1026804) is going to make the switch in near future too). After this change the value is no longer comparable to ```Date.now()``` in chrome which is likely to impact your [```eventToMetaData```](https://github.com/bugsnag/bugsnag-js/blob/master/src/bugsnag.js#L515) function logic for calculating ```millisecondsAgo```.

Note that the current logic is actually broken in Firefox, as it does not use an epoch timeStmap for  most input events.

I am the developer who is making the change in Chrome, and I came across your library as I was doing bulk static analysis. This [blog post](https://developers.google.com/web/updates/2016/01/high-res-timestamps?hl=en) has additional details about the change but I am happy to answer any questions you have as well.

[timestamp-mdn]: https://developer.mozilla.org/en-US/docs/Web/API/DOMTimeStamp
[highres-timestamp-mdn]: https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp

wordofchristian:
closes #130

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -25,7 +25,10 @@
     eventsRemaining = 10,
     // The default depth of attached metadata which is parsed before truncation. It
     // is configurable via the `maxDepth` setting.
-    maxPayloadDepth = 5;
+    maxPayloadDepth = 5,
+
+    // helper function for relative times
+    millisecondsAgo = makeMillisecondsAgo();
 
   // #### Bugsnag.noConflict
   //
@@ -512,7 +515,7 @@
   // Populate the event tab of meta-data.
   function eventToMetaData(event) {
     var tab = {
-      millisecondsAgo: new Date() - event.timeStamp,
+      millisecondsAgo: millisecondsAgo(event.timeStamp),
       type: event.type,
       which: event.which,
       target: targetToString(event.target)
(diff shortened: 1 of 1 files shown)


kattrali:
🎉 Thanks, @wordofchristian.

src/bugsnag.js
```diff
@@ -708,6 +711,29 @@
...
+  //
+  // This is necessary because depending on the browser the event.timeStamp could be a
+  // DOMTimeStamp or a DOMHighResTimeStamp
+  function makeMillisecondsAgo() {
+    function timeNear(a, b) {
+      var d = 1000 * 60 * 5;
+      return a > b - d && a < b + d;
+    }
+
+    var testTimeStamp = (document.createEvent("CustomEvent")).timeStamp;
```
Krystosterone: Doesn't this break compatibility for older browsers? (`document.createEvent`)

wordofchristian: You are correct. That's my fault. #146 should fix it though. How does that look to you?

PR #149, opened 2016-07-07
Breadcrumbs

wordofchristian:
The UI isn't yet ready on Bugsnag.com but here is a screenshot:

breadcrumbs1

##  Bugsnag.leaveBreadcrumb(value, [metaData])

Add a breadcrumb to the array of breadcrumbs to be sent to Bugsnag when the next exception occurs

```javascript
Bugsnag.leaveBreadcrumb("Increased Volume")
```

Each breadcrumb can also have additional diagnostic metadata added.

```javascript
Bugsnag.leaveBreadcrumb('Increased Volume', {
  from: 5,
  to: 11
})
```

The last 25 breadcrumbs get stored and sent along with all payloads to bugsnag.

## Automatic breadcrumbs

Includes automatic creation of breadcrumbs for certain events.

- **Clicks**
- **Errors**
- **Console**
  - log
  - warn
  - error
- **Navigation**
  - Page load
  - DOMContentLoaded
  - Page hide
  - Page show
  - popstate
  - history.pushState
  - history.replaceState

## New configuration options

To disable all automatic breadcrumbs.

```javascript
Bugsnag.autoBreadcrumbs = false;
```

You can also disable any individual automatic breadcrumb groups:

```javascript
Bugsnag.autoBreadcrumbsClicks = false;
Bugsnag.autoBreadcrumbsErrors = false;
Bugsnag.autoBreadcrumbsConsole = false;
Bugsnag.autoBreadcrumbsNavigation = false;
```

If you only want to enable a single type of automatic breadcrumb you can disable all of them and then
enable the individual group you want:

```javascript
// disable all automatic breadcrumbs except click events
Bugsnag.autoBreadcrumbs = false;
Bugsnag.autoBreadcrumbsClicks = true;
```

## TODO

- [x] Add typescript header for leaveBreadcrumb
- [x] Address remaining inline TODO comments in the diff


src/bugsnag.js
```diff
@@ -175,6 +225,181 @@
     }
   }
 
+  // Setup breadcrumbs for click events
+  function trackClicks() {
+    if(!getBreadcrumbSetting("autoBreadcrumbsClicks", true)) {
+      return;
+    }
+
+    window.addEventListener("click", function(event) {
```
eanakashima: Would we need to do feature detection here to prevent this from throwing an error in older IE versions?

wordofchristian: Does the polyfill further down for `addEventListener` allow this?

eanakashima: I think we only polyfill some minor behavior differences on different versions of `addEventListener`, so if it's not already there in a browser I don't know that we'd want to rely on bugsnag-js to polyfill it. As a third party it seems nicer/safer for us to not add new things to window if they're not there already, so I'd vote for us just skipping any automatic breadcrumbs that rely on events in browsers without addEventListener. (We could consider adding `attachEvent` support if a lot of people request it, I guess?)

wordofchristian: 👍

src/bugsnag.js
```diff
@@ -175,6 +225,181 @@
...
+  }
+
+  // Setup breadcrumbs for history navigation events
+  function trackNavigation() {
+    if(!getBreadcrumbSetting("autoBreadcrumbsNavigation")) {
+      return;
+    }
+
+    // check for browser support
+    if (!history || !history.state || !history.pushState || !window.history.pushState.bind) {
```
eanakashima: If `window.history` is undefined in older browsers, I believe the first part of this will still throw, unless we are doing `var history = window.history` someplace I missed?

snmaynard:
What effect does this have on the compressed and minified js size?

wordofchristian:
Before: 
```
Size:    26810
Ugly:     6922
Gzip:     3072
```
After:
```
Size:    36585
Ugly:    10196
Gzip:     4068

```

apsoto:
Looks useful, but I use bugsnag explicitly because it works on embedded browsers (Opera TV).  Please make sure this does the appropriate checks to prevent problems on non-mainstream browsers.

wordofchristian:
> Looks useful, but I use bugsnag explicitly because it works on embedded browsers (Opera TV). Please make sure this does the appropriate checks to prevent problems on non-mainstream browsers.

@apsoto Our plan is to do browser feature detection around the new features in order to degrade gracefully. The `leaveBreadcrumb()` function itself should have the same browser support that the rest of the library has, but some of the automatic breadcrumbs won't work in older browsers. 

We are also planning on releasing this as a new major version (3.0), so you will be able to test it before deciding to upgrade. Any bug reports about Opera TV specific issues will be greatly appreciated.

src/bugsnag.js
```diff
@@ -252,6 +503,66 @@
     }
   }
 
+  // Compare if two objects are equal.
+  // TODO check if this would fail if the properties are traveresed in different orders
```
jacobmarshall: Answer: Yes, it does fail 😛

src/bugsnag.js
```diff
@@ -504,7 +504,6 @@
   }
 
   // Compare if two objects are equal.
-  // TODO check if this would fail if the properties are traveresed in different orders
   function isEqual(obj1, obj2) {
     serialize(obj1) === serialize(obj2);
```
jacobmarshall: Me thinks there also might be a `return` missing? Unless you're using it for testing right now, in which case I'll get lost and quit bugging you :P

README.md
```diff
@@ -163,6 +198,57 @@ this with:
 Bugsnag.autoNotify = false;
 ```
 
+###autoBreadcrumbs
```
loopj: @kattrali looks like bugsnag-js hasn't been moved to new docs format

src/bugsnag.js
wordofchristian: Doh! good catch

README.md
```diff
@@ -163,6 +186,57 @@ this with:
...
+
+By default, we automatically create breadcrumbs for the following types of events:
+
+- **Clicks**
+- **Errors**
```
eanakashima: Maybe it's worth saying "Uncaught errors" for clarity?

README.md
```diff
@@ -163,6 +186,57 @@ this with:
...
+- **Navigation**
+  - Page load
+  - DOMContentLoaded
+  - Page hide
+  - Page show
+  - popstate
+  - history.pushState
+  - history.replaceState
+
+
```
eanakashima: Maybe a `### Disabling breadcrumbs` header or something would be helpful here, just for easy reading?

src/bugsnag.js
```diff
@@ -118,6 +119,61 @@
...
   };
 
+  // #### Bugsnag.leaveBreadcrumb(value, [metaData])
+  //
+  // Add a breadcrumb to the array of breadcrumbs to be sent to Bugsnag when the next exception occurs
+  // - `value` (string|object) If this is an object, it will be used as the entire breadcrumb object
+  //   and any missing `type`, `name` or `timestamp` fields will get default values.
+  //   if this is a string and metaData is provided, then `value` will be used as the `name` of the
+  //   breadcrumb.
+  //   if `value` is a string and is the only argument the breadcrumb will have `custom` type and
```
eanakashima: s/custom/manual/?

kattrali:
🎉

kattrali:
...That is a mesmerizing GIF.

sandstrom:
@wordofchristian Great feature, just implemented rc1 and it works well.

Some thoughts:

1. UI for the breadcrumbs could be improved, i.e. the console output is needlessly verbose now. Instead of using the `severity` and `message` keywords in the UI it should simply use the icon + the message + severity string.

2. Great if we could manually create the built-in things, e.g. console messages. I got it working by looking at the source, but I assume those are private APIs that could change? or are they not?

wordofchristian:
@sandstrom Yes we are planning on documenting those types at some point. We didn't initially so as to keep the API and documentation simpler. You should be able to continue using it the way you have been, and it will continue to work.             

```javascript
Bugsnag.leaveBreadcrumb({
  type: "error|navigation|user|log|manual",
  name: "Anything",
  metaData: {
    anyKey: "any value",
    foo: "bar",
    biz: "baz",
  }
});
```

Issue #117, opened 2015-09-04
Offline support?


machty:
It seems there's no support for enqueuing notifications for later when the app is offline. Is there any plan to add this in the future?

jacobmarshall:
How exactly would that work. And how is that related to JS errors?

machty:
Client-side analytics platforms (one example being Localytics) often have error handling on their uploads, such that if the upload fails, it's saved for a later upload in localStorage (but perhaps just storing in memory is all we need). Then at some point in the future, it tries to send a batch of all prior unsent payloads. Bugsnag could adopt a similar approach.

It seems like an important feature just because, without it, you're losing data on any exceptions thrown if the user has a flaky connection at the time of the exception (mobile apps are particularly prone to this).

jacobmarshall:
I see what you mean. To be honest I'd find it unlikely to be implemented into the library itself, but I like the idea. It would seem that it's something that can be implemented on a per-app basis, based on your own specifications. Having such a system would add additional 'bloat' to the existing js file, for something that would probably be an opt-in feature anyway.

Just my opinion, sorry. But I'd never thought about it before now. Very smart.

gonsfx:
For me, this is actually very important. Do any other Bugtracking Services support this for JavaScript-based Errors?

abrbhat:
This will also enable an easier integration with Cordova/Ionic applications

wordofchristian:
Notable: https://developers.google.com/web/updates/2015/12/background-sync

jacobmarshall:
So I've put together some code for supporting in-memory error event retrying. There are a few of things to note, which need some feedback (cc @wordofchristian @kattrali).

- Right now, the `notify.bugsnag.com/js` endpoint responds with an empty response. Meaning the `onerror` event on the `Image` object (when using the default `notifyHandler`) is fired instead of the `onload` event (because a zero byte image isn't valid).
 - This would need to be changed so we can retry requests for users with flakey connections, where they do have "online capability", but no actual connectivity.
 - XHR doesn't have this issue, as you can extract the response `status` from the XHR object, and determine whether the request was successful or not.
- I still have yet to implement support for IE8, which does have online/offline events & support, but triggers events off the `document.body` rather than `window`.
- There is currently no limit on how many events can be queued in-memory. Do you reckon it's worth adding one? If so, how large/should it be configurable?

jacobmarshall:
[...]

PR #151, opened 2016-07-10
Offline support

jmshal:
#117


eanakashima:
Hey @jacobmarshall, thanks for opening this, really excited about it. There's been a bit of a delay on us merging in this and one or two other large PRs, because it will put us over the [4604 gzip'd bytes of js](https://github.com/bugsnag/bugsnag-js/pull/57#issuecomment-47064821) size discussed in prior issues, so we're thinking a little more about how we want to structure additional features on top of bugsnag-js. We'll report back when we have an update. Sorry for the wait!

jacobmarshall:
Sounds good, thanks for letting me know.

My two cents on the matter - maybe making these kinds of larger additions available as addons (`bugsnag-js/workers` and `bugsnag-js/offline`) for people to optionally include wouldn't be such a bad idea? People bundling their app's dependencies in a seperate vendor file (for instance) probably have a much larger overall file size anyway.

ticky:
@eanakashima is this likely to land anytime soon?

We’ve got some issues we’re trying to pin down specifically around connectivity issues so being able to enqueue events for when the network connection isn’t in trouble would be greatly advantageous! :)

jmshal:
> @eanakashima is this likely to land anytime soon?
> 
> We’ve got some issues we’re trying to pin down specifically around connectivity issues so being able to enqueue events for when the network connection isn’t in trouble would be greatly advantageous! :)

@eanakashima If the answer is *likely not*, I am happy to do some "streamlining" to reduce the bundle size this weekend.

@ticky As an aside, you can always manually patch/pull in a custom version of Bugsnag with these changes - if that's something you're into?

btecu:
@ticky [AirBrake ](https://airbrake.io/) supports offline errors.

eanakashima:
@ticky it's definitely on our radar, but unfortunately I don't think it's likely to land in the next week or two. As @jmshal mentioned, as an alternative, it's safe to replace the cdn-ified version of bugsnag-js with a custom version that you bundle with your own JS files (just be sure to include it first so it can shim the right things and report the greatest number of errors), so you could test a fork like this out in the meantime if you wanted to.

@jmshal if you have the time/interest, that would be most welcome. ✨

ticky:
We’re already packaging it in with Webpack and pulling in from npm, so swapping it out for a git ref is doable!

That said, the changes in this PR seem outdated compared to the latest release, and I don’t think I have enough context to resolve the conflicts myself in a rebase!

jmshal:
> That said, the changes in this PR seem outdated compared to the latest release, and I don’t think I have enough context to resolve the conflicts myself in a rebase!

Ah yes, sorry about that, @ticky, just sorting that out for you now.

jmshal:
@ticky It's worth noting that this feature only works when setting `notifyHandler` to `xhr`, because the way the endpoint currently responds to the browser causes the default `notifyHandler` (using `Image` objects) to trigger the `onerror` event, as the response is not a valid renderable image.

It's also been a while since this branch was created, so I still need to fully re-test the functionality to ensure nothing has broken in the mean time. My initial post-merge tests seemed to function correctly.

bengourley:
Need to decide whether to work this into the current version or the next major.

bengourley:
@jmshal we're going to close this off and ensure that we add offline support to the next major version

jmshal:
🕺 looking forward to it!

jacobg:
@bengourley Was offline support ever added? It's particularly important for hybrid mobile apps (e.g., cordova).

ticky:
@jacobg as far as I can tell, offline support hasn’t been added at all. #117 is still open, and I can find no indication that the library handles failures to send reports at all.

PR #161, opened 2016-08-02
trim target text in click events

### LINKED ISSUE #158: Trim whitespace targetText function()
There is too much whitespace getting sent when generating event `targetText` metadata for breadcrumbs. 

Should just be able to trim it.

wordofchristian:
closes #158

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -503,7 +503,7 @@
   // extract text content from a element
   function nodeText(el) {
     var text = el.textContent || el.innerText;
-    return truncate(text, 40);
+    return truncate(text.trim(), 40);
   }
 
   // Create a label from tagname, id and css class of the element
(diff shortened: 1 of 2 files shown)


src/bugsnag.js
```diff
@@ -503,7 +503,7 @@
   // extract text content from a element
   function nodeText(el) {
     var text = el.textContent || el.innerText;
-    return truncate(text, 40);
+    return truncate(text.trim(), 40);
```
jacobmarshall: Won't this break IE8 support? As `String.prototype.trim` was [first supported in IE9](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim#Browser_compatibility).

wordofchristian: Good, catch, should be fixed on master

PR #162, opened 2016-08-02
Remove lastEvent tracking

### LINKED ISSUE #160: Remove lastEvent tracking
This feature is now pretty much redundant with breadcrumbs and is no longer necessary.

wordofchristian:
closes #160

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -12,7 +12,6 @@
 // The `Bugsnag` object is the only globally exported variable
 (function (window, old) {
   var self = {},
-    lastEvent,
     lastScript,
     previousNotification,
     shouldCatch = true,
@@ -26,10 +25,7 @@
     eventsRemaining = 10,
     // The default depth of attached metadata which is parsed before truncation. It
     // is configurable via the `maxDepth` setting.
-    maxPayloadDepth = 5,
-
-    // helper function for relative times
-    millisecondsAgo = makeMillisecondsAgo();
+    maxPayloadDepth = 5;
 
   // #### Bugsnag.noConflict
   //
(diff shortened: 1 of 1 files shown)


jacobmarshall:
Sorry, I missed this. What was the reason behind removing last event tracking?

wordofchristian:
We decided it seems redundant now that we have breadcrumbs. Is there some reason we should keep it around?

jacobmarshall:
Just noticed the original issue. Though breadcrumbs do cover off a lot of the ground which `lastEvent` did, there are still times where `lastEvent` helped provide a lot of clarity (where user integration is not involved). For instance `onmessage` on iframes, or `onload`/`onreadystatechange` for XHR.

Maybe it's worth including network requests (XHR/fetch) as another breadcrumb type? Just thinking aloud really. And maybe most people's use cases don't need to class XHR as a valuable event, however (especially on a timeline, like breadcrumbs) it would provide even more clarity. The only issue here is the bulk that adding more and more breadcrumb puts on the script file.

jacobmarshall:
It's also not just XHR & iframes. There are many more events which are only really useful if they're the "last event" that happened, which don't get captured by breadcrumbs.

For instance, seeing that the last event is `keydown` on a `<textarea>`, instead of an `onclick` on a `<li>` for a typeahead component, narrows down the entry point of an error.

What do you think about a special type of breadcrumb which only shows the last occurrence for these kinds of events, so the feed doesn't get spammed? So the breadcrumb can still be used, but the functionality of `lastEvent` isn't lost entirely. Again, just thinking aloud. Not sure how other people use `lastEvent`, and whether or not we're alone here 😛

jacobmarshall:
Sorry for triple commenting, but maybe this is worth me opening a new issue for this?

eanakashima:
@jacobmarshall thanks, some great points here. I think our long-term goal is that breadcrumb tracking will be able to handle all of the lastEvent use cases and more — but there are a few we're missing right now. I've opened another issue (#166) to discuss which ones we'd like to add — if you think of any more, please let us know. ✨

jacobmarshall:
Thanks @eanakashima, I appreciate it.

PR #164, opened 2016-08-02
Remove deprecated userId setting

### LINKED ISSUE #153: Remove userId setting in v3
Seeing as #149 was just merged into `master` and it increments the notifier version to 3.0.0-rc.1, I just thought it was worth mentioning that the `userId` setting was deprecated and due to be removed in v3.

See https://github.com/bugsnag/bugsnag-js/blob/master/src/bugsnag.js#L771

eanakashima:
We used to support sending a `userId` setting, but this has been deprecated in favor of the [user](https://github.com/bugsnag/bugsnag-js#user) object. This PR will remove userId support for v3.

Closes #153 (h/t @jacobmarshall)
👋 @kattrali @wordofchristian

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,16 @@ Changelog
 -----
 
 ### Enhancements
-  
+
 - Add support for sending "breadcrumbs" - notable events leading up to an error
   [Christian Schlensker](https://github.com/wordofchristian)
   [#149](https://github.com/bugsnag/bugsnag-js/pull/149)
 
-  
+### Deprecations/Breakages
+
+- Remove support for deprecated `userId` in favor of the `user` object.
+
+
 2.5.0 (2016-01-29)
 -----
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,16 @@ Changelog
 3.0.0-rc.1 (2016-07-22)
 -----
 
+This release adds support for automatic and custom breadcrumb tracking.
+Breadcrumbs replace the private implementation of "last event" tracking.
+It also removes support for the deprecated `userId` setting, which has been
+replaced by the `user` object.
+
 ### Enhancements
 
 - Add support for sending "breadcrumbs" - notable events leading up to an error
   [Christian Schlensker](https://github.com/wordofchristian)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,12 +5,16 @@ Changelog
...
 
 ### Enhancements
-  
+
 - Add support for sending "breadcrumbs" - notable events leading up to an error
   [Christian Schlensker](https://github.com/wordofchristian)
   [#149](https://github.com/bugsnag/bugsnag-js/pull/149)
 
-  
+### Deprecations/Breakages
```
eanakashima: Do we have a more standardized header for this kind of thing?

wordofchristian: We should also add the note about removing `lastEvent` here.

eanakashima: Yeah. I'm sitting on that diff right now, I'll push it after we merge this.

kattrali: > Do we have a more standardized header for this kind of thing?

I’ve been working it into a summary paragraph before the sections. [This is maybe an example](https://github.com/bugsnag/bugsnag-cocoa/blob/master/CHANGELOG.md#500), as is [this](https://github.com/bugsnag/bugsnag-php/blob/master/CHANGELOG.md#300-2016-07-07). If we need some instructions for how to migrate, I’d recommend an UPGRADING.md similar to [bugsnag-php](https://github.com/bugsnag/bugsnag-php/blob/master/UPGRADING.md)

eanakashima: @kattrali thanks. Would something like this be more in line with our other repos? Open to suggestions.

```markdown

3.0.0-rc.1 (2016-07-22)
-----

This release adds support for automatic and custom breadcrumb tracking.
Breadcrumbs replace the private implementation of "last event" tracking.
It also removes support for the deprecated `userId` setting, which has been
replaced by the `user` object.

### Enhancements

- Add support for sending "breadcrumbs" - notable events leading up to an error
  [Christian Schlensker](https://github.com/wordofchristian)
  [#149](https://github.com/bugsnag/bugsnag-js/pull/149)

```

kattrali: Looks good to me! 👍

PR #165, opened 2016-08-03
Send time error occurred in payload

eanakashima:
We currently show the time the error was _received_ as the starting point in the breadcrumb list, but if the client clock is far off from our server time, it can look like there's a gap between all the breadcrumbs and the error. This adds a "Device.time" timestamp like we have for our mobile notifiers, so we can show the whole breadcrumb list in client time. The new version will look like this, changing the top field from "Error Received" to "Error Occurred" in the UI:

screen shot 2016-08-02 at 6 09 53 pm

👋  @kattrali @wordofchristian

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ replaced by the `user` object.
 - Add support for sending "breadcrumbs" - notable events leading up to an error
   [Christian Schlensker](https://github.com/wordofchristian)
   [#149](https://github.com/bugsnag/bugsnag-js/pull/149)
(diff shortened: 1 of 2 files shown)


wordofchristian:
👍 looks good

kattrali:
I think `device` should be lowercase (though it seems to work both ways?), to match the other implementations. I’ll make a note of it in the API spec though. Looks good ⭐

GrahamCampbell:
This is something we might want to do on the PHP notifier too, in particular if we have batch sending enabled.

Issue #166, opened 2016-08-03
Add support for more types of automatic breadcrumbs


eanakashima:
Over on #162 @jacobmarshall mentioned some cases where our old `lastEvent` tracking would provide information we don't currently collect automatically with breadcrumbs. I wanted to open an issue to collect feedback on other types of breadcrumbs that might be worth adding, to make sure the new breadcrumb feature covers all the cases that last event did. Please suggest more. Here are some I pulled from Jacob's comments: 

- [ ] `message` event from an iframe
- [ ] XHR/fetch tracking
  - onreadystatechange
  - load/error
- [ ] Keyboard events
  - we'll want to do something to prevent logging user input, e.g. roll these up with a count and only show the last keypress value

jacobmarshall:
`message` events from Web Workers may also be pretty cool to catch with breadcrumbs. Simply because for us, as an encrypted product, we use Web Workers everywhere, and knowing whether a message came through correctly would be a huge bonus. However I see definitely understand if this was pushed away from being part of the core library, as it's probably not as critical for most apps.

Regardless, I see a huge benefit in the originally listed events alone. I am happy to put together a PR, if you'd like? Or if you'd just like to collect feedback for now, I'll sit on this for the time being.

As a side note, I'm not sure how much pressure there is to keep the library tiny anymore. I know when Conrad was working with you guys there was definitely an effort to keep the library as TINY as possible, which is great, but the breadcrumb code has bulked it up a bit. I wonder whether or not it's worth making some alternatives for people who would like to make the most of the amazing  breadcrumb functionality, and for people who aren't interested? Such as providing bundles which provide bugsnag-js core & detailed breadcrumbs, and maybe even commonjs modules for people who wish to decide which type of breadcrumbs they're only interested in. But maybe this all ends up making it harder for people to jump straight into having the full Bugsnag experience? (Thinking aloud again)

Anyway, holler if you'd like to see a PR 💃

wordofchristian:
We'd welcome a PR for web workers. We currently have the `process` type set aside for these. We are already using it in the mobile/native libraries, but haven't started anything for JS.  You'd know better what kind of things you'd like to track.

The size goal is to keep the library under 4.5 KB minified

jacobmarshall:
Sweet, cheers, web workers it is then 🏹

bengourley:
Need to decide whether we want to add this to the current version vs. the next major.

PR #170, opened 2016-08-06
[WIP] Support for web worker breadcrumbs

jmshal:
Still early, and needs some more thorough testing (mainly to make sure overriding `Worker` doesn't affect existing behaviour), however this adds support for sending breadcrumbs on some key web worker events.
- [x] Creating a worker
- [x] Sending messages to a worker
- [x] Receiving messages from a worker
- [x] Terminating a worker

screenshot 2016-08-06 20 25 22

Each worker is given an auto incrementing id, as (at least in my case) when working with Workers you usually use more than one to distribute the load. So every breadcrumb includes the associated Worker's id.

Workers seem to put up a bit of a fight when trying to intercept them. As their `message` event doesn't bubble up to the `window` object. So manual overriding is necessary.

Some methods need to be bound directly to the instance, to make it clear which worker is sending/receiving messages. It would've been nicer if IE10 supported `WeakMap`, so the `Worker.prototype.(postMessage/terminate)` methods could be enhanced instead. _But I'll probably investigate a bit further into potential alternatives._

See #166

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -410,6 +410,63 @@
     enhance(history, "replaceState", wrapBuilder(buildReplaceState));
   }
 
+  function trackWorkers() {
+    if(!getBreadcrumbSetting("autoBreadcrumbsWorkers", true)) {
+      return;
+    }
+
+    var workers = 0;
+
+    window.Worker = hijackConstructor(window.Worker, function(url) {
+      var id = ++workers;
+      var startTime = Date.now();
+
+      // Bound helper function for sending worker breadcrumbs
+      //
+      // If IE10 supported WeakMap this could be done by enhancing the prototype, not
+      // the instance's methods directly.
+      function trackWorkerBreadcrumb(name, metaData) {
+        if (metaData.message) {
+          metaData.message = serialiseGenericObject(metaData.message);
+        }
+        self.leaveBreadcrumb({
+          type: "process",
+          name: name,
+          metaData: merge({
+            id: id
+          }, metaData)
+        });
+      }
+
+      trackWorkerBreadcrumb("Created worker", {
+        url: url
+      });
+
+      this.addEventListener("message", function(event) {
+        // TODO Handle automatic error catching from workers
+
+        trackWorkerBreadcrumb("Message from worker", {
+          message: event.data
+        });
+      });
+
+      // Warning: Calling the `postMessage` or `terminate` prototype methods directly
+      // against the worker object will bypass these breadcrumbs.
+
+      enhance(this, "postMessage", function(message) {
+        trackWorkerBreadcrumb("Message to worker", {
+          message: message
+        });
+      });
+
+      enhance(this, "terminate", function() {
+        trackWorkerBreadcrumb("Terminated worker", {
+          uptime: (Date.now() - startTime) + "ms"
+        });
+      });
+    });
+  }
+
   //
   // ### Script tag tracking
   //
@@ -485,7 +542,7 @@
     if (typeof oldFunction === "function") {
       object[property] = function() {
         newFunction.apply(this, arguments);
-        oldFunction.apply(this, arguments);
+        return oldFunction.apply(this, arguments);
       };
     }
   }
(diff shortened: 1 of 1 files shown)


src/bugsnag.js
```diff
@@ -410,6 +410,63 @@
...
   }
 
+  function trackWorkers() {
+    if(!getBreadcrumbSetting("autoBreadcrumbsWorkers", true)) {
+      return;
+    }
+
+    var workers = 0;
+
+    window.Worker = hijackConstructor(window.Worker, function(url) {
```
wordofchristian: I'm worried hijacking this constructor might be a little too invasive. We do it for console.log and I already don't like how that affects line numbers. We're considering creating an "examples" folder in the repo or a cookbook wiki page to demonstrate advanced customizations and breadcrumb use cases.

jacobmarshall: Yeah, I completely agree with you. You didn't see my face when writing this (🙃). Do you mean you're considering creating some docs which will show how to *manually* hijack (for instance - in this case) breadcrumbs into workers?

jacobmarshall: Idk, something like like...

```js
const worker = new Worker('./my-happy-little-worker.js');
bugsnag.attachBreadcrumbs(worker); // would setup subsequent events/method overrides
worker.postMessage('dance 💃'); // auto-breadcrumb
```

The thing is, going back to the `console.log` issue. If any of these methods (`postMessage` or `terminate`) errored, the stacktrace should still be valid, but the console would just show the top frame as coming from bugsnag.js. Which is arguably not as terrible - but I do understand it's still not favourable. We get this already with `setTimeout` and `requestAnimationFrame` anyway.

src/bugsnag.js
```diff
@@ -410,6 +410,63 @@
...
+
+      enhance(this, "postMessage", function(message) {
+        trackWorkerBreadcrumb("Message to worker", {
+          message: message
+        });
+      });
+
+      enhance(this, "terminate", function() {
+        trackWorkerBreadcrumb("Terminated worker", {
+          uptime: (Date.now() - startTime) + "ms"
```
wordofchristian: Should be ` new Date().getTime()`

src/bugsnag.js
```diff
@@ -930,6 +987,29 @@
...
+      return instance;
+    };
+    NewConstructor.prototype = Constructor.prototype;
+    NewConstructor.prototype.constructor = NewConstructor.constructor = NewConstructor;
+    return NewConstructor;
+  }
+
+  function serialiseGenericObject(obj) {
+    var str;
+    str = JSON.stringify(obj, null, "  ");
```
wordofchristian: This could be combined with previous line

src/bugsnag.js
wordofchristian: I'm coming around the way you have it currently. Internally, we currently disable the console.log breadcrumb in development to get around the line number problem. I'm actually starting to wonder if the default value for `autoBreadcrumbsConsole` and `autoBreadcrumbsWorker` should be based on the release stage.

wordofchristian:
What are your opinions on this @kattrali, @eanakashima, @snmaynard

src/bugsnag.js
jacobmarshall: That actually makes a lot of sense, however as long as it's clear that's the case so you guys don't get a million reports of "some breadcrumbs aren't showing" 😜

That or documenting and recommending what you guys do internally as a solution? Kinda ugly, but much more open and less magic.

src/bugsnag.js
jacobmarshall: All browsers which support workers, support `Date.now()`, but I can change it for consistency sake?

kattrali:
I think for this to be most usable, having it off in development (along with the console ones) is probably the best option. The line number manipulation has come up a few times as troublesome in debugging, so it would probably be best to not make it worse. In general though I think this is a good feature and very useful.

bengourley:
@jmshal we're going to close this off and ensure it either happens by default in the next major version OR is easily pluggable.

PR #173, opened 2016-08-09
Send device time with all error reports

eanakashima:
Previously, this would only happen for auto-notified error reports. This allows us to send it every time we send an error report to bugsnag.


wordofchristian:
Looks good 👍 🍰

PR #175, opened 2016-08-17
Improve notifyException definition

### LINKED ISSUE #148: Typescript definitions for notifyException are wrong
`notifyException(exception: Error, name?: string, metaData?: any, severity?: string): void;`
***
does not account for the valid method call of
`Bugsnag.notifyException(e, { foo: 5 })`

I suggest you define an additional method definition that omits name.

kattrali:
* Allow for metadata-only case
* Increase parameter documentation

Post-merge, the definitions on DefinitelyTyped will need to be updated, as well as the [tests](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/bugsnag/bugsnag-tests.ts) to include additional cases for method arguments, like:

```typescript
Bugsnag.notifyException(new Error("Something broke"));
Bugsnag.notifyException(new Error("Something broke"), "Oh no", {item: "foo"});
Bugsnag.notifyException(new Error("Something broke"), "Oh no", {item: "foo"}, "warning");
Bugsnag.notifyException(new Error("Something broke"), {item: "foo"});
```

Fixes #148

### DIFF
--- a/src/bugsnag.d.ts
+++ b/src/bugsnag.d.ts
@@ -80,11 +80,28 @@ interface BugsnagStatic {
      */
     noConflict(): BugsnagStatic;
 
-    /** Send caught exceptions to Bugsnag. Valid severity values are `info`,
-     *  `warning`, and `error`, in order of increasing severity.
-    */
-    notifyException(exception: Error, name?: string, metaData?: any,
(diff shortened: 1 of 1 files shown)


wordofchristian:
Looks good 👍

Issue #169, opened 2016-08-05
support notifyException(string)


joshunger:
It would be helpful to support notifyException(string).  Why have different methods?  This is the most confusing part of the API when using it.

kattrali:
Thank you for the report, @joshunger. I think the intent here was to differentiate between sending a report for a concrete type versus a name/message, though perhaps that intent was not clear. Is there an issue with using `notify()` instead?

joshunger:
Yeah, it wasn't clear until we got this into production.  The issue with `notify()` is that it doesn't include the stack.  Is there a way to configure it so `notify()` includes the stack?  Maybe I had it wrong.

There are also some cases where a JavaScript API will throw a function where the incoming is an Error or a String.  I believe the Bluebird Promise framework is one such example.  So, it would be helpful to not have to add this logic to my code.

jacobmarshall:
> The issue with notify() is that it doesn't include the stack.

It does include the stack, it generates one on the fly (https://github.com/bugsnag/bugsnag-js/blob/master/src/bugsnag.js#L109).

kattrali:
Thanks for the update, @joshunger, and the code spelunking, @jacobmarshall :)

`notify()` should include the stack up until the point notify was called, although it has variable quality depending on the browser used.
 
Regarding usage, I’ve updated the documentation on [reporting handled errors](https://docs.bugsnag.com/platforms/browsers/reporting-handled-errors/) to hopefully be clearer. 

Combining the `notify` and `notifyException` interfaces is something we should consider in another major revision, though for now it sounds like the larger issue is resolved. We can reopen or start a new issue if there is a problem with generated stacktraces.

joshunger:
@jacobmarshall @kattrali thanks, I think I was looking at a Promise exception which typically doesn't include stack traces.

Let me ask you one additional question, what is the expected behavior when you call `notify()` with a string?

kattrali:
> Let me ask you one additional question, what is the expected behavior when you call `notifyException()` with a string? 

The behavior under these circumstances is undefined, unless that string happens to respond as an Error would to properties like `name` and `sourceURL`. We could potentially enforce the contract by checking `instanceof Error`. I’d recommend either using `notify()` with a custom error name (maybe the name of the promise?) or creating a new Error from the string and `notifyException()`

[...]

PR #176, opened 2016-08-18
Do type checking on notifyException function

wordofchristian:
Addresses https://github.com/bugsnag/bugsnag-js/issues/169#issuecomment-240792642

Also added a new test and refactored some old tests.


src/bugsnag.js
```diff
@@ -78,6 +78,16 @@
     if (!exception) {
       return;
     }
+
+    if (typeof exception === "string") {
+      log(
+        "Bugsnag.notifyException() was called with a string. Expected instance of Error. " +
+        "To send a custom message instantiate a new Error or use Bugsnag.notify('<string>')." +
+        " see https://docs.bugsnag.com/platforms/browsers/#reporting-handled-exceptions"
```
GrahamCampbell: Very thorough. 👍

snmaynard: Could it call through to notify here?

wordofchristian: Good idea

eanakashima:
🎇  Looks good, thanks!

PR #177, opened 2016-08-22
Sauce labs integration

wordofchristian:
__NOTE: SQUASH WHEN MERGING__

This sets up sauce labs integration for Travis CI and local testing. Local browser tests require a sauce labs account. To make this process easier, I added the karma test runner.

See changes to [CONTRIBUTING.md](https://github.com/bugsnag/bugsnag-js/blob/a7c3e6fbfcb1c9904b0c1be71a8d2e61ca789f78/CONTRIBUTING.md) for instructions.

This also includes some forked source code from the [karma-mocha project](https://github.com/karma-runner/karma-mocha). This was necessary to support IE 6-8. Once karma-mocha releases their next version to npm we can remove this fork.


CONTRIBUTING.md
```diff
@@ -10,35 +10,54 @@ Contributing
...
-browsers you can get your hands on. The easiest way
-to get these is from [modern.ie](https://www.modern.ie/en-gb/virtualization-tools#downloads).
+Then ensure that you have `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` environment
+variables set.
 
-### Headless (Using PhantomJS)
+Finally run:
+
+```
+karma start --single-run
```
eanakashima: If we're going to reference the karma commands directly in the docs, it would be nice to tell people to `npm install -g karma-cli` at some point, maybe when they are installing sauce-connect.

martin308: can we stick it in the package.json scripts so that we don't have to globally install it?

wordofchristian: I could be wrong but I think it is one of those packages that prefers to be installed globally

eanakashima:
screen shot 2016-08-23 at 11 09 10 am

✨ yassss

CONTRIBUTING.md
jacobmarshall: > I could be wrong but I think it is one of those packages that prefers to be installed globally

`karma-cli` does specify `preferGlobal`, however adding it to the `devDependencies` probably wouldn't do any harm. If anything it would make updating it in the future much simpler. And because locally installed dependencies are added to the PATH when running a script, future contributors won't have to globally install multiple versions of `karma-cli`.

Meaning `karma start --single-run` can turn into `npm test` and `npm test --browsers sl_ie_6, sl_ie_7` (a few lines down).

PR #178, opened 2016-09-08
Fixing 'no implicit any' TypeScript error

olemartinorg:
This fixes the 'no implicit any' error if you have noImplicitAny set in tsconfig.json. As far as I could see, leaveBreadcrumb() never returns any value.


wordofchristian:
Thanks for this. Looks good. I'll merge as soon as I figure out what is making travis unhappy.

PR #179, opened 2016-09-08
Create "Bugsnag loaded" breadcrumb

wordofchristian:
This ensures that there is always at least one breadcrumb in the payload. Normally the first breadcrumb would be from the DOMContentLoaded event, but there are some situations when that event get's triggered before Bugsnag has finished setting up. At least with this breadcrumb you have at least some idea of when the initial page load occurred.


eanakashima:
🚀  This is great.

Issue #182, opened 2016-09-16
trackClicks function does not have permission to access property "target"


antpaw:
Firefox users get this error in my application

Error: Permission denied to access property "target"

<img width="696" alt="screen shot 2016-09-16 at 20 38 02" src="https://cloud.githubusercontent.com/assets/108543/18597347/e0827ef8-7c4d-11e6-8b18-0ff6ca502cec.png">

wordofchristian:
It seems like this is caused by a [CORS issue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Property_access_denied). 

We have some docs related to solving this.
https://docs.bugsnag.com/platforms/browsers/cors/

Though, we should update the library to fail gracefully in these situations.

antpaw:
i'm not sure if we need CORS because the javascript is hosted on the same domain https://www.tickaroo.com/ unless this error happens inside some ad sdk which i can't change. Am I right about this? 

btw this is how your blogpost looks in my chrome on the newest osx
<img width="783" alt="screen shot 2016-09-17 at 10 04 21" src="https://cloud.githubusercontent.com/assets/108543/18606940/269995dc-7cbe-11e6-8a78-4ccbe06f237e.png">

eanakashima:
> i'm not sure if we need CORS because the javascript is hosted on the same domain

Yep, if you host all your JS on the same domain, you should be fine. One possibility is that this is happening inside an iframe or something like that — an ad might be the culprit.

> btw this is how your blogpost looks in my chrome on the newest osx

The page should download a web font that has support for those characters. Any chance you have an adblocker or extension that's blocking google fonts? Or one of the font requests is slow or timing out?

antpaw:
okay thanks, our app uses alot of iframe ads and i cant change them. maybe its because i also have source sans installed on my mac

eanakashima:
> maybe its because i also have source sans installed on my mac

Ah that might do it, if you have the font already installed but your local version doesn't have the same character set. I'll open a ticket for us to look into it. Here's what the graphic looks like when everything loads, if that info is of interest:

![screen shot 2016-09-19 at 1 29 30 pm](https://cloud.githubusercontent.com/assets/187987/18647926/3518438c-7e6d-11e6-89be-38a741b73ec8.png)

eanakashima:
Mostly addressed by #183

PR #183, opened 2016-09-17
WIP fail gracefully when accessing event target

wordofchristian:
Addresses #182 . 

Need a reliable way to reproduce this error and possibly add a test.


eanakashima:
Tests and more info on how to reproduce would be great, but this seems useful by itself — going to merge it down and we can come back to the rest of it 🔜

PR #184, opened 2016-09-17
Add value check for nodeText function

antpaw:
This will make it easer to follow breadcrumbs for events that happened on submit buttons ect.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -521,6 +521,9 @@
 
   // extract text content from a element
   function nodeText(el) {
+    if (el.getAttribute('type') === 'password') {
+      return 'Password value can not be shown or it is empty';
+    }
     var text = el.textContent || el.innerText || el.value || "";
     text = text.replace(/^\s+|\s+$/g, ""); // trim whitespace
     return truncate(text, 140);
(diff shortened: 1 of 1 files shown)


jacobmarshall:
If I'm not mistaken, this could add sensitive information to the breadcrumbs, as clicking on a password field would capture this event... or am I missing something here?

antpaw:
good point

eanakashima:
Thanks @antpaw, this is a nice idea (and thanks to @jacobmarshall for a great spot on the password issue). I think we may want to scale it back to just submit buttons for now, since collecting the contents of text inputs might be problematic for some customers from a privacy perspective, but collecting submit button text seems great. Going to merge this in to tweak it a bit and add some tests around the behavior.

alfredbez:
Is there any way to collect values from `input` elements, especially `text` ones? We need that to see, what was entered in our search box.

wordofchristian:
@alfredbez It's definately possible, but we don't want to build it into the library as a default for security concerns. You can track this yourself, however, using the `Bugsnag.leaveBreadcrumb()` method.

```javascript
const form = document.querySelector('#search-form')
const input = document.querySelector('#search-form input')

form.addEventListener('submit', () => {
 Bugsnag.leaveBreadcrumb('Search submitted', { value: input.value });
});
```

PR #185, opened 2016-09-20
Exclude test folder in npm package

tsov:
The test folder should be excluded from npm. It also seems to be causing issues with some Sprockets setup in Rails.


eanakashima:
@tsov thanks for the report. Since the [fallback behavior](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) in the case of no .npmignore is to use the contents of .gitignore, it seems like we might want to combine some of what we have in there with this. Do we need any of these?
```
.eslintrc
.npmignore
.travis.yml
Gruntfile.js
bower.json
browsers.json
docs
dist
karma.config.js
test
```

tsov:
Sorry for the delay! Ideally only these files should be included in the package:

```
dist/
package.json
component.json
LICENSE.txt
```

What do you think?

eanakashima:
Sounds good — I've pulled in your change and added a few more things to make the list match up to what we discussed here. Please let us know if you run into any more issues.

PR #190, opened 2016-09-30
use console.log breadcrumbs even if console is not defined.

### LINKED ISSUE #189: Non working `autoBreadcrumbsConsole` preference
This `if` checks only first condition, which usually returns `true` and any other conditions aren't checked:
```
if(!window.console || typeof window.console.log !== "function" || !getBreadcrumbSetting("autoBreadcrumbsConsole")) {
  return;
}
```

https://github.com/bugsnag/bugsnag-js/blob/d1a157861e15d67cfe7864c8879269a30e05e90a/src/bugsnag.js#L294

wordofchristian:
Currently we  unnecessarily disable breadcrumbs for console.log if console isn't defined. We should be able to track console.log statements in these situations.

closes #189

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -291,7 +291,7 @@
 
   // Setup breadcrumbs for console.log, console.warn, console.error
   function trackConsoleLog(){
-    if(!window.console || typeof window.console.log !== "function" || !getBreadcrumbSetting("autoBreadcrumbsConsole")) {
+    if(!getBreadcrumbSetting("autoBreadcrumbsConsole")) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


src/bugsnag.js
```diff
@@ -306,6 +306,11 @@
       });
     }
 
+    // make an empty object onto which we attach our fake logging functions
+    if (typeof window.console === "undefined") {
+      window.console = {};
```
eanakashima: We tend to be pretty conservative about adding things to the global scope, and I'm pretty worried about doing this, especially with something that doesn't work like a polyfill — if someone out there is doing something silly like using the presence/absence of `window.console` to do some kind of feature detection, I don't want to mess it up. It might be worth having a larger discussion to figure out what the most sensible thing to do here is, in the absence of `console.log` support.

Issue #156, opened 2016-07-26
Console log breadcrumb changes line number of log messages


wordofchristian:
Because of we are overriding console.log to capture breadcrumbs, those messages when printed to the console get attributed to [our script](https://github.com/bugsnag/bugsnag-js/blob/f5dfe8f6b2bc508a39da1c28584e133ebc08d395/src/bugsnag.js#L481) rather than the original source.

![](https://www.dropbox.com/s/rer76ol6zud4g69/Screenshot%202016-07-26%2015.38.00.png?dl=1)

It would be nice to find a way to avoid this.

wordofchristian:
Did a lot of searching on this problem. There are [several ways](http://stackoverflow.com/questions/13815640/a-proper-wrapper-for-console-log-with-correct-line-number) to duplicate the original console.log function without altering the line number, but I haven't yet found an approach that let's you perform some logic before calling the original. 

Some browsers have a feature called [blackboxing](https://developer.chrome.com/devtools/docs/blackboxing) which will allow people to ignore the bugsnag.js library in stacktraces, but this is less than ideal as it requires extra setup.

ptrwtts:
I need this functionality to debug an app. My standard method of logging to the console is useless without line numbers. Is there any way to disable this, at least during development, without completely turning off bugsnag?

wordofchristian:
@ptrwtts 

Yes, you can disable any of the automatic breadcrumbs https://docs.bugsnag.com/platforms/browsers/configuration-options/#autobreadcrumbs

in code:
```javascript
Bugsnag.autoBreadcrumbsConsole = false;
```

as script tag attribute
```
<script src="//d2wy8f7a9ursnm.cloudfront.net/bugsnag-3.min.js"
        data-apikey="YOUR-API-KEY-HERE"
        data-autobreadcrumbsconsole="false"></script>
```

We're actually considering disabling the console breadcrumbs in development by default.

ekeric13:
For bugsnag via npm version 3.3.2 I was running into some issues.

For some reason `Bugsnag. autoBreadcrumbsConsole = false` was not working.

But `Bugsang. disableAutoBreadcrumbsConsole()` worked.

Looks like there was some changes to v4 so probably not relevant to people using that version.

DmitryGonchar:
Nope, same in v4

PhilibertDugas:
[...]

PR #191, opened 2016-09-30
Disable automatic breadcrumb logging except in production

wordofchristian:
Meant to address #156 

This change defaults the automatic breadcrumbs for console logging to be disabled unless the release stage is set to production. 

Unfortunately this means that the breadcrumb will be disabled if you don't end up setting the release stage at all. I think this is preferable though to the alternative of having the console log line numbers messed up and not knowing why.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -291,7 +291,9 @@
 
   // Setup breadcrumbs for console.log, console.warn, console.error
   function trackConsoleLog(){
-    if(!getBreadcrumbSetting("autoBreadcrumbsConsole")) {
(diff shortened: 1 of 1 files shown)


wordofchristian:
This is going to require more thought. There is a problem with the setup of console log tracking the way it is currently architected. If it is disabled during initial setup, it cannot be re-enabled after the fact.

We might need to change how the breadcrumb settings work so that we have setter functions so that when they get set we can create the wrappers for native browser functions.

wordofchristian:
Abandoning in favor of #199

PR #194, opened 2016-10-04
Limit total number of breadcrumbs

### LINKED ISSUE #193: Limit the number of breadcrumbs
When using breadcrumbs, I noticed that they are never cleared or limited. In a long-running SPA, this can lead to significantly large breadcrumbs.

Our application, for example is to be opened once per day and then worked with the whole day without a reload or real page navigation.

I would suggest to limit the breadcrumbs to 50 (just a random number!) by default.

eanakashima:
This caps the total quantity of breadcrumbs we'll track and send at 20. This should keep things more in line with our other notifier libraries and help keep the payload size more manageable. 

Fixes #193. @wordofchristian would you mind giving this a review when you have time?

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -194,7 +194,7 @@
 
       // limit breadcrumb trail length, so the payload doesn't get too large
       if (breadcrumbs.length > breadcrumbLimit) {
-        breadcrumbs = breadcrumbs.slice(0, breadcrumbLimit);
+        breadcrumbs = breadcrumbs.slice(-breadcrumbLimit);
       }
     }
   };
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -18,6 +18,9 @@
     ignoreOnError = 0,
     breadcrumbs = [],
 
+    // Cap total breadcrumbs at 20, so we don't send a giant payload.
+    breadcrumbLimit = 20,
+
     // We've seen cases where individual clients can infinite loop sending us errors
     // (in some cases 10,000+ errors per page). This limit is at the point where
     // you've probably learned everything useful there is to debug the problem,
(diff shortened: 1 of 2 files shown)


test/test.bugsnag.js
```diff
@@ -508,6 +508,21 @@ describe("Bugsnag", function () {
 
         assert.equal(crumb.metaData.message.length, 140);
       });
+
+      it("limits total breadcumbs to 20", function () {
```
kattrali: "breadcrumbs"

src/bugsnag.js
```diff
@@ -188,6 +191,11 @@
     } else {
       crumb.name = truncate(crumb.name, 32);
       breadcrumbs.push(truncateDeep(crumb, 140));
+
+      // limit breadcrumb trail length, so the payload doesn't get too large
+      if (breadcrumbs.length > breadcrumbLimit) {
+        breadcrumbs = breadcrumbs.slice(0, breadcrumbLimit);
```
kattrali: If I am reading this correctly, this logic removes the newest breadcrumb in the case that the list is too long, instead of the oldest one. If that is the case, we should probably shift crumbs onto the list or slice off the first element.

test/test.bugsnag.js
```diff
@@ -508,6 +508,21 @@ describe("Bugsnag", function () {
...
+          Bugsnag.leaveBreadcrumb("I am breadcrumb " + i);
+        }
+        Bugsnag.notify("Something");
+
+        // Do shenanigans to get around IE<9 not supporting Object.keys
+        breadcrumbs = requestData().params.breadcrumbs;
+        for (key in breadcrumbs) {
+          if (breadcrumbs.hasOwnProperty(key)) breadcrumbCount++;
+        }
+        assert.equal(breadcrumbCount, 20);
```
kattrali: Might want to test here that the correct breadcrumb is removed.

src/bugsnag.js
eanakashima: @kattrali ah hilarious, yup, great catch. 🌈  I'll update!

wordofchristian: This is correct. `breadcrumbs = breadcrumbs.slice(-1 * breadcrumbLimit)` would be the way to do it.

eanakashima:
@kattrali added some fixes for the things you mentioned, thanks for the review! 🔬

kattrali:
No problem, looks good!

PR #195, opened 2016-10-04
Don't create global console object if none exists

wordofchristian:
To address https://github.com/bugsnag/bugsnag-js/pull/190#pullrequestreview-2818426


eanakashima:
👍  LGTM

PR #197, opened 2016-10-05
Allow calling Bugsnag.notify() with no arguments

### LINKED ISSUE #196: Calling Bugsnag.notify() with an undefined/null argument silently fails
If Bugsnag.notify() is called with no argument, or an undefined/null variable it will silently fail. The error is send to Bugsnag but dropped as it doesn't have an error class.

wordofchristian:
fixes #196

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -76,6 +76,7 @@
   // will not see it in your dashboard.
   self.notifyException = function (exception, name, metaData, severity) {
     if (!exception) {
+      log("Bugsnag.notifyException() called with no arguments");
       return;
     }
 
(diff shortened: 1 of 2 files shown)


snmaynard:
I think we should actually notify in this case

wordofchristian:
Should we just send `"Bugsnag.notify() called with no arguments"` as the name?

snmaynard:
as the message maybe. Can we put something like BugsnagError as the class?

src/bugsnag.js
```diff
@@ -115,6 +118,12 @@
   // Notify Bugsnag about an error by passing in a `name` and `message`,
   // without requiring an exception.
   self.notify = function (name, message, metaData, severity) {
+    if (!name) {
+      name = "UnspecifiedBugsnagError";
```
eanakashima: Might be worth pulling this to the top of the file where we declare some other constant-esque variables?

eanakashima:
👍  Looks good! One thought: since this is a manual call to Bugsnag.notify, maybe the error class doesn't need to have "Error" in it? A class name like `BugsnagNotify` might get closer to suggesting the typical severity.

PR #198, opened 2016-10-06
Fix tests not reloading when using watch mode

wordofchristian:
Changes to the test file weren't getting reloaded when running `npm run test:watch`. Reordering the karma files so `"test/test.bugsnag.js"` comes before ` { pattern: "test/**/*", watched: true, included: false, served: true }` fixes the problem. 

Also upgraded karma to the latest version.


eanakashima:
👍  💯  but also 💻 ❔ 🔮

PR #199, opened 2016-10-06
Add breadcrumb configuration methods

### LINKED ISSUE #156: Console log breadcrumb changes line number of log messages

### LINKED ISSUE #189: Non working `autoBreadcrumbsConsole` preference

wordofchristian:
fixes #156 
fixes #189 

This PR creates new methods on the Bugsnag objects for enabling/disabling the different automatic breadcrumbs. 

```
Bugsnag.enableAutoBreadcrumbs();
Bugsnag.disableAutoBreadcrumbs();
Bugsnag.enableAutoBreadcrumbsClicks();
Bugsnag.disableAutoBreadcrumbsClicks();
Bugsnag.enableAutoBreadcrumbsErrors();
Bugsnag.disableAutoBreadcrumbsErrors();
Bugsnag.enableAutoBreadcrumbsConsole();
Bugsnag.disableAutoBreadcrumbsConsole();
Bugsnag.enableAutoBreadcrumbsNavigation();
Bugsnag.disableAutoBreadcrumbsNavigation();
```

These are necessary because the console and navigation breadcrumbs in particular need to wrap native browser functions in order to work. This has some drawbacks as seen in #156 which is why many people will want to disable them in the development environment. Unfortunately current API for configuring these options by assigning an attribute on the Bugsnag object (`Bugsnag.autoNotifyBreadcrumbsConsole = false`) does not give us the opportunity to do the necessary setup/teardown. 

Even though these new methods are only necessary for the console and navigation breadcrumbs, I went ahead and created functions for all the breadcrumb options to keep the API consistent.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -324,14 +324,12 @@
     }
 
     // keep track of functions that we will need to hijack
-    var originalLog, originalWarn, originalError;
+    var nativeLog = console.log,
+      nativeWarn = console.warn,
+      nativeError = console.error;
 
     self.enableAutoBreadcrumbsConsole = function() {
       self.autoBreadcrumbsConsole = true;
-
-      originalLog = console.log;
-      originalWarn = console.warn;
-      originalError = console.error;
 
       enhance(console, "log", function() {
         trackLog("log", arguments);
@@ -349,9 +347,9 @@
     self.disableAutoBreadcrumbsConsole = function() {
       self.autoBreadcrumbsConsole = false;
 
-      console.log = originalLog;
-      console.warn = originalWarn;
-      console.error = originalError;
+      console.log = nativeLog;
+      console.warn = nativeWarn;
+      console.error = nativeError;
     };
 
     if(getBreadcrumbSetting("autoBreadcrumbsConsole", true)) {
@@ -365,9 +363,6 @@
 
   // Setup breadcrumbs for history navigation events
   function setupNavigationBreadcrumbs() {
-    // setup variables of functions that we will need to hijack.
-    var nativePushState, nativeReplaceState;
-
 
     function parseHash(url) {
       return url.split("#")[1] || "";
@@ -471,12 +466,13 @@
       return;
     }
 
+    // keep track of native functions
+    var nativePushState = history.pushState,
+      nativeReplaceState = history.replaceState;
+
     // create enable function
     self.enableAutoBreadcrumbsNavigation = function() {
       self.autoBreadcrumbsNavigation = true;
-      // keep track of native functions
-      nativePushState = history.pushState;
-      nativeReplaceState = history.replaceState;
       // create hooks for pushstate and replaceState
       enhance(history, "pushState", wrapBuilder(buildPushState));
       enhance(history, "replaceState", wrapBuilder(buildReplaceState));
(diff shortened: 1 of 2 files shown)


src/bugsnag.js
```diff
@@ -261,17 +261,16 @@
...
-    }
-
+  function setupClickBreadcrumbs() {
     if (!_hasAddEventListener) {
       return;
     }
 
-
     var callback = function(event) {
+      if(!getBreadcrumbSetting("autoBreadcrumbsClicks", true)) {
```
eanakashima: I know this is just moving what we had before, but the second argument here doesn't do anything and it might be nice to take it out to avoid confusion.

src/bugsnag.js
```diff
@@ -314,34 +318,56 @@
...
+        trackLog("error", arguments);
+      });
+    };
+
+    self.disableAutoBreadcrumbsConsole = function() {
+      self.autoBreadcrumbsConsole = false;
+
+      console.log = originalLog;
+      console.warn = originalWarn;
+      console.error = originalError;
```
eanakashima: This could be bad in a case where someone calls one of the disable functions before enable has been called. Maybe we could handle that case with something like:

```javascript
 console.log = originalLog || console.log;
```

src/bugsnag.js
```diff
@@ -427,22 +453,83 @@
...
+  };
+
+  self.disableAutoBreadcrumbsClicks = function() {
+    self.autoBreadcrumbsClicks = false;
+  };
+
+  self.enableAutoBreadcrumbs = function() {
+    self.enableAutoBreadcrumbsClicks();
+    self.enableAutoBreadcrumbsConsole();
+    self.enableAutoBreadcrumbsNavigation();
```
eanakashima: Just curious — how come we don't have to call `enableAutoBreadcrumbsErrors` here?

src/bugsnag.js
wordofchristian: Good catch, I went with an alternative that should also cover the edge case of calling one of the `enable` functions twice.

eanakashima:
🍰  Looks nice to me.

@kattrali I know this is a little different from what we do for some of the other notifiers, but do the function names, etc. at least seem in the same genre?

src/bugsnag.js
eanakashima: Nice 👍

src/bugsnag.js
```diff
@@ -790,8 +875,10 @@
   // get breadcrumb specific setting. When autoBreadcrumbs is true, all individual events are defaulted
   // to true. Otherwise they will all defaul to false. You can set any event specicically and it will override
   // the default.
-  function getBreadcrumbSetting(name) {
-    var fallback = getSetting("autoBreadcrumbs", true);
+  function getBreadcrumbSetting(name, fallback) {
```
kattrali: I’m a little confused on what’s going on here, sometimes `getBreadcrumbSetting` is called with `true` and other times with no argument. Are those resolving to the same thing? If so, I think it makes more sense to remove the `fallback` argument.

Issue #200, opened 2016-10-07
Infinite recursion in truncateDeep


Omniroot:
The `truncateDeep` function (for truncating strings in nested objects) doesn't account for mutual inclusion or self-nesting, and can lead to an infinite recursion.

PR #202, opened 2016-10-07
Fixes infinite recursion in truncateDeep

Omniroot:
Fixes https://github.com/bugsnag/bugsnag-js/issues/200

### DIFF
--- a/test/test.bugsnag.js
+++ b/test/test.bugsnag.js
@@ -540,6 +540,17 @@ describe("Bugsnag", function () {
         // Confirm we kept the most recent 20 breadcrumbs instead of the first 20
         assert.equal(requestData().params.breadcrumbs[19].metaData.message, "I am breadcrumb 20");
       });
+
+      it("lets me set metaData with self nesting", function() {
+        var metaData = {a: metaData, b: "text"};
+        Bugsnag.leaveBreadcrumb("deepCrumb", metaData);
+        Bugsnag.notify("Something");
+
+        var crumb = requestData().params.breadcrumbs[1];
+
+        assert.equal(crumb.name, "deepCrumb");
+        assert.equal(crumb.metaData.b, "text");
+      });
     });
 
     describe("click tracking", function () {
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -21,7 +21,6 @@
 
     // Cap total breadcrumbs at 20, so we don't send a giant payload.
     breadcrumbLimit = 20,
-    
 
     // We've seen cases where individual clients can infinite loop sending us errors
     // (in some cases 10,000+ errors per page). This limit is at the point where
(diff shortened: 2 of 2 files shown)


eanakashima:
@Omniroot thanks for the report and the PR. If you have a chance to tweak the approach to not rely on `indexOf` that'd be rad — if not we can take a look into doing it on our end 🔜  and hopefully add some tests to prevent a regression in the future.

Omniroot:
I fixed up the code so that it doesn't use indexOf. I'm not quite sure how to write test cases for the `truncateDeep` function, as it's not exposed through a public interface. The basic test case I would do is give it `var a = {a: a, b: 'long text', c: 'long text'}` with length `4` and expect an object with the same structure as `var a = {a: a, b: 'long', c: 'long'}`.

wordofchristian:
@Omniroot I generally like to find a public method that uses the private method and test it indirectly. The leaveBreadcrumb method should work https://github.com/bugsnag/bugsnag-js/blob/master/src/bugsnag.js#L143-L154

```javascript
var metaData = {metaData: metaData, b: 'long text', c: 'long text'};
Bugsnag.leaveBreadcrumb('deepCrumb', metaData);
```

wordofchristian:
Unfortunately we have an issue with TravisCI right now in that it won't send our SauceLabs credentials for non-bugsnag contributors. We'll have to run the tests ourselves until we get a workaround figured out.

Omniroot:
I added a test case for self-nested metaData. My headless testing passed, but I'm running on a Mac, so I can't run the IE 6 tests on my end.

wordofchristian:
@Omniroot Ran the sauce labs tests and everything looks good. Did you have any other feedback @eanakashima ?

Omniroot:
ping

test/test.bugsnag.js
```diff
@@ -540,6 +540,17 @@ describe("Bugsnag", function () {
         // Confirm we kept the most recent 20 breadcrumbs instead of the first 20
         assert.equal(requestData().params.breadcrumbs[19].metaData.message, "I am breadcrumb 20");
       });
+
+      it("lets me set metaData with self nesting", function() {
+        var metaData = {a: metaData, b: "text"};
```
eanakashima: This sets `metaData.a` to `undefined`. To make it recursively point to itself, you'd need to do something like

```javascript
var metaData = {a: "text"};
metaData.b = metaData;
```

Omniroot:
Thanks @eanakashima. I thought I had a quick fix for this issue, but it looks like it may be a bit more involved than I have time for at the moment. Can someone from your team take this over?

Omniroot:
ping

jacobmarshall:
I'm happy to complete this PR if needs me, @eanakashima

wordofchristian:
@jacobmarshall Go for it, if not we can take care of it by Friday

kattrali:
Closing in favor of #207 and #213. Fix released in v3.0.7.

PR #207, opened 2016-11-02
Fixes infinite recursion in truncateDeep

### LINKED ISSUE #200: Infinite recursion in truncateDeep

jacobmarshall:
Takes over from #202 (fixes #200)

The main issue was the fact that the new object (`newObject`) was being added to the `traversed` array, and thus never actually being matched by the traversed checking loop.

This also marks any recursive objects with `[RECURSIVE]`, and I've also added to test to check that. But from memory the tests can't be run by non-bugsnag peeps right now, is that correct @eanakashima @wordofchristian?

Cheers @Omniroot for starting work on this branch :)

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -709,7 +709,13 @@
   function truncateDeep(object, length) {
     var traversed = [], index;
     function _truncateDeep(object, length) {
-      index = traversed.indexOf(object);
+      index = traversed.length - 1;
+      while (index > 0) {
+        if (traversed[index] === object) {
+          break;
+        }
+        index -= 1;
+      }
       if (index !== -1) {
         return traversed[index];
       } else if (typeof object === "object") {
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -21,7 +21,6 @@
 
     // Cap total breadcrumbs at 20, so we don't send a giant payload.
     breadcrumbLimit = 20,
-    
 
     // We've seen cases where individual clients can infinite loop sending us errors
     // (in some cases 10,000+ errors per page). This limit is at the point where
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -707,22 +707,30 @@
 
   // truncate all string values in nested object
   function truncateDeep(object, length) {
-    if (typeof object === "object") {
-      var newObject = {};
-      each(object, function(value, key){
-        if (value != null && value !== undefined) {
-          newObject[key] = truncateDeep(value, length);
-        }
-      });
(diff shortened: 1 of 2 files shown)


eanakashima:
Thanks @jacobmarshall! The test for adding the `"[RECURSIVE]"`value is failing for me, I'll take a look and see if I can tweak that to get it working and ship this out.

jacobmarshall:
> Thanks @jacobmarshall! The test for adding the "[RECURSIVE]"value is failing for me, I'll take a look and see if I can tweak that to get it working and ship this out.

No problem. Will do. I wasn't sure if it was the CI, or my inexperience in writing tests. Of course it was the latter 😋

eanakashima:
Ahhh this all reminds me that our documentation about how to run the tests could be a lot more user-friendly. Hopefully #208 will help a bit.

jacobmarshall:
> Ahhh this all reminds me that our documentation about how to run the tests could be a lot more user-friendly. Hopefully #208 will help a bit.

Oh wow, I didn't realise that! Haha.

I have however found the cause of the tests failing. Just working on it now 👍

eanakashima:
Nice! btw, if you are running into any performance issues with the tests, I'm looking into those now, doesn't seem related to this PR at all.

jacobmarshall:
Just wanted to update you, @eanakashima, because it looks like this will need a bit of re-thinking. It's currently failing because the `merge` function is causing brand new objects for each level it goes down. Meaning that the `truncateDeep` function cannot distinguish that it's recursive.

Are you happy if I instead moved the recursive check into the `merge` function? That way the `truncateDeep` won't need to do any of those checks, because it will expect a normalised object.

eanakashima:
> Are you happy if I instead moved the recursive check into the merge function? That way the truncateDeep won't need to do any of those checks, because it will expect a normalised object.

Nice catch. That sounds like a good solution to me. 👍

jacobmarshall:
Sorry for not updating you last weekend. I spent too much time on Saturday trying to track down probably something that's relatively simple, but I've had little luck so far. I've managed to break a huge amount of the tests by modifying the `merge` function. I will take another look this Saturday and let you know how it goes.

@Omniroot wasn't kidding about it being a bit more involved 😝 

The fact the code needs to be as minimal as possible is also adds some more complexity to things. But I don't think it'll end up being an issue in this case. If everything I've done so far ends up working - I'm pretty sure I've just missed something small.

PR #208, opened 2016-11-04
Document how to run the tests in a browser

eanakashima:
I realized our docs make it look like you have to use karma and lots of node packages to run the tests, which can be daunting. This updates our docs to be clear about the fact that you can just run them in a browser like so:

yay browser tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,13 @@ npm run test:quick
 Any large changes should be tested in old IEs (we support IE 6!), and any other
 browsers you can get your hands on.
 
+While it's possible to use npm and karma to run the tests in various browsers
+from the command line, the easiest way is typically to open the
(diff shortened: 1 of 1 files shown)


eanakashima:
NB there are some timing/performance issues with these tests hanging at 82-93% complete, which I'm trying to resolve before I merge this.

jacobmarshall:
In relation to the note about Chrome test failures, I get the same thing happen even when running a local http server :/

eanakashima:
Huh. Good to know. The saucelabs Chrome tests pass, but they run an older version of Chrome, maybe it's a regression in newer versions. Looking into it.

eanakashima:
Ok, I can only reproduce if I open the file directly, on Saucelabs (Chrome 54) and with Karma `karma start --single-run --browsers Chrome` everything is fine. Maybe something to do with how karma wraps things?

PR #212, opened 2016-11-12
Add a hard breadcrumb limit (pulls in #211)

eanakashima:
This merges in #211 with a few updates:

- Add a hard limit of 2x the default limit (40 breadcrumbs)
- Add a test for the configurable limit behavior
- Fix some whitespace weirdness


eanakashima:
This change seems to break Firefox 31. Master is ok. Still looking into it.

PR #214, opened 2016-12-22
Add license to package.json

pzrq:
So it displays without an * in tools like:
https://www.npmjs.com/package/license-checker


eanakashima:
👍  thanks!

Issue #112, opened 2015-06-17
Errors in rejected promises not handled


nathanwelch:
Hi I'm not sure if this is expected or not but we have errors happening inside a try/catch in a promise where the catch rejects the promise with the caught error. Chrome shows these as errors in the dev console, but Bugsnag doesn't seem to get them. 

I put together a JS fiddle with bugsnag loaded in to demonstrate the error: https://jsfiddle.net/zz050pg4/2/ . If you just put in your bugsnag api key on the first line, you can see that it notifies correctly for errors outside the promise. The promise code below is in a library and is basically:
```
var errPromise = new Promise(function (resolve, reject) {
    try {
        // call some of our source code to cause error
    } catch (err) {
        reject(err);
    }
})
```

Any thoughts?

jacobmarshall:
This is because you are catching the errors yourself. Promises automatically reject if an error is uncaught. If you need the caught error to be passed to Bugsnag, I would recommend doing something along the lines of...

```js
new Promise(function (resolve, reject) {
    try {
        // throw an error
    } catch (err) {
        reject(err);
        Bugsnag.notifyException(err);
    }
});
```

or, better yet...

```js
new Promise(function (resolve, reject) {
    throw new Error('Whoops, I threw a bad error!');
})
.then(function () { /* Won't ever get here... */ })
.catch(Bugsnag.notifyException);
```

It turns out that `Bugsnag.notifyException` is compatible with promises, because the first argument accepts an `Error`, and promises only pass one argument to `then` and `catch`.

nathanwelch:
Thanks for the reply! Yeah that's what I was thinking we'd need to do but since Chrome showed the caught error I thought there might be another way. Unfortunately the try/catch is in another library's code and they don't provide any hooks for error handling.

flushentitypacket:
IMO, this is a bit inconvenient. I'd prefer if the uncaught exception thing would trigger when there's no explicit `catch`

```js
import Bugsnag from 'bugsnag-js' // add uncaught exception reporting
new Promise(() => { throw new Error('i wish this were reported to Bugsnag'); });
new Promise.reject().catch(() => console.log('makes sense not to report this, though'));
```

jacobmarshall:
> IMO, this is a bit inconvenient. I'd prefer if the uncaught exception thing would trigger when there's no explicit catch

[...]

PR #221, opened 2017-01-26
Support for catching unhandled rejections

jacobmarshall:
See #112 

This PR adds support for catching uncaught/unhandled Promise rejections.

Just to clarify, the handler checks the `event.reason` is "error-like" to prevent [this](https://github.com/bugsnag/bugsnag-js/blob/master/src/bugsnag.js#L93-L97) message from showing - which could be otherwise rather confusing.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -1185,6 +1185,19 @@
       });
     }
 
+    if ("onunhandledrejection" in window) {
+      // browsers with promise support have `addEventListener`
+      window.addEventListener("unhandledrejection", function (event) {
+        var err = event.reason;
+        if (err && (err instanceof Error || err.message)) {
+          self.notifyException(err);
+        } else {
+          // Hopefully the reason is a serializable value (this may yield less than useful results if reject() is abused)
+          self.notify("UnhandledRejection", err);
+        }
+      });
+    }
+
     // EventTarget is all that's required in modern chrome/opera
     // EventTarget + Window + ModalWindow is all that's required in modern FF (there are a few Moz prefixed ones that we're ignoring)
     // The rest is a collection of stuff for Safari and IE 11. (Again ignoring a few MS and WebKit prefixed things)
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -542,6 +542,14 @@
     self.disableAutoBreadcrumbsNavigation();
   };
 
+  self.enableNotifyUnhandledRejections = function() {
+    self.notifyUnhandledRejections = true;
+  };
+
+  self.disableNotifyUnhandledRejections = function() {
+    self.notifyUnhandledRejections = false;
+  };
+
   //
   // ### Script tag tracking
   //
@@ -1188,12 +1196,14 @@
     if ("onunhandledrejection" in window) {
       // browsers with promise support have `addEventListener`
       window.addEventListener("unhandledrejection", function (event) {
-        var err = event.reason;
-        if (err && (err instanceof Error || err.message)) {
-          self.notifyException(err);
-        } else {
-          // Hopefully the reason is a serializable value (this may yield less than useful results if reject() is abused)
-          self.notify("UnhandledRejection", err);
+        if (getSetting("notifyUnhandledRejections", false)) {
+          var err = event.reason;
+          if (err && (err instanceof Error || err.message)) {
+            self.notifyException(err);
+          } else {
+            // Hopefully the reason is a serializable value (this may yield less than useful results if reject() is abused)
+            self.notify("UnhandledRejection", err);
+          }
         }
       });
     }


jacobmarshall:
> I like the idea of having a listener for unhandledrejection, but it's worth mentioning that it has very poor cross-browser support right now, and unfortunately I don't see a way to polyfill it that would be in scope for bugsnag-js.

That's true. I kinda want to see what can be done, in terms of hijacking the Promise chain (in the smallest way possible - this may be thinking too wishfully :P), but I think you're right, it seems like it would definitely add some bloat.

> Since we don't have much single-browser functionality in bugsnag-js, I'm wondering that the best way would be to make clear to people that this feature is single-browser. I wouldn't want them to see only rejections for Chrome and think that their app is somehow more broken in Chrome than any other browser. Maybe have it off by default, and then have a prominent note about browser support next to the docs for how to turn it on?

👍 I think that's a good way to make it absolutely clear to users they're enabling a feature with very limited browser support. For people who really care about that, at least they would get a clear understanding as to what to expect.

> Aside: If you want to listen globally for unhandled rejections across browsers, it may be worth checking out bluebird or another library that will set this up — then you can notify Bugsnag in the rejection handling function.

Definitely worth noting. I've not used bluebird that often, but it would seem that bluebird users essentially get a `unhandledrejection` _"polyfill"_ built-in, is that right?

eanakashima:
> That's true. I kinda want to see what can be done, in terms of hijacking the Promise chain...

If you want to investigate, don't let me stop you! I couldn't think of a straightforward way to do it but I didn't dig too far in.

> 👍 I think that's a good way to make it absolutely clear to users they're enabling a feature with very limited browser support. For people who really care about that, at least they would get a clear understanding as to what to expect.

Sweet. If nobody else chimes in with a reason not to, I say we go with that. We can update to add the setting after if you don't want to take it on in this PR.

>  I've not used bluebird that often, but it would seem that bluebird users essentially get a unhandledrejection "polyfill" built-in, is that right?

Yeah, it gives you a cross-browser way to listen for all unhandled rejections out of the box. I haven't dug in to see how they do it under the hood, but seems handy.

jacobmarshall:
> If you want to investigate, don't let me stop you! I couldn't think of a straightforward way to do it but I didn't dig too far in.

I'll probably investigate this last/maybe as a follow up. Just because I know it's gonna be a stretch at least 😛 

> Sweet. If nobody else chimes in with a reason not to, I say we go with that. We can update to add the setting after if you don't want to take it on in this PR.

I'll happily take it on - something like the enabling/disabling of the breadcrumbs?

> Yeah, it gives you a cross-browser way to listen for all unhandled rejections out of the box. I haven't dug in to see how they do it under the hood, but seems handy.

It's lit 🔥

eanakashima:
> I'll happily take it on - something like the enabling/disabling of the breadcrumbs?

Awesome, yeah, breadcrumbs are probably a good example to look at.

jacobmarshall:
Cheers @eanakashima :)

jacobmarshall:
Sorry for the delay, @eanakashima. I have added a `enableNotifyUnhandledRejections()` function, as well as a `disableNotifyUnhandledRejections()`. I'm thinking that there's probably no need for the latter (let me know and I can quickly obliterate it from existence).

I also forgot that the docs are no longer part of this repo, but I put together some info... which you may or may not use for the actual docs. But I thought it was worth putting together some resources to help people with the lack of browser support with native Promises.

---

## Notifying Unhandled Rejections (Promises)

Currently there is minimal browser support for globally catching unhandled rejections from Promises. However for browsers that do support it, the JavaScript notifier will automatically send Bugsnag any unhandled rejections that occur within your app.

Check out [this MDN article](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection) to see which browsers are currently supported.

### How to enable

Because of the lack of browser support, this feature needs to be explicitly enabled.

```js
Bugsnag.enableNotifyUnhandledRejections();
```

### Third party Promise support

If you are using a third party Promise library, it's worth noting that they may not be compatible with the browser's built in `onunhandledrejection` event. So additional work may be required to catch them.

For instance, if you're using [bluebird](http://bluebirdjs.com/docs/api/error-management-configuration.html#global-rejection-events), Bugsnag will be able to automatically catch unhandled rejections, where as the RSVP.js provides their own mechanism for catching these occurances (to name just one).

Below are some examples of some popular third-party Promise libraries, and their own build-in catch-alls. If you're using a library not listed below, it's worth checking out their docs to see if they provide similar functionality.

#### RSVP.js

RSVP exposes an `error` event, which gives you a simple way to notify unhandled rejections.

```js
RSVP.on('error', function(reason) {
    Bugsnag.notifyException(reason);
});
```

#### $q (angular 1.x)

If you're using angular in your app, the [angular-bugsnag](https://github.com/wmluke/angular-bugsnag) library uses the `$exceptionHandler` service to catch any unhandled $q rejections.

*Notes:*
- [$qProvider.errorOnUnhandledRejections](https://docs.angularjs.org/api/ng/provider/$qProvider#errorOnUnhandledRejections)
- [$exceptionHandler](https://docs.angularjs.org/api/ng/service/$exceptionHandler)

#### Other

You can manually add `Bugsnag.notifyException` to your Promise chains to manually catch any rejections, like so;

```js
doSomethingDangerous()
    .then(andThenSomethingElse)
    .catch(Bugsnag.notifyException);

```

However, if you need a solution which works more automatically, rather than on a case-by-case basis, you could use a library like [unhandled-rejection](https://www.npmjs.com/package/unhandled-rejection).

```js
const unhandledRejection = require("unhandled-rejection");

const rejectionEmitter = unhandledRejection({
    timeout: 20
});

rejectionEmitter.on("unhandledRejection", (error, promise) => {
    Bugsnag.notifyException(error);
});
```

eanakashima:
@jacobmarshall I gave this a spin and it is working great, thank you. The docs also look fantastic. When we deploy the next release I'll open a PR to pull them into our docs repo. 🥇 

I'll also make a note to myself to see if we can add some docs for jQuery deferreds. It looks like the [state of the art](https://github.com/jquery/jquery/issues/2908) is to set the deferred [exception hook](https://github.com/jquery/jquery/blob/5f35b5b406ae7d504de86a3f0a5647b2fdf4f2af/src/deferred/exceptionHook.js) to your own handler, e.g.

```javascript
jQuery.Deferred.exceptionHook = Bugsnag.notifyException;
```

but it doesn't seem to allow you to filter out deferreds that are handled at some point later in the promise chain. 🤔  I'll have to do some digging and see if there's a nicer option.

eanakashima:
Update: this is live in the latest release (3.1). I've pulled in @jacobmarshall's docs and added some more info about handling deferred rejections as well:

- [configuration options](https://docs.bugsnag.com/platforms/browsers/configuration-options/#notifyunhandledrejections)
- [Promise-specific docs](https://docs.bugsnag.com/platforms/browsers/reporting-unhandled-promise-rejections/)

Thanks again!

PR #224, opened 2017-01-27
Modified my changes to package.json

Taguiar:
I did some changes and I hope you like it.
There is something wrong during compilation, but it is not with my modification.

Best regards,

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "bugsnag-js",
-	"description":"The Bugsnag is an error reporter for JavaScript that automatically detects and reports errors and exceptions in your client-side JavaScript code.",
+	"description": "The Bugsnag is an error reporter for JavaScript that automatically detects and reports errors and exceptions in your client-side JavaScript code.",
 	"keywords":[
 		"exception",
 		"bugsnag",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,5 +1,12 @@
 {
     "name": "bugsnag-js",
+	"description": "The Bugsnag is an error reporter for JavaScript that automatically detects and reports errors.",
```
wordofchristian: I would remove the word "The" when referring to Bugsnag. I think this should read.

> Automatically detect JavaScript errors on your websites, collect detailed diagnostic information, and get notified about errors in real-time.

package.json
```diff
@@ -1,5 +1,12 @@
...
+	"keywords":[
+		"exception",
+		"bugsnag",
+		"error",
+		"JavaScript"
```
wordofchristian: I would add "browsers" and "monitoring" here

package.json
wordofchristian: Also please convert tabs to spaces. We use 2 space indents.

kattrali:
Updated the PR with some suggested changes and merged manually. Thanks again, @Taguiar!

PR #225, opened 2017-01-27
Document linter usage

eanakashima:
I noticed that we never explain how to run our linter in CONTRIBUTING.md, let's fix that. ✨  This PR:

- documents eslint usage
- adds an npm wrapper: you can now run with `npm run lint` because I forget `grunt eslint` literally every time I open this repo
- fixes up a lint error for an undefined test variable in some versions of eslint

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,14 @@ Install the dependencies for the project:
 npm install
 ```
 
+### Linting
+
+We have some basic lint rules set up to catch style issues. Run with:
+
+```
+npm run lint
+```
+
 ### Headless (Using PhantomJS)
 
 ```
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "license": "MIT",
     "main": "src/bugsnag.js",
     "scripts": {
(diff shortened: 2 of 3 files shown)


wordofchristian:
I'm wondering if we should just remove the grunt command for this and change the lint command to run eslint directly: 

```json
"scripts": {
  "lint": "eslint src/bugsnag.js test/test.bugsnag.js"
}
```

Then we can get rid of the grunt plugin and not have the command in two places. 

I've kind of taken a liking to the whole [using npm as a build tool](https://www.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/) philosophy. It seems appropriate for simple tasks like this.

eanakashima:
As much as I'd like to do that, we call this task from a bunch of other grunt tasks, so it's not totally straightforward to remove the module and lint task — you can always exec the npm task from grunt but you'd end up having to write a fair amount of code to get the results to print out nicely on screen. Generally I don't love the presence of grunt in this repo and would agree on the suggestion for migrating to npm, but in the interest of getting this minor improvement shipped, I'd vote for bumping that to a separate PR to think about later.

PR #229, opened 2017-02-24
Web Worker example

jacobmarshall:
Created an example app which uses the notifier with web workers (+ breadcrumbs).

The code portion is done, minus any changes you see fit. But I still have to put together a README with a quick overview of the parts of the example which are worth highlighting.

Here's a quick list of the features this example app demonstrates;
- Breadcrumbs
  - Messages to & from the worker
  - Starting & terminating
- Adding a meta data tab containing the worker details
- Capturing uncaught errors from *within* the worker context (including stack trace)
- Capturing out-of-scope errors like syntax errors within the worker
- Handling multiple workers (with ids)


jacobmarshall:
I'm ignoring the Travis errors, because there is no way this PR could be breaking & un-breaking the build like that 😛

wordofchristian:
Yeah, it looks like there is an intermittently failing test in the build.

wordofchristian:
> @kattrali @wordofchristian does this look mostly consistent with how we're doing example/cookbook stuff on the other notifiers?

Yes, this closely matches the style of the react native example.

PR #234, opened 2017-04-21
Add a log statement when Bugsnag has finished loading (can be disabled using disableLog)

loopj:


src/bugsnag.js
```diff
@@ -1253,6 +1258,9 @@
     self.leaveBreadcrumb({ type: "navigation", name: "Bugsnag Loaded" });
   }
 
+  // Log that we've successfully loaded
+  log("JavaScript error reporting initialized (https://bugsnag.com)", false);
```
wordofchristian: Maybe we could also include some other diagnostic information, such as release stage and whether or not autonotify is enabled

apexskier:
If this featured is added, please don't make it log by default. In general, I expect to not see anything in the console in production unless something has broken. I'd be okay with this being a feature that's enabled with an `enableStartupLog` option.

bengourley:
@loopj did you still want this feature added?

Based on @apexskier's feedback I think we should at least provide a configuration option to switch off the log.

Let me know and I'll take it from here if so.

loopj:
Closing for now - we can reassess in future

PR #236, opened 2017-04-26
Use test project key instead of dashboard app key

eanakashima:
@kattrali @wordofchristian This swaps the key in our test examples to point to a Bugsnag test project instead of the dashboard project. Seem ok? They are a bit noisy to see in the dashboard project but I'm not sure if there's a historical reason we did that.


wordofchristian:
Seems ok from my perspective. I'm not aware of the historical reasons for this, if there are any.

Issue #241, opened 2017-06-02
If `stacktrace` has not the standard format error won't show up in dashboard


luisatmaniak:
While using `window.fetch` and `Bugsnag.enableNotifyUnhandledRejections` when an error occurs, somehow the stacktrace gets lost causing bugsnag.js to send `stacktrace:TypeError: Failed to fetch` to bugsnag's servers, which in turn makes the error non-visible on the dashboard. 

I suspect this has to do with the expected format for `stacktrace`

jmshal:
Hi, @luisatmaniak.

If I'm understanding this issue correctly, you're finding that the part of the stacktrace from before making the call to `fetch()` is being omitted from the error report?

While you may be seeing something like this in your dev tools (if you async stacktraces enabled)...

<img width="327" alt="screenshot 2017-06-03 10 57 41" src="https://cloud.githubusercontent.com/assets/4710291/26747674/8c85d3ac-484b-11e7-92eb-c71af07f0f12.png">

... the JS context only receives the uppermost portion of the stacktrace - the part above where it says *`Promise (async)`*. And that's what gets sent to Bugsnag. It is also the case for JS timers like `setTimeout`.

If this isn't it, please would you be able to provide a snippet of code that can be used to replicate the issue?

luisatmaniak:
Yes, that's it. The async stacktrace is lost, and it just sends a string with little information. Here's my current workaround:

```js
Bugsnag.beforeNotify = (error) => {
  error.stacktrace = error.stacktrace.replace(/chrome-extension/g, 'chrome_extension');

  if (!error.stacktrace.match(/at .+$/)) {
    error.stacktrace = error.stacktrace + '\n    at x (unknown:0:0)';
  }
};
```

As you can see this comes from a chrome extension, I replace `chrome-extension` as per the documentation.

jmshal:
From my own personal tests, it doesn't seem possible to retrieve the *full* stacktrace in JavaScript, but I'll do some more experiments to see if there's anything that can be done to, at least, make the situation better.

Though I don't see there ever being a nice solution to this when using things like `window.onerror` or `unhandledRejection` unless browsers add support for in-JS async stacktraces.

I had a thought about maybe something like this;

```js
function something() {
    fetch('https://example.com')
        .then(res => throw new Error('Simulate something bad happening'))
        .catch(Bugsnag.asyncNotify());
}
```

[...]

PR #242, opened 2017-06-03
Add asyncNotifyException method (for one-level deep async stacktraces)

jmshal:
See https://github.com/bugsnag/bugsnag-js/issues/241#issuecomment-305942433 for more context & example.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -94,12 +94,14 @@
   // In the example above, whatever code ends up calling `badFunction` will also be
   // included in the stacktrace when the `fetch` Promise chain rejects.
   self.asyncNotifyException = function () {
-    // line 1: <generated>
-    // line 2: Error
-    // line 3: generateStacktrace()
-    // line 4: asyncNotifyException()
-    var beforeStacktrace = generateStacktrace()
-      .split('\n').slice(4).join('\n');
+    var beforeStacktrace;
+    try {
+      throw new Error('');
+    } catch(err) {
+      beforeStacktrace = stacktraceFromException(err)
+        .replace('Error\n', '') // Remove Chrome's empty "Error" line
+        .split('\n').slice(1).join('\n');
+    }
     return function (err) {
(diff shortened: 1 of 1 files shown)


jmshal:
Tested in Chrome, Safari and Firefox.

src/bugsnag.js
```diff
@@ -66,6 +78,36 @@
   // ### Manual error notification (public methods)
   //
 
+  // #### Bugsnag.autoNotifyException
```
bengourley: this should say asyncNotifyException

src/bugsnag.js
```diff
@@ -29,6 +29,18 @@
     // is configurable via the `maxDepth` setting.
     maxPayloadDepth = 5;
 
+  var stacktraceName;
+  try {
+    throw Error('');
+  } catch(err) {
+    stacktraceName = (
```
bengourley: Some comment on which browsers use which error property for the stack trace would be nice.

jmshal:
Thanks @bengourley! I have my own doubts about this actually making it into the notifier. I feel at this point that it would be better if it was detailed within the docs (as a snippet), simply because of it's small size & very specific use-case. Although I do see this as a hack, I also feel like this is the best we're going to get, at least for a while - I spent a lot of time, and tried a bunch of different methods, but I settled on this as it was the least "involved". I usually like to stay optimistic about these things, but I don't think I've really seen this area of JS evolve much.

bengourley:
> I feel at this point that it would be better if it was detailed within the docs (as a snippet)

I very much agree with this.

I'll close this PR but thanks your efforts in this area! Async tracing is definitely an area JS needs to catch up in… Node's [async hooks](https://nodejs.org/api/async_hooks.html) are a start but it sounds like TC39 are (or will be) [working on something](https://github.com/node-forward/discussions/issues/28#issue-53809642) for native JS, which means we're probably in for a rocky ride of incompatibility/competing standards in the future!

Issue #188, opened 2016-09-26
hide PII div's in breadcrumbs


jhuangtw-dev:
Hi there! is there a way to exclude elements that may contain PII info from BugSnag's breadcrumbs?
similar to this behavior: http://help.fullstory.com/spp/138664

wordofchristian:
There isn't currently a UI for this like the behavior you described on full story. The best way to accomplish it would be with a [before notify hook](https://docs.bugsnag.com/platforms/browsers/customizing-error-reports/) . This let's you modify the exception payload before it is sent to Bugsnag.

```js
Bugsnag.beforeNotify = function(payload, metaData) {
  // Filter out sensitive information
  payload.breadcrumbs = payload.breadcrumbs.map(function(crumb){
    // Check 
    if(crumb.metaData.targetSelector.match(/SOMEREGEX/) {
      crumb.metaData.targetText = 'HIDDEN';
    }
    return crumb;
  });
}
```

PR #246, opened 2017-06-21
WIP: Anonymize PII's from autoBreadcrumbs

juhoha:
We're required by law to not send any personally identifiable information to third parties. Current implementation of autoBreadcrumbs leaks pretty much everything to Bugsnag. Maintaining regexps for every variation of potentially sensitive information as suggested in #188 didn't really seem feasible, so we rolled our own solution. I'm sending it to you as well, because it seems like at least one other user has had a similar problem.

This feature attaches itself to nodeText(el). If the feature is enabled, the clicked element and all of it's ancestors are checked for a data-pii attribute. If the attribute is found a placeholder is returned. If not, the contents descendants of the clicked element that contain data-pii attribute are replaced by placeholders while rest of the content remains intact. 

While this feature makes anonymization extremely easy, it requires extra DOM-traversal and won't be used by most of the users, so it should be disabled by default.

Even then, traversing down goes through entire subtree of the clicked element and is unacceptably slow: 5-30ms with ~1k elements, relatively new mid-tier laptop and chrome. Using a querySelector and displaying only placeholders seems near-instant (<1ms), so that might work...

I'll be on vacation for a week starting tomorrow. I'll get back on this once I get back. Throw me some comments in the mean while!

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -528,7 +528,16 @@
     self.autoBreadcrumbsClicks = false;
   };
 
+  self.enableautoBreadcrumbsAnonymization = function() {
+    self.autoBreadcrumbsAnonymization = true;
+  };
+
+  self.disableautoBreadcrumbsAnonymization = function() {
+    self.autoBreadcrumbsAnonymization = false;
+  };
+
   self.enableAutoBreadcrumbs = function() {
+    self.disableautoBreadcrumbsAnonymization();
     self.enableAutoBreadcrumbsClicks();
     self.enableAutoBreadcrumbsConsole();
     self.enableAutoBreadcrumbsErrors();
@@ -536,6 +545,7 @@
   };
 
   self.disableAutoBreadcrumbs = function() {
+    self.disableautoBreadcrumbsAnonymization();
     self.disableAutoBreadcrumbsClicks();
     self.disableAutoBreadcrumbsConsole();
     self.disableAutoBreadcrumbsErrors();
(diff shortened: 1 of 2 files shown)


juhoha:
After checking the tools available in IE6, decided this probably isn't worth the while.

kattrali:
Sorry about the lack of response, @juhoha. I noted to myself to come back to this once you were back but lost it in the shuffle :) 

I'd be interested to hear more about how you've been using this. Given the performance implications of doing this at all times, it would be hard to merge as-is. Have you been using this in particular browsers or under certain conditions?

juhoha:
@kattrali no problem. This has been a low prio project for me as well.

So the breadcrumbs collect the InnerText of 20 last elements clicked before an error was thrown. These elements might contain loads of personally identifiable information. In our case for example social security numbers, medical data and communication with medical staff, some of which are generated by the user as he's using the service. Generating filters for the already extracted InnerText on the fly would be very hard, messy and hacky. Declaring filtered parts directly to DOM would make this a world easier: There's no need to know what kind of substrings are considered private information, only slap a property on the elements that contain personal information and they become 'private'. For example chat message elements, form elements, etc.

On vacation I had an idea on performance. Instead of traversing all descendants of clicked element, the correct text could be generated by querySelecting all PII-descendants of clicked elements. With very cursory testing it seemed querySelecting by attribute on Chrome is about two orders of magnitude faster than traversing the tree. Using getElementsByClassname might be faster, I'd imagine indexing strategy for classes to be more aggressive than for data-attributes (and there's also compability). I didn't really expect IE6 to support querySelectorAll, but learning it lacks the support for getElementsByClassname was discouraging enough to give up plans of optimization. This was also influenced by a chat where we concluded that living without breadcrumbs altogether isn't really that big of a deal.

Any way if anyone else feels the burn to fork the fork and continue working on this, I'll leave my thoughts here. Something like this wouldn't need to walk the whole subtree.

- If E = (1) or any ancestor of E is private, the whole content is private. This part doesn't change.
- Use getElementsByClassname to fetch private elements marked in red P = (5,9)
- For each P Walk up to E, circle elements encountered C = (1,2,4) unless some of C would be in P.
- Select all direct children of every C that is not in C or P: D = (3,8,9).
- InnerText of E with private contents removed is in-order(D).map(innerText).join('')

Beautiful, isn't it...


Or something like that. Didn't write that down at the time, there's probably some edge cases to handle. And problems:

- Handling references, comparisons and state for elements sounds daunting in JS, but there's probably a solution.
- Afaik getElementsByClassname is in pre-order, last step requires post-order. There might be a feasible way around this, dunno.

Overall it's probably an overkill to go through all this just to preserve the contents of non-private elements when the clicked element has both private and non-private elements in it. The sensible (and easy) option might just be to censor the whole innerText, if the clicked element has any private descendants. Might return to the simple version at some point even though IE6 is the leading cause of alcoholism and male pattern baldness.

PR #252, opened 2017-08-04
Log console.group and console.groupCollapsed

TobiasBales:
Since it adds some context to the console.logs etc that happen in that group

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -356,6 +356,14 @@
       enhance(console, "error", function() {
         trackLog("error", arguments);
       });
+
+      enhance(console, "group", function() {
+        trackLog("group", arguments);
+      });
+
+      enhance(console, "groupCollapsed", function() {
+        trackLog("groupCollapsed", arguments);
+      });
     };
 
     self.disableAutoBreadcrumbsConsole = function() {


bengourley:
Hi @TobiasBales, thanks for the contribution. Is there any reason you didn't include `console.groupEnd()` in your additions?

KarstenBuckstegge:
Answering for @TobiasBales because he can't comment here for some reason:

"This was the first "quick win" since it shows when groups starts. I figured it is pretty obvious when a  group ends (maybe not and I can add that if there is value in it).

In a better, second step (currently wasn't able to figure out how to do it) it would be great to capture all `console.{info,err,log}` calls while in a group and putting them in as on breadcrumb when groupEnd is called, not sure how to structure data for the breadcrumbs then so the usable breadcrumbs in the ui have a decent structure and add more value than the current setup.

I am open to any suggestions for this."

fractalwrench:
@TobiasBales some users have been unable to comment on Bugsnag Github issues when an [adblocker extension is enabled](https://docs.bugsnag.com/platforms/browsers/faq/#is-bugsnag-blocked-by-ad-blockers). We'd recommend temporarily disabling any content-blocking extension to check if that is the case.

bengourley:
> it would be great to capture all console.{info,err,log} calls while in a group and putting them in as on breadcrumb

This sounds like a great idea.

We're currently a point where this module is going to be re-written in the very near future. If this feature is really important to you, I'll look at merging it as it is, otherwise I'll ensure it gets added to the requirements of the next major version. Let me know what you think!

TobiasBales:
Well, we built around it on our side but we'd like to integrate it into more components.
So I guess the question is - is there a rough timeframe for the rewrite? ;)

bengourley:
We don't have any concrete dates yet and we're still in the specification phase, so a while off. I'll look at integrating your changes in a patch version of the current module 🙂

PR #253, opened 2017-08-15
Example project improvements

bengourley:
Hey @kattrali, I gave the the basic JS example some ❤️  and set the precedent for adding future examples. Let me know what you think!

[Edit] I'm going to continue adding to this pull request, so that it may eventually contain:

- [x] Improvements to the plain js example
- [x] Vue.js example
- [x] React example
- [x] Angular 4 example
- ~Vue.js sever-rendered example~

I aim to keep this PR in a continually mergeable state, so that even if the list isn't complete, the individual tasks themselves are.

### DIFF
--- a/examples/vue/index.html
+++ b/examples/vue/index.html
@@ -1,7 +1,7 @@
 <!doctype html>
 <html>
   <head>
-    <title>Bugsnag: Plain JS Example</title>
+    <title>Bugsnag: Vue.js Example</title>
     <script src="../../src/bugsnag.js" data-apikey="API-KEY-GOES-HERE"></script>
     <script src="./node_modules/vue/dist/vue.js"></script>
   </head>
@@ -46,9 +46,11 @@
       // in the console it'll need to be logged.
       Vue.config.errorHandler = function (err, vm, info) {
         Bugsnag.notifyException(err, {
-          vueErrorInfo: info,
-          component: vm ? vm.$options.name || vm.$options._componentTag || 'app root' : undefined,
-          props: vm ? vm.$options.propsData : undefined
+          vue: {
+            errorInfo: info,
+            component: vm ? vm.$options.name || vm.$options._componentTag || 'app root' : undefined,
+            props: vm ? vm.$options.propsData : undefined
+          }
         })
         console.error(err)
       }
@@ -86,7 +88,7 @@
           }
         },
         methods: {
-          // Sets the data in a way that cause the next render() of this
+          // Sets the data in a way that causes the next render() of this
           // component to throw an error (due to referencing a property
           // of an object that doesn't exist)
           triggerRenderError: function () {
@@ -101,7 +103,7 @@
               JSON.parse('definitely not json')
             })
           },
-          // Changes the value being watched so that it throws an error
+          // Changes the value being watched such that it throws an error
           triggerWatchError: function () {
             this.doAWatchError = true
             setTimeout(function () {
--- a/examples/vue/index.html
+++ b/examples/vue/index.html
@@ -1,0 +1,115 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Bugsnag: Plain JS Example</title>
+    <script src="../../src/bugsnag.js" data-apikey="API-KEY-GOES-HERE"></script>
+    <script src="./node_modules/vue/dist/vue.js"></script>
+  </head>
+  <style>
+    body {
+      font-family: 'Menlo', monospace; line-height: 1.7;
+      color: #444; background-color: #fafafa;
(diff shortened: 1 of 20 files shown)


examples/vue/index.html
```diff
@@ -0,0 +1,115 @@
...
+    </script>
+
+    <script>
+
+      // Provide an error handler to Vue.js. Any lifecycle errors that previously would have
+      // been caught and logged by Vue.js can be sent on to Bugsnag here. This error handler
+      // completely replaces the default behaviour, which means if you want to see the error
+      // in the console it'll need to be logged.
+      Vue.config.errorHandler = function (err, vm, info) {
+        Bugsnag.notifyException(err, {
```
kattrali: If the metadata is nested, the top level values are used as tab names on the dashboard, otherwise anything in here is placed in a tab named "Custom". This data could instead be grouped in a "Vue" tab.

```diff
    Bugsnag.notifyException(err, {
+     vue: {
+       errorInfo: info,
-       vueErrorInfo: info,
        component: vm ? vm.$options.name || vm.$options._componentTag || 'app root' : undefined,
        props: vm ? vm.$options.propsData : undefined
+     }
    })
```

examples/vue/index.html
```diff
@@ -0,0 +1,115 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Bugsnag: Plain JS Example</title>
```
kattrali: I figure the title is just copypasta :)

bengourley: Doh :)

examples/vue/index.html
bengourley: ahh cool, I didn't know this

bengourley:
Added example for Angular 4 🎉 .

I tried really hard to keep the sheer amount of files required for a "hello world" project down to a minimum, but this is apparently _quite hard™_ with Angular 4.

Firstly I tried the Angular CLI (`ng new <app>`) but this output an entire kitchen sink. I found https://github.com/angular/quickstart which is a bit slimmer, and I trimmed away as much unnecessary code as I could. Hopefully the parts that are relevant for our example aren't lost in the rest of the scaffolding/boilerplate that appears to be necessary.

As a result, running the example is slightly more complex (running a web server _and_ the typescript compiler), but it's still wrapped up in the single `npm start` command.

bengourley:
haha, yeah I did manage to get it pretty small in the end. I personally wouldn't want to _encourage_ somebody to use Angular, so I think I won't maintain a "micro" quickstart for it for now 😆 

I've been thinking about your suggestion regarding dev/production mode and playing around with it in a few ways. I haven't managed to come up with anything that makes sense as yet.

Here are my thoughts…

- `Bugsnag` exists on the page as a "global singleton"
- setting `Bugsnag.releaseStage = '…'` from within the `ErrorHandler` constructor _seems_ like the wrong thing to do because
  - `ErrorHandler` is _not_ a singleton, and will be constructed many times
  - modifying a global singleton every time a class is instantiated seems like a pretty unreasonable side-effect
- the only redeeming detail here is that it will (probably) always be modified in the same way so will ultimately be a no-op for everything but the 1st run

This lead me down the path of thinking…
- set the `releaseStage` on a per `notifyException()` basis
- this appears to be what bugsnag-js does anyway (i.e. it takes the config from the global singleton and provides it to every [payload](https://docs.bugsnag.com/api/error-reporting/#json-payload))
- with this approach there are no side effects and theres is no modification of a singleton from instances of classes

The downside of this is that only errors caught by the Angular `ErrorHandler` will have the correct `releaseStage` value. Any errors caught outside of the Angular app will have the default value which may not match. I imagine this would result in pretty confusing reporting in the dashboard?

So my gut feel here is that we have two options:

1. we default to `releaseStage = 'development'` and hook into the angular `setProdMode()` function, where we then set `releaseStage = 'production'`.
  👍 we probably will be getting it right for the users
  👎 monkey patching `setProdMode()` sounds error-prone/dangerous
1. rather than trying to do something clever to interpret what the user wanted to do, we document that when they _don't_ call `setProdMode()` they should also set `releaseStage = 'development'`
  👍 no crazy business
  👎 allowing room for user error

I think also when we provide a proper framework adapter as a packaged module, this kind of problem is likely to disappear, or at least be significantly reduced.

Sorry for the essay 📚  Let me know what you think @kattrali!

kattrali:
👍👍 Thanks for looking into it, that sounds ideal. I'll also put it down for the future when we think more about separate modules.

bengourley:
@kattrali pushed a React (v16 specific) example ⚛️

I think that should probably wrap things up for this PR… saving the Vue server example for a rainy day? Could you please review the React copy and example, and if all ok we can get this merged?

PR #255, opened 2017-08-22
Generate test coverage

bengourley:
Text summary is output to terminal and detailed reports are generated in ./coverage.

[Edit] Ok, I think this is ready to merge now!


bengourley:
Ok @kattrali this should have been super simple, right? But since IE and VMs are concerned, it took me forever to get to the bottom of this 😖

Long story short, automated IE6 tests and code coverage reports are mutually exclusive (with the current test setup). Unfortunately the instrumented JavaScript produced by [`karma-coverage`](https://github.com/karma-runner/karma-coverage)/[`istanbul`](https://github.com/gotwarlost/istanbul) causes IE6 to become unresponsive. Although the tests themselves pass, the fact that it hangs means Sauce doesn't teardown/exit and the test times out.

IE7+ seem ok.

Which shall it be, **automated IE6 tests** or **automated coverage reports**?

bengourley:
btw here is the current coverage summary:

```
=============================== Coverage summary ===============================
Statements   : 74.9% ( 382/510 )
Branches     : 72.45% ( 234/323 )
Functions    : 60.23% ( 53/88 )
Lines        : 74.9% ( 382/510 )
================================================================================
```

kattrali:
Well, hmm. It would be pretty hard to give up testing the bottom of our supported browser matrix, though coverage is also important. Maybe we could skip coverage for IE6 and split the CI build a matrix? Then we could run sauce labs on IE6 on one branch and everything else on the other. There's some time overhead to this, but not as much as splitting each browser into its own run.

wordofchristian:
If we remove IE6 from the tests then we can also likely remove https://github.com/bugsnag/bugsnag-js/blob/master/test/karma-mocha/adapter.js

I believe it was added because the karma-mocha adapter was breaking in IE6

bengourley:
@kattrali sweet idea! Based on your suggestion I created two separate karma confs to provide two separate test runs: one for old IEs (6,7,8,9) and one for the rest. I factored out the shared config so we're not repeating ourselves at all.

Now that we have coverage being generated (fingers crossed, as the CI runs while I type this…) do we want to do anything with it? e.g. send it to https://coveralls.io/ or similar, so we can have a badge and/or automated checks for PRs?

kattrali:
🙏

Coveralls sounds good! We're using it on Python and Ruby already too.

bengourley:
All green! 🎉

PR #258, opened 2017-09-07
Parse array-typed options from data-* attrs. Fixes 249.

bengourley:
Array-typed settings are split on comma, with whitespace trimmed.

This fixes the inability to set "notifyReleaseStages", which can now be done as follows:

```html
  <script src="…" data-notifyreleasestages="production"></script>
  <script src="…" data-notifyreleasestages="production,staging"></script>
```

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -887,6 +887,10 @@
     if (setting === "false") {
       setting = false;
     }
+    // some settings may need extracting from comma separated strings -> arrays
+    if (name === "notifyReleaseStages" && typeof setting === "string") {
+      setting = setting.split(/\s*,\s*/);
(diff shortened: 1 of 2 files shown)


src/bugsnag.js
```diff
@@ -921,6 +925,8 @@
     // Check if we should notify for this release stage.
     var releaseStage = getSetting("releaseStage", "production");
     var notifyReleaseStages = getSetting("notifyReleaseStages");
+    // console.log(releaseStage)
+    // console.log(notifyReleaseStages)
```
kattrali: Is this supposed to be committed?

bengourley: 😱

PR #259, opened 2017-09-14
Add handled/unhandled properties to payload

bengourley:
Added `handledState` object argument to `sendToBugsnag()` detailing:

- `severity`
- `unhandled `
- `severityReason`

`payload.defaultSeverity` is set based on whether the user sets severity (or changes it via `beforeNotify` callback) such that `payload.severity === handledState.severity`.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -1296,7 +1296,7 @@
             lineNumber: err.lineNumber || err.line,
             columnNumber: err.columnNumber ? err.columnNumber + 1 : undefined
           }, metaData, {
-            severity: "error",
+            originalSeverity: "error",
             unhandled: true,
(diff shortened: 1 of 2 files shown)


snmaynard:
We might also need to change the code in the wrap function to report severity_reason timer_callback and pass the attributes required along as well

snmaynard:
It's certainly possible that would have to wait until we re-architect the lib too though.

bengourley:
> We need to make some changes to the unhandled promise listener stuff though I think

Good spot, I missed this. Added in 296e1b4.

bengourley:
I _think_ 7b34982 addresses the `wrap()` issue. However I think `exception_handler` is not the most appropriate severity reason, but it's the best we have right now.

PR #262, opened 2017-10-02
Update handled unhandled for Unity

fractalwrench:
Adds a handledState param to the notify method, which by default will be a handledException.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -111,9 +111,10 @@
 
     var state;
 
-    if (handledState) {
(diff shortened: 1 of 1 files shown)


kattrali:
This is fine though I think this should go into a separate `unity` branch rather than making it publicly accessible. Otherwise, I think there needs to be some validation here that the fourth parameter is a handled state.

bengourley:
I'm going to close this off for now. Let's look into it with the new notifier architecture soon ;)

PR #267, opened 2017-10-30
Add speculative inline script content plugin

bengourley:
Attempting to get the contents of an inline `<script>` as the `code` property of a `stackframe` when an error is caught/reported.

Results are mixed, the DOM doesn't provide access to the original page source meaning line numbers are inconsistent… for example

image

### DIFF
--- a/browser/plugins/inline-script-content.js
+++ b/browser/plugins/inline-script-content.js
@@ -1,0 +1,21 @@
+const { map } = require('../../base/lib/es-utils')
+
+module.exports = {
+  init: (client) => {
+    const addInlineContent = report => {
+      report.stacktrace = map(report.stacktrace, frame => {
+        if (frame.file.replace(/#.*$/) !== window.location.href.replace(/#.*$/)) return frame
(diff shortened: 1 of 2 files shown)


browser/plugins/inline-script-content.js
```diff
@@ -0,0 +1,21 @@
+const { map } = require('../../base/lib/es-utils')
+
+module.exports = {
+  init: (client) => {
+    const addInlineContent = report => {
+      report.stacktrace = map(report.stacktrace, frame => {
+        if (frame.file.replace(/#.*$/) !== window.location.href.replace(/#.*$/)) return frame
+        if (!frame.file || !frame.lineNumber) return frame
+        const start = Math.max(0, frame.lineNumber - 5 - 1)
+        const end = frame.lineNumber + 5 - 1
```
kattrali: What is the magic with `5` and `1` in these lines?

bengourley:
Hah, so line numbering is 1-based whereas the array indexing is 0-based.
The 5 is to grab the 5 surrounding lines above and below the culprit 🙂

PR #269, opened 2017-11-04
Don't crash when unhandled rejections have a null reason

kchodorow:
Up to line 1251, everything checks if err is null. Then
stacktraceFromException attempts to access err.stack and friends which,
of course, throws.  This creates a dummy object instead.


bengourley:
Wonderful 🎉 thanks for the contribution!

bengourley:
Released in `v3.2.2`

PR #280, opened 2017-12-06
An improvement for error argument in the notify method

### LINKED ISSUE #278: v4 errorClass vs. errorMessage
Hi, after update to v4 I wanted to accomplish the same logging result but it is quite "hacky". How can we set `errorClass` and `groupingHash` from `notify` method?

Our v3 implementation was:
`bugsnag.notify(errorClass, errorMessage, data, 'error');`

### Expected behavior
There should be probably some parameter in the notify method or as a field in opts to set `errorClass` and `groupingHash`.

### Version
v4.0.0

### Additional information
What I do currently is that I set the `errorMessage` even as an empty string (first param of `notify` method) and in `beforeSend()` I modify `errorClass` and add a `groupingHash` if provided in data.

```ts
notify (consoleSeverity: BugsnagSeverity, errorName: string, data: object, message?: string): void {
    this.bugsnagClient.notify(
        message || '',
        {
            beforeSend: (report: Report): void => {
                report.errorClass = errorName;
                if (report.metaData.hasOwnProperty('groupingHash')) {
                    report.groupingHash = (<any> report.metaData).groupingHash;
                }
            },
            metaData: {...data},
            severity: consoleSeverity,
        },
    );
}
```

rokerkony:
The type is not complete on purpose. All scalar values are still possible but not recommended.

Closes #278

### DIFF
--- a/types/client.d.ts
+++ b/types/client.d.ts
@@ -15,7 +15,7 @@ export class Client {
   public configure(opts: IConfig): Client;
   public transport(transport: ITransport): Client;
   public logger(logger: ILogger): Client;
-  public notify(error: any, opts?: INotifyOpts): boolean;
+  public notify(error: NotifiableError, opts?: INotifyOpts): boolean;
   public leaveBreadcrumb(name: string, metaData?: any, type?: string, timestamp?: string): Client;
 }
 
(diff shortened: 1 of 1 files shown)


bengourley:
Wonderful, thanks!

Issue #279, opened 2017-12-05
v4 Failing To Load With Laravel Mix


aaronhuisinga:
### Expected behavior
Bugsnag loads and functions properly when running [Laravel Mix](https://github.com/JeffreyWay/laravel-mix)

### Observed behavior
Currently, v4 will not initiate and work properly with Laravel Mix. It works when running Mix in development mode, but as soon as assets are compiled using the production command, Bugsnag fails to load with the error:
```
r._logger.warn(e) is not a function
```
It appears that something with how Webpack is compiling the libraries is preventing Bugsnag from loading properly. I'm genuinely not sure if this is an issue with Bugsnag or Mix (or maybe I'm just missing something simple), but I figured it made sense to open an issue since Mix is a pretty popular library and I'm assuming I'm going to be the only one with this issue. v3 worked without issue using Mix, and my v4 code worked fine when running via development.

### Steps to reproduce
1. Install Laravel Mix ([Instructions](https://github.com/JeffreyWay/laravel-mix/blob/master/docs/installation.md))
2. Install Bugsnag v4.0.0 as NPM dependency
3. Edit content of app.js file referenced in webpack.mix.js to load bugsnag-js and attempt to initiate Bugsnag
4. Run ``npm run production`` command in Terminal
5. Load website to see error referenced above

I apologize for the work involved in reproducing this one! Kind of a pain.

### Version
v4.0.0

### Additional information
Again, I'm uncertain if this is something with Bugsnag JS or Laravel Mix (@JeffreyWay), and I can open this issue in their repository if it's a better fit there. This is, however, the only library I've had any problems with while using Mix, and Bugsnag JS v3 worked properly.

Let me know if there is anything more I need to dig up.

bengourley:
Hi @aaronhuisinga thanks for the report! I've managed to reproduce, so thanks for the detailed steps.

> I apologize for the work involved in reproducing this one! Kind of a pain.

Don't worry 😂 this was relatively straightforward to get set up!

I'll let you know where I get with it.

bengourley:
It looks like by default laravel mix uses the [`drop_console` compressor option](https://github.com/mishoo/UglifyJS2#compressor-options) in uglify.

I'm surprised this doesn't break other libararies and user code! I'll see if there's something we can do about it.

JeffreyWay:
@bengourley Hmm, I wasn't under the impression that activating that option as a default would potentially break any libraries. I can tag a new release that removes it, if needed.

[...]

PR #281, opened 2017-12-06
Sidestep uglify's drop_console option

bengourley:
#279 was caused by uglify's `drop_console` option being set in [Laravel Mix](https://github.com/JeffreyWay/laravel-mix)'s webpack setup. I think this is a bug with the `drop_console` functionality but the maintainer of uglify-js [seems to disagree](https://github.com/mishoo/UglifyJS2/issues/2557#issuecomment-349601340).

Assigning the console method to another identifier prevents the option from taking effect (which I think goes to show how incomplete/problematic the `drop_console` feature is in the first place).

### DIFF
--- a/browser/index.js
+++ b/browser/index.js
@@ -95,10 +95,12 @@ module.exports = (opts, userPlugins = []) => {
 
 const getPrefixedConsole = () => {
   const logger = {}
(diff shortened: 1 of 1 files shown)


jbrooksuk:
👍

PR #290, opened 2017-12-16
updating npm instructions and url reference.

tremlab:
updated npm steps to make sure local reference to bugsnag can install and run in the example app. Removed yarn option for clarity.


tremlab:
After team discussion, rewriting example and instructions to use the global url source for bugsnag-js, not the user's local installation.  Removed bugsnag-js from dependencies, reviewed setup instructions (now the original yarn and npm will work).

PR #292, opened 2017-12-20
updating Typescript example app with expanded configuration, and full npm instructions.

tremlab:


tremlab:
The Typescript example app, because of the definition requirements, must have an npm install. The directions here show how to get bugsnag-js installed, and then run the example app.  (the global url option, in the js & sourcemaps examples, will not work here.)

PR #295, opened 2017-12-28
Check for CustomEvent.detail that Bluebird uses for unhandled rejections

emhagman:
I've noticed that unhandled Bluebird rejections do not have stack traces. This is because Bluebird fires a CustomEvent that contains the rejection reason inside `CustomEvent.detail.reason` instead of just having it stored in `Error.reason`.

See here: http://bluebirdjs.com/docs/api/error-management-configuration.html


emhagman:
I will revisit this. Upon testing, this has issues because of how Bluebird longStackTrace works.

PR #300, opened 2018-01-05
fixing where user & metadata are displayed.

tremlab:
currently, user & metadata set at initialization are being saved inside the APP tab, rather than in their own separate tabs.


bengourley:
Shoot, thanks for this! Publishing a patch release now.

PR #303, opened 2018-01-11
Updating example app with the latest config options

tremlab:
user, metadata & session details updated.


bengourley:
Merged changes here https://github.com/bugsnag/bugsnag-js/commit/ff8eae84ef3ab23ff4dc9fa765d314c99019ca1e but tidied up the git history.

PR #304, opened 2018-01-11
Fix logger.log that does not exist

### LINKED ISSUE #305: bugsnag throws an error when dispatching custom click event in Electron webview
### Expected behavior
Bugsnag should not crash when doing the following in Electron renderer console:
```javascript
document.querySelector('webview').dispatchEvent(new MouseEvent('click'));
```

### Observed behavior
```
TypeError client._logger.log is not a function 
    ...\node_modules\bugsnag-js\dist\bugsnag.js:1587:24 
```

### Steps to reproduce
Run the following snippet in an Electron renderer containing a webview:
```javascript
document.querySelector('webview').dispatchEvent(new MouseEvent('click'));
```

### Version
`4.1.2`

alexstrat:
Fix #305

### DIFF
--- a/browser/plugins/interaction-breadcrumbs.js
+++ b/browser/plugins/interaction-breadcrumbs.js
@@ -13,7 +13,7 @@ module.exports = {
       } catch (e) {
         targetText = '[hidden]'
         targetSelector = '[hidden]'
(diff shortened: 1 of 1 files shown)


bengourley:
This is awesome, thanks!

PR #306, opened 2018-01-12
adding session attribute to definitions for Typescript

tremlab:
"Object literal may only specify known properties, and 'autoCaptureSessions' does not exist in type 'string | IConfig'."

### DIFF
--- a/types/common.d.ts
+++ b/types/common.d.ts
@@ -7,6 +7,7 @@ interface IConfig {
   autoBreadcrumbs?: boolean;
   autoNotify?: boolean;
   appVersion?: string;
+  autoCaptureSessions?: boolean;
   endpoint?: string;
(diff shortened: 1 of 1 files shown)


types/common.d.ts
```diff
@@ -7,6 +7,7 @@ interface IConfig {
   autoBreadcrumbs?: boolean;
   autoNotify?: boolean;
   appVersion?: string;
+  autoCaptureSessions?: boolean;
   endpoint?: string;
   sessionEndpoint?: string;
   autoCaptureSessions?: boolean;
```
bengourley: It's already in the definition for this type?

tremlab: argh!!!!!  apologies, Ben.  :)

PR #314, opened 2018-01-17
Support async beforeSend callbacks

bengourley:
The list of supported functions that can be passed as a beforeSend callback is now as follows:

- sync function that returns "false" or calls `report.ignore()` to cancel
- sync function that returns a promise. Resolve promise with `false` or `report.ignore()` to cancel. `reject(err)` to signify error.
- async function (node-style) with signature (report, cb). Callback with `cb(null, false)` or `report.ignore()` to cancel. `cb(err)` to signify error.

`resolve()`/`reject()` or `cb()` must always be called if using promise/async
interface.


bengourley:
⚠️ Note this tips us over our perf budget of 10kB minified/gzipped ⚠️ 

I anticipate when Babel 7 comes out we will be able to deduplicate some of bloaty output for classes… [see the bit about "loose mode for classes is a lot looser" in this 7-alpha release](https://github.com/babel/babel/releases/tag/v7.0.0-alpha.15)

I tried some rather unsafe transforms the `bundle-size-optim` branch but I think it'd be safer and better to wait for the new Babel.

bengourley:
Closing because this is now pretty stale.

PR #315, opened 2018-01-17
Typescript example - updated user/metadata config, added session.  

tremlab:
changed the error examples as the unhandled throw() error wasn't providing interesting data for the dashboard.


bengourley:
Tidied and merged here https://github.com/bugsnag/bugsnag-js/commit/8cfa050904f0ea07bf45788ab5529c3919cd44e8

PR #317, opened 2018-01-18
Improve unhandled promise rejection support

bengourley:
- extract errors from where bluebird puts them
- serialize rejection reasons better (null, undefined, DOMException)
- Don't send a stacktrace if it originates in Bugsnag's code
- Hint to check the promise tab in error message


browser/plugins/test/unhandled-rejection.test.js
```diff
@@ -36,13 +37,49 @@ describe('plugin: unhandled rejection', () => {
...
+          sendReport: (logger, config, payload) => {
+            const report = payload.events[0].toJSON()
+            expect(report.severity).toBe('error')
+            expect(report.unhandled).toBe(true)
+            expect(report.exceptions[0].errorClass).toBe('UnhandledRejection')
+            expect(report.exceptions[0].message).toBe('Rejection reason was not an Error. See "Promise" tab for more detail.')
+            expect(report.severityReason).toEqual({ type: 'unhandledPromiseRejection' })
+            expect(report.metaData.promise['rejection reason']).toEqual({
+              DOMException: {
+                name: 'AbortError',
```
snmaynard: Should errorClass not be AbortError in this case? Message could be message from the DOMException too. I guess you did this to highlight why it has no stacktrace? I've seen a few dashboards that have a lot of UnhandledRejections in them and so anything we can do to distinguish them in the list view would be good.

bengourley: Good point. The only downside I see of this is it would loose the call to action to look at the "promise" tab, which may or may not contain a "code" property too. But having the more descriptive name/message is probs the better tradeoff.

snmaynard: We could put something in the dashboard panel that talks about no stacktraces on javascript projects to explain that and get the best of both worlds

PR #328, opened 2018-02-22
Strip querystring (and fragment) from stackframe.file

bengourley:
Querystring and fragments in `stackframe.file` causes a few issues with sourcemaps and grouping. Removing them resolves this.


README.md
```diff
@@ -1,5 +1,5 @@
 # Bugsnag error reporter for JavaScript
-![10.23kB](https://img.shields.io/badge/size-10.23kB-green.svg)
+![10.29kB](https://img.shields.io/badge/size-10.29kB-green.svg)
```
bengourley: Without 64d618b6380f5338b27c73af07f1dcf5712f0e5e the size became `10.39kB`.

browser/plugins/test/strip-query-string.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+const plugin = require('../strip-query-string')
+
+const Client = require('../../../base/client')
+const VALID_NOTIFIER = { name: 't', version: '0', url: 'http://' }
+
+describe('plugin: strip query string', () => {
+  it('should strip querystrings from urls', () => {
+    expect(
+      plugin._strip('https://bath.tub.xyz/assets/js/app.js?id=49c3ad475f69072745e3')
```
kattrali: Will it strip query strings with multiple items?

```js
plugin._strip('https://bath.tub.xyz/assets/js/app.js?id=49c3ad475f69072745e3&name=bob')
```

browser/plugins/test/strip-query-string.test.js
```diff
@@ -0,0 +1,57 @@
...
+    ).toBe('https://bath.tub.xyz/assets/js/app.js')
+    expect(
+      plugin._strip('http://bath.tub.xyz/assets/js/app.js#nmfrz')
+    ).toBe('http://bath.tub.xyz/assets/js/app.js')
+    expect(
+      plugin._strip('http://bath.tub.xyz:8808/assets/js/app.js#nmfrz')
```
kattrali: Will it strip URLs with both query strings and fragments?

browser/index.js
```diff
@@ -3,25 +3,40 @@ const version = '__VERSION__'
...
+const pluginContext = require('./plugins/context')
+const pluginRequest = require('./plugins/request')
+const pluginThrottle = require('../base/plugins/throttle')
+const pluginConsoleBreadcrumbs = require('./plugins/console-breadcrumbs')
+const pluginNavigationBreadcrumbs = require('./plugins/navigation-breadcrumbs')
+const pluginInteractionBreadcrumbs = require('./plugins/interaction-breadcrumbs')
+const pluginInlineScriptContent = require('./plugins/inline-script-content')
+const pluginSessions = require('./plugins/sessions')
+const pluginIp = require('./plugins/ip')
+const pluginStripQueryString = require('./plugins/strip-query-string')
```
kattrali: This seems like a sensible change - plugins had names purely for the sake of fetching them out by name.

browser/plugins/test/strip-query-string.test.js
bengourley: It will but there isn't a test. I'll add one.

browser/index.js
bengourley: Yeah, they needed to be accessible by name and also enumerable. But the pretty object keys wouldn't get compressed, so I realised I could just define them all as variables and then stick them in an array for better results.

Issue #324, opened 2018-02-05
Performance issues when logging large amounts of data


tur-nr:
### Expected behavior
No bottlenecks whenever bugsnag is leaving breadcrumbs.

### Observed behavior
Bugsnag has a huge impact on performance whenever logging large amounts of data to the console. _Development only._

![image](https://user-images.githubusercontent.com/817134/35787795-ea6c7dd0-0a84-11e8-9192-8a496fe63687.png)

![image](https://user-images.githubusercontent.com/817134/35787813-0afc48f0-0a85-11e8-9b97-0a87fee64a4a.png)

### Steps to reproduce
1. Add bugsnag to an application with `{ consoleBreadcrumbsEnabled: true }`.
2. Open performance tab in Chrome.
3. Record.
4. Do tasks in Application that log a decent set of data.
5. Stop recording.
6. Review performance.

### Version
v4.2.0

### Additional information
Just to give your team some context. We are developing quite a large scale application that handles large amounts of data sets. During development we log these states to the console for review after certain actions. So we can follow the applications state pre/post action.

Our setup is your standard React + Redux. However we use Redux Logger that subscribes to the store's actions and state changes, logging them to the console.

We recently upgrade to v4 from v3 and found that parts of our application where extremely sluggish during development. Upon giving it a performance review we noticed the stringify and leave breadcrumbs functions where very time consuming due to application state being logged.

bengourley:
Hey @tur-nr thanks for the report. In development you should switch off the `autoBreadcrumbsEnabled` option if you're logging large objects. We need to stringify the objects in order to test for duplicates and things like that, so it's just best to only enable that option in environments where you're not doing tonnes of logging.

bengourley:
Hey @tur-nr. I'd still recommend switching off `consoleBreadcrumbsEnabled` if you know you're going to be logging large objects frequently. However, try the latest release v4.3.1. We've optimised the breadcrumb process to avoid perf issues as much as possible. Let us know how you get on! Cheers.

PR #329, opened 2018-03-02
Console breadcrumb perf

bengourley:
This PR is attempting to addresses #324 as best as we can.

---

### Part A: Optimize console breadcrumb collection

The function that intercepts `console[method]` calls and leaves a breadcrumb can be optimized for better performance:

https://github.com/bugsnag/bugsnag-js/blob/cce080ce30f99d7a78b4099f34357376c47c0bf7/browser/plugins/console-breadcrumbs.js#L10-L25

The various ideas for reducing the amount of work this function does are as follows:

- Avoiding calling `.join()`. On large strings this probably takes a big hit.
- Always do `JSON.stringify(arg)` and then only fall back to `String(arg)` when that errors. This avoids a double-hit on big objects.
- Skip the `spaces` (`2`) argument in `JSON.stringify()`
- Swap the regex test for an `indexOf()` call

These ideas were codified into a [benchmark](https://benchmarkjs.com/) suite for comparison.

<details>
<summary>Expand for the full benchmark suite</summary>

```html
<!DOCTYPE html>
<html>
  <head>
    <title>This is a title</title>
    <script src="/node_modules/lodash/lodash.js"></script>
    <script src="/node_modules/platform/platform.js"></script>
    <script src="/node_modules/benchmark/benchmark.js"></script>
    <script>
    var jsonExample = { /* omitted for brevity. Insert some large JSON here */ }
  </script>
  </head>
  <body>
    <p>Open console for benchmark results</p>
    <script>
    // Array#reduce
    const reduce = (arr, fn, accum) => {
      let val = accum
      for (let i = 0, len = arr.length; i < len; i++) val = fn(val, arr[i], i, arr)
      return val
    }

    // Array#filter
    const filter = (arr, fn) =>
      reduce(arr, (accum, item, i, arr) => !fn(item, i, arr) ? accum : accum.concat(item), [])

    // Array#map
    const map = (arr, fn) =>
      reduce(arr, (accum, item, i, arr) => accum.concat(fn(item, i, arr)), [])
    </script>
    <script>
      var method = 'info'
      function current(args) {
        return {
          severity: /^group/.test(method) ? 'log' : method,
          message: map(args, arg => {
            // do the best/simplest stringification of each argument
            let stringified = String(arg)
            // unless it stringifies to [object Object], use the toString() value
            if (stringified !== '[object Object]') return stringified
            // otherwise attempt to JSON stringify (with indents/spaces)
            try { stringified = JSON.stringify(arg, null, 2) } catch (e) {}
            // any errors, fallback to [object Object]
            return stringified
          }).join('\n')
        }
      }
      function noJoin(args) {
        return {
          severity: /^group/.test(method) ? 'log' : method,
          ...reduce(args, (accum, arg, i) => {
            // do the best/simplest stringification of each argument
            let stringified = String(arg)
            // unless it stringifies to [object Object], use the toString() value
            if (stringified === '[object Object]') {
              try { stringified = JSON.stringify(arg, null, 2) } catch (e) {}
            }
            // any errors, fallback to [object Object]
            accum[`arguments[${i}]`] = stringified
            return accum
          }, {})
        }
      }
      function optimisticStringify(args) {
        return {
          severity: /^group/.test(method) ? 'log' : method,
          message: map(args, (arg, i) => {
            let stringified
            try {
              stringified = JSON.stringify(arg, null, 2)
            } catch (e) {
              stringified = String(arg)
            }
            return stringified
          }).join('\n')
        }
      }
      function noJsonSpaces(args) {
        return {
          severity: /^group/.test(method) ? 'log' : method,
          message: map(args, arg => {
            // do the best/simplest stringification of each argument
            let stringified = String(arg)
            // unless it stringifies to [object Object], use the toString() value
            if (stringified !== '[object Object]') return stringified
            // otherwise attempt to JSON stringify (with indents/spaces)
            try { stringified = JSON.stringify(arg) } catch (e) {}
            // any errors, fallback to [object Object]
            return stringified
          }).join('\n')
        }
      }
      function noRegex(args) {
        return {
          severity: method.indexOf('group') === 0 ? 'log' : method,
          message: map(args, arg => {
            // do the best/simplest stringification of each argument
            let stringified = String(arg)
            // unless it stringifies to [object Object], use the toString() value
            if (stringified !== '[object Object]') return stringified
            // otherwise attempt to JSON stringify (with indents/spaces)
            try { stringified = JSON.stringify(arg, null, 2) } catch (e) {}
            // any errors, fallback to [object Object]
            return stringified
          }).join('\n')
        }
      }
      function best(args) {
        return {
          severity: method.indexOf('group') === 0 ? 'log' : method,
          ...reduce(args, (accum, arg, i) => {
            // do the best/simplest stringification of each argument
            let stringified = String(arg)
            // unless it stringifies to [object Object], use the toString() value
            if (stringified === '[object Object]') {
              try {
                stringified = JSON.stringify(arg/*, (k, v) => {
                  if (v === undefined) return 'undefined'
                  if (typeof v === 'function') return '[function]'
                  return v
                }*/)
              } catch (e) {}
            }
            accum[`arguments[${i}]`] = stringified
            return accum
          }, {})
        }
      }
      function nostringify(args) {
        return {
          severity: /^group/.test(method) ? 'log' : method,
          ...reduce(args, (accum, arg, i) => {
            accum[`argument[${i}]`] = arg
            return accum
          }, {})
        }
      }
    </script>
    <script>
      var suite = new Benchmark.Suite;

      function run(fn) {
        fn(['a', 'b', 'c'])
        fn({
          nested: [ 'obj', 'ects'],
          val: 102020404,
          str: '123412312312312'
        })
        fn('loggly message')
        fn(jsonExample)
        fn(jsonExample, jsonExample, jsonExample)
      }

      suite
        // add tests
        .add('current', function() {
          run(current)
        })
        .add('noJoin', function() {
          run(noJoin)
        })
        .add('optimisticStringify', function() {
          run(optimisticStringify)
        })
        .add('noJsonSpaces', function() {
          run(noJsonSpaces)
        })
        .add('noRegex', function() {
          run(noRegex)
        })
        .add('best?', function() {
          run(best)
        })
        .add('nostringify', function() {
          run(best)
        })
        // add listeners
        .on('cycle', function(event) {
          console.log(String(event.target));
        })
        .on('error', console.error)
        .on('complete', function() {
          console.log('done')
          // console.log(this);
          // console.log('Fastest is ' + this.filter('fastest').map('name'));
        })
        // run async
        .run({ 'async': true });
    </script>
  </body>
</html>
```
</details>

----

The results were as follows:

```
current x 123,886 ops/sec ±2.22% (48 runs sampled)
noJoin x 139,807 ops/sec ±1.68% (45 runs sampled)
optimisticStringify x 66,106 ops/sec ±2.32% (53 runs sampled)
noJsonSpaces x 128,184 ops/sec ±1.34% (55 runs sampled)
noRegex x 126,763 ops/sec ±1.83% (52 runs sampled)
best? x 144,012 ops/sec ±3.13% (46 runs sampled)
nostringify x 147,995 ops/sec ±2.17% (50 runs sampled)
```

`best?` was a combination of all the ideas which performed better than `current`. And `nostringify` did nothing at all, to get an idea of the performance "ceiling". So the only idea that didn't prove more performant was the optimistic stringify.

As a result the "best" implementation was adopted and applied in this PR. The perf gain seems definite but still within the same order of magnitude: __1.17x faster__.

### Part B: Optimize breadcrumb deduplication

The function that stores a breadcrumb results in duplicate serialization when it checks for duplicates. The result of this serialization could be cached so that it doesn't result in duplicate work every time a breadcrumb is left.

https://github.com/bugsnag/bugsnag-js/blob/cce080ce30f99d7a78b4099f34357376c47c0bf7/base/client.js#L99-L126

It turns out that:

- no other notifiers do duplicate breadcrumb checks
- the `maxBreadcrumbs` option prevents unlimited duplicates from being added
- the likelihood of a duplicate ever being found was rare – they would have to happen in the same millisecond

Some benchmarks were run (benchmark code not included for brevity):

```
current x 631 ops/sec ±2.30% (78 runs sampled)
no dupe checking x 18,441 ops/sec ±4.61% (79 runs sampled)
no length checking x 651 ops/sec ±1.69% (81 runs sampled)
direct assign (not push) x 631 ops/sec ±2.37% (81 runs sampled)
length management, copy not slice() x 649 ops/sec ±2.37% (81 runs sampled)
no dupe check x 19,298 ops/sec ±3.11% (82 runs sampled)
```

Quite unsurprisingly, not doing the duplicate checking was a huge perf gain. Roughly __30x faster__.

As a result, this PR removes the duplicate checking. Breadcrumbs are now only serialized once – when they are sent in an error report.


browser/plugins/test/console-breadcrumbs.test.js
```diff
@@ -18,7 +18,20 @@ describe('plugin: console breadcrumbs', () => {
...
       // make sure it's null-safe
       console.log(null)
-      expect(c.breadcrumbs.length).toBe(2)
+      console.log({
+        foo: [ 1, 2, 3, 'four' ]
+      }, {
+        pets: {
+          cat: 'scratcher',
+          dog: 'pupper',
+          rabbit: 'sniffer'
```
kattrali: 🐱 🐶 🐰

PR #331, opened 2018-03-09
fix(deps): Upgrade cuid dependency to fix rollup bundling issue

bengourley:
The version of cuid used did not play well with rollup's bundling, due to this issue:
https://github.com/ericelliott/cuid/issues/41. When cuid was added to bugsnag, an update wasn't
available, but v2.x was published in Jan 2018 which fixes this issue. All we needed to do is
upgrade the dependency. Unfortunately the updates broke for ie8 in the process (which we need
to support) so we depend on our forked version for now.


browser/plugins/test/sessions.test.js
```diff
@@ -17,6 +17,7 @@ describe('plugin: sessions', () => {
         expect(session.notifier).toEqual(VALID_NOTIFIER)
         expect(session.sessions.length).toBe(1)
         expect(session.sessions[0].id).toBeTruthy()
+        expect(session.sessions[0].id.length).toBeGreaterThan(10)
```
bengourley: Added this extra assertion to ensure that ids at least 10 chars are generated.

Cawllec:
Works as a fix - approval depends on travis

PR #332, opened 2018-03-12
Switch (u)uid implementation

bengourley:
`cuid()` was great but it caused problems when bundling with rollup (#331). Newer versions of cuid that have been published fix the rollup issues but it [no longer supports IE8](https://github.com/ericelliott/cuid/pull/99) (which we need). The maintainers didn't want to merge our PR.

Our options were to

1. maintain a fork of cuid with the minor change needed to support IE8
2. find a different uid implementation

Every alternative implementation I tried added bytes to our minified bundle, but I was focused on reducing bundle size if possible. One of the benefits of (1.) would have been that we could remove the `.slug()` feature of that library, which we didn't use. However, the smallest option I tried was this [uuid v4](https://gist.github.com/LeverOne/1308368) implementation. It wasn't on npm so I added it directly to our codebase.

This shaves .23kB from the bundle size, and brings us back down to the "budget" of 10kB ⚖️.

Now raise my concerns…
- we currently use the id generator for unique session ids
- the pipeline does not currently even use the ids themselves
- how likely is a collision with this new implementation?
- what are the ramifications of having a collision now, and in the future?

In the tests that I added, I run 100k iterations to check for collisions. I found this to be the largest number that meant the unit tests terminate within a few seconds. I bumped the value up to 1M and ran the tests multiple times without collision on my machine. However, the vector most likeley to cause collisions probably isn't running the same algorithm sequentially on one machine, it's probably some freak occurrence of two machines on opposite sides of the world having the same series of `Math.random()` results.

On balance, I think the perf gain from a smaller bundle outweighs these issues. But my take is that we should bear them in mind – especially when the pipeline starts using the session ids.


ericelliott:
In old versions of Chrome, that v4 UUID algorithm has a disturbingly high collision rate, even on one machine.

The issue was [patched in 2015](https://v8project.blogspot.com/2015/12/theres-mathrandom-and-then-theres.html), but if you're supporting old environments, [you should be aware of it](https://medium.com/@betable/tifu-by-using-math-random-f1c308c4fd9d), and you should double check that the vulnerability does not exist in any of your supported environments.

Even with a better implementation, you'll still run into [birthday paradox problems](https://en.wikipedia.org/wiki/Birthday_problem) if you add distributed id generation to the mix, because as you add id generating hosts, you lose entropy exponentially for simultaneously generated ids (PSRNGs typically use system time as a random seed - as the number of simultaneous hosts climbs linearly, chances of collisions rise exponentially).

The question you need to ask yourselves is, "what are the consequences of a collision?"

Are those consequences easier to deal with that maintaining a fork of something like cuid?

bengourley:
Thanks for your input @ericelliott, it's really helpful.

We had a bunch of internal discussion around this yesterday too – raising similar issues. Thankfully we don't have to worry so much about people stuck on old versions of "evergreen" browsers, since the vast majority get auto-upgraded. I can imagine the `Math.random()` implementation in IE8/9 having some flaws in it too, though.

The distributed hosts problem is what I was hinting at when I mentioned that the lack of collisions in the tests were probably not a good indicator of the likelihood of collisions in the real world.

In the end we figured that the consequences are not bad enough to spend more time worrying about collisions at this point. As I mentioned, our pipeline doesn't actually use the id that we generate right now… it's there to future-proof the API and to support future features. Before implementing anything that relies on uniqueness we could analyse the incoming ids to determine how regular collisions are with this implementation.

I appreciate you taking the time to discuss these issues here and also for producing `cuid` in the first place, it's a great approach and I would use it in a snap if the situation allowed 👍

bengourley:
Closing in favour of #331.

PR #333, opened 2018-03-15
feat(transports): Default to https:// for error/session reporting

bengourley:
The default configuration was a protocol relative url "//…" which would match the protocol of the page. This was to support IE8/9, as they would only send XDomain POST requests to a URL on the same protocol as the current page (see point 7 [here](https://blogs.msdn.microsoft.com/ieinternals/2010/05/13/xdomainrequest-restrictions-limitations-and-workarounds/)). However the relative protocol breaks in a couple of places, namely developing locally on a file:// url and in Electron development.

This change means that http will only ever be used in IE8/9 when the protocol of the page is http, OR if the library is manually configured to point to a non-ssl endpoint.


bengourley:
Regarding this code climate failure – it's complaining about similar code in `browser/transports/x-domain-request.js`

I tried factoring out the common parts into a function which would make codeclimate happy – however, it increased the bundle size – presumably because the duplication actually gzips well! I think the extracting into a function in this instance makes the code harder to follow. I'm happy to keep it as-is and ignore this warning if that seems reasonable to everybody else?

kattrali:
Yeah this looks like a false positive - seems better to mark it as invalid on code climate rather than build something convoluted to handle it.

PR #334, opened 2018-03-16
Add network breadcrumbs plugin

wordofchristian:
This adds a breadcrumb plugin for activity from XMLHTTPRequests and window.fetch. This should cover all major javascript networking libraries. The approach has been specifically tested with [jQuery](http://api.jquery.com/), [axios](https://github.com/axios/axios), [superagent](https://github.com/visionmedia/superagent), and the [native fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)

## Breadcrumbs
For XMLHTTPRrequest, three possible breadcrumbs are triggered:

**XMLHTTPRequest succeeded**

- Happens when when a XMLHTTPRequest completes successfully.
- contains the following metadata `{ request: 'GET https://some.url', status: 200 }`

**XMLHTTPRequest failed**

- Happens when when a XMLHTTPRequest completes with a response status of 4xx or 5xx.
- contains the following metadata `{ request: 'GET https://some.url', status: 404 }`

**XMLHTTPRequest error**

- Happens when when a XMLHTTPRequest could not complete (e.g. due to a CORS error).
- contains the following metadata `{ request: 'GET https://some.url' }`

For window.fetch requests the following breadcrumbs are triggered:

**fetch() succeeded**

- Happens when when a fetch request completes successfully.
- contains the following metadata `{ request: 'GET https://some.url', status: 200 }`

**fetch() failed**

- Happens when when a fetch request completes with a response status of 4xx or 5xx.
- contains the following metadata `{ request: 'GET https://some.url', status: 404 }`

**fetch() error**

- Happens when when a fetch request could not complete (e.g. due to a CORS error).
- contains the following metadata `{ request: 'GET https://some.url' }`

## Support

Does not support IE8 because of lack of addEventListener. Does not support XDomainRequest for the same reason. I attempted to accomplish the same functionality by falling back to onload and onerror, but there was not a good way to prevent duplicate breadcrumbs and make sure we didn't destroy the user's onload/onerror callbacks.

## TODO

- [x] Finish writing integration tests
- [x] Ensure that a breadcrumb is not left by Bugsnag.notify() network request


browser/plugins/network-breadcrumbs.js
```diff
@@ -0,0 +1,144 @@
...
+ * Leaves breadcrumbs when network requests occur
+ */
+module.exports = {
+  init: (_client) => {
+    client = _client
+    monkeyPatchXMLHttpRequest()
+    monkeyPatchFetch()
+  },
+  destroy: () => {
+    restoreFunctions.forEach(restore => restore())
```
bengourley: I really want to find a better way of doing this. We basically only need these destroy methods for tests, so I want a way to have them compile out of the production build. Don't worry about addressing this I just wanted to raise it as it's been bugging be for a bit.

browser/plugins/network-breadcrumbs.js
```diff
@@ -0,0 +1,144 @@
+/* global XMLHttpRequest */
+const breadcrumbType = 'network'
+let restoreFunctions = []
```
bengourley: this can be `const`

browser/plugins/network-breadcrumbs.js
```diff
@@ -0,0 +1,144 @@
...
+// ------------------------------------------------------------------------------------------------
+// create keys to safely store metadata on the request object
+const requestSetupKey = 'BUGSNAG:SETUP'
+const requestUrlKey = 'BUGSNAG:REQUEST_URL'
+const requestMethodKey = 'BUGSNAG:REQUEST_METHOD'
+
+// generic function for monkey patching both XMLHttpRequest and XDomainRequest
+function monkeyPatchXMLHttpRequest () {
+  if (!('addEventListener' in XMLHttpRequest.prototype)) return
+  let nativeOpen = XMLHttpRequest.prototype.open
```
bengourley: this can be `const`

browser/plugins/network-breadcrumbs.js
```diff
@@ -0,0 +1,144 @@
...
+
+  // override native open()
+  XMLHttpRequest.prototype.open = function open (method, url) {
+    // store url and HTTP method for later
+    this[requestUrlKey] = url
+    this[requestMethodKey] = method
+
+    // if we have already setup listeners, it means open() was called twice, we need to remove
```
bengourley: how did you discover this edge case?

browser/plugins/network-breadcrumbs.js
```diff
@@ -0,0 +1,144 @@
...
+    XMLHttpRequest.prototype.open = nativeOpen
+  })
+}
+
+function handleXHRLoad () {
+  if (this[requestUrlKey] === client.notifier.url) {
+    // don't leave a network breadcrumb from bugsnag notify calls
+    return
+  }
+  let metaData = {
```
bengourley: This can be `const`

browser/plugins/network-breadcrumbs.js
```diff
@@ -0,0 +1,144 @@
...
+  if (this.status >= 400) {
+    // contacted server but got an error response
+    client.leaveBreadcrumb('XMLHttpRequest failed', metaData, breadcrumbType)
+  } else {
+    client.leaveBreadcrumb('XMLHttpRequest succeeded', metaData, breadcrumbType)
+  }
+}
+
+function handleXHRError () {
+  if (this[requestUrlKey] === client.notifier.url) {
```
bengourley: I think you want `client.config.endpoint`

`client.notifier.url='https://github.com/bugsnag/bugsnag-js'` is the github url for the notifier

Additionally you'll want `client.config.sessionEndpoint` if session tracking is enabled.

browser/plugins/test/network-breadcrumbs.test.js
```diff
@@ -0,0 +1,171 @@
...
+          })
+
+          request.send()
+        })
+      })
+
+      it('should not leave a breadcrumb for request to bugsnag notify endpoint', (done) => {
+        setup((client) => {
+          const request = new XMLHttpRequest()
+          request.open('GET', VALID_NOTIFIER.url)
```
bengourley: this needs to be updated to use `config.endpoint`/`config.sessionEndpoint`

browser/plugins/test/network-breadcrumbs.test.js
```diff
@@ -0,0 +1,171 @@
...
+              request: 'GET /'
+            }
+          }))
+          done()
+        })
+      })
+    })
+    it('should leave a breadcrumb when a fetch() has a failed response', (done) => {
+      setup((client) => {
+        global.fetch('http://jsonplaceholder.typicode.com/posts/asdf').then(() => {
```
bengourley: does this test depend on http://jsonplaceholder.typicode.com/posts/asdf being available?

browser/plugins/test/network-breadcrumbs.test.js
```diff
@@ -0,0 +1,171 @@
...
+      })
+    })
+
+    // Browsers that support XDomainRequest will throw a real error when trying to do a cross
+    // domain request. We don't need a breadcrumb for this since bugsnag will track that error
+    if (!XDomainRequest) {
+      it('should leave a breadcrumb when an XMLHTTPRequest has a network error', (done) => {
+        setup((client) => {
+          const request = new XMLHttpRequest()
+          request.open('GET', 'https://api.bugsnag.com')
```
bengourley: does this test rely on https://api.bugsnag.com being accessible and responding correctly every time?

browser/plugins/network-breadcrumbs.js
wordofchristian: I tested this behavior in the browser console 
demo

browser/plugins/test/network-breadcrumbs.test.js
wordofchristian: This test assumes that the request will fail because of a CORS error

browser/plugins/test/network-breadcrumbs.test.js
wordofchristian: This test assumes that that endpoint will be reachable and will return a 404.

PR #336, opened 2018-04-03
Network breadcrumbs tweaks

bengourley:
@wordofchristian I wasn't sure if you were working on the `network-breadcrumbs` branch anymore, so I made these changes here just in case to avoid conflict.

in no particular order…

- fixed the tests
- refactored the network breadcrumb tests to use jasmine's `beforeEach()`/`afterEach()` utilities. Avoids some boilerplate and prevents a race-condition with async tests and the `destroy()` function being called _before_ the tests end.
- made the tests not rely on the network and responses from web servers on the internet
- some style nits, including removing reference to `XDomainRequest` which doesn't get monkeypatched due to the issues mentioned
- swapped remaining `client.notifier.url` reference
- added envify to make the test-only `plugin.destroy()` functions get compiled out for the production bundle


wordofchristian:
👍

PR #340, opened 2018-04-19
feat(logger): Allow logger to be set in config

### LINKED ISSUE #302: Add configuration option to disable console debugging
### Expected behavior
To be able to configure the console debugging verbosity

### Observed behavior
In the console, `[bugsnag] Loaded` is printed during every app load.  Consequently the console output in our end to end testing suite gets filled with those messages.

### Steps to reproduce
Load an application with bugsnag in the browswer, and notice in the console that `[bugsnag] Loaded!` is printed there

### Version
4.0.3

### Additional information
[This](https://github.com/bugsnag/bugsnag-js/blob/781fe3d7d9702466c0ef34b91e97bc6735986f7f/base/client.js#L67) seems to be the line in question.

This is really just an issue for us in our testing environment, so being able to configure the verbosity level, or suppressing that one line would be great.  Thanks!

bengourley:
Previously configuration would happen _before_ the client.logger() method could be called, so the "[bugsnag] loaded!" message would be printed before the logging could be disabled. This commit allows the logger to be set in the configuration object. Logging can be completely disabled by passing `logger: null`.

Fixes #302


bengourley:
The reason the error messages can have the property name removed is because we build a message containing the key already in `configure()`:

https://github.com/bugsnag/bugsnag-js/blob/4132b433796f39415b603cb71d68eb2e783a2728/base/client.js#L58

browser/index.js
```diff
@@ -129,18 +123,6 @@ module.exports = (opts, userPlugins = []) => {
     : bugsnag
 }
 
-const getPrefixedConsole = () => {
```
Cawllec: This is only available/used inside the `browser/index.js` exported function correct?
So it's removal shouldn't be a breaking change?

bengourley: Correct, it was "private" to that module. It's been moved to `browser/config.js` where it is now used.

PR #341, opened 2018-04-23
Enable session tracking by default

bengourley:
Enables session tracking by default, and makes some requisite changes to ensure the relevant endpoints are properly configured. The options `endpoint` and `sessionEndpoint` are soft-deprecated, and they are replaced by the `endpoints: { notify, sessions }` object.

Previously the validation framework was context-free – each value was validated indpendently. Now, some options need to know the state of other options when validating and setting defaults, so this has been updated accordingly.

----

Running through the following scenerios and asserting the notifier yields sensible behaviour:

| Verified | Description           | Config          | Expected behaviour |
| -------- | -------------- |-------------------------| --------------------------|
| ✅| basic, only api key set | `{ apiKey: 'API_KEY' }` | no warnings, sessions sent|
|✅| setting both deprecated options | `{ apiKey: 'API_KEY', endpoint: '/foo', sessionEndpoint: '/bar }`| warning for deprecated options, migrate to current config, sessions sent |
|✅| setting only deprecated endpoint | `{ apiKey: 'FOOBAR', endpoint: '/foo' }`| warning for deprecated options, warning for missing sessions endpoint, sessions not sent |
|✅| setting both new options correctly | `{ apiKey: 'FOOBAR', endpoints: { notify: '/foo', sessions: '/bar' } }`| no warnings, sessions sent |
|✅| typo/unrecognised key in endpoints | `{ apiKey: 'FOOBAR', endpoints: { notify: 'abc', sossiens: '/bar' } }`| config validation error thrown |
|✅| notify endpoint configured, session endpoint missing, auto session tracking off | `{ apiKey: 'FOOBAR', endpoints: { notify: '/foo' }, autoCaptureSessions: false }`| warning session endpoint is not set, sessions not sent |
|✅| as above but manual `startSession()` called | `{ apiKey: 'FOOBAR', endpoints: { notify: '/foo' }, autoCaptureSessions: false }`| warning session endpoint is not set, warning that session is not sent |
|✅| session tracking explicitly on and endpoints configured but session endpoint configuration missing | `{ apiKey: 'FOOBAR', endpoints: { notify: '/foo' }, autoCaptureSessions: true }`| config validation error thrown |
|✅| endpoints configured but notify endpoint configuration missing | `{ apiKey: 'FOOBAR', endpoints: { sessions: '/bar' } }`| config validation error thrown |


bengourley:
Additionally I did a review of the scenarios with a developer not on the platforms team:

```
    Each example page listed below configures the bugsnag js client slightly differently.
    The configuration is shown in the body of the page. Any warnings/errors will show
    in the console.

    Here is some info you need to know:

    - If you don't configure endpoints, the notifier will send error reports and
      sessions to our normal APIs (*.bugsnag.com).
    - Sessions are enabled by default.

    Navigate to each page with your console open and answer the following questions,
    based on what you see:

    - did bugsnag start ok?
    - was the configuration valid?
    - were any deprecated options used?
    - will it send…
      - error reports?
      - sessions?
```

They were able to answer each question correctly for each scenario, with no context other than the code that configured the client and the console output.

The update 93e7cfb88b3195f2f7c7f6ccc0afeb9f7dd12ef9 came out of this review.

.codeclimate.yml
```diff
@@ -5,6 +5,9 @@ checks:
       threshold: 7
   return-statements:
     enabled: false
+  method-lines:
+    config:
+      threshold: 30
```
fractalwrench: What was the previous threshold - has this been increased or decreased?

base/client.js
```diff
@@ -97,7 +95,7 @@ class BugsnagClient {
   }
 
   leaveBreadcrumb (name, metaData, type, timestamp) {
-    if (!this._configured) throw new Error('Bugsnag must be configured before calling leaveBreadcrumb()')
+    if (!this._configured) throw new Error('client not configured')
```
fractalwrench: If I understand things correctly, this change isn't necessarily required for the on-by-default session tracking? The change itself is fine, but might be better in a separate PR in future to make review easier

base/client.js
```diff
@@ -227,8 +227,9 @@ const normaliseError = (error, opts, logger) => {
       err = new Error(String(error))
       break
     case 'function':
-      logger.warn(`Usage error. notify() called with a function as "error" parameter`)
-      err = new Error('Bugsnag usage error. notify() called with a function as "error" parameter')
+      const msg = 'notify() called with a function'
```
fractalwrench: There are a few variations on "called with" (a, an, etc) here. If space is a real issue, then using something in the following format _might_ help reduce the gzipped size, and also state the expected parameter type: "notify() expected error/opts parameters, received: $badThing"

base/client.js
```diff
@@ -250,4 +252,9 @@ const hasNecessaryFields = error =>
   (typeof error.name === 'string' || typeof error.errorClass === 'string') &&
   (typeof error.message === 'string' || typeof error.errorMessage === 'string')
 
+const generateConfigErrorMessage = errors =>
+`Bugsnag configuration error\n${map(errors, (err) => `"${err.key}" ${err.message} \n    got ${stringify(err.value)}`).join('\n\n')}`
```
fractalwrench: If you're golfing, there's a bit of whitespace in this message, if not, this is fine :)

base/config.js
```diff
@@ -22,18 +22,26 @@ module.exports.schema = {
...
   },
-  endpoint: {
-    defaultValue: () => 'https://notify.bugsnag.com',
-    message: 'should be a URL',
-    validate: () => true
-  },
-  sessionEndpoint: {
-    defaultValue: () => 'https://sessions.bugsnag.com',
-    message: 'should be a URL',
-    validate: () => true
```
fractalwrench: Shouldn't `endpoint` and `sessionEndpoint` remain on Configuration, for backwards compatibility?

base/config.js
```diff
@@ -22,18 +22,26 @@ module.exports.schema = {
...
+        stringWithLength(val.notify) &&
+        // endpoints.sessions must be set unless session tracking is explicitly off
+        (obj.autoCaptureSessions === false || stringWithLength(val.sessions))
+      ) &&
+      // ensure no keys other than notify/session are set on endpoints object
+      filter(keys(val), k => !includes([ 'notify', 'sessions' ], k)).length === 0
   },
   autoCaptureSessions: {
-    defaultValue: () => false,
+    defaultValue: (val, opts) => opts.endpoints === undefined || (!!opts.endpoints && !!opts.endpoints.sessions),
```
fractalwrench: This is just my curious question about JS syntax - what does `!!` do?

base/config.js
```diff
@@ -81,17 +89,17 @@ module.exports.schema = {
 }
 
 module.exports.mergeDefaults = (opts, schema) => {
-  if (!opts || !schema) throw new Error('schema.mergeDefaults(opts, schema): opts and schema objects are required')
+  if (!opts || !schema) throw new Error('opts and schema objects are required')
   return reduce(keys(schema), (accum, key) => {
-    accum[key] = opts[key] !== undefined ? opts[key] : schema[key].defaultValue()
+    accum[key] = opts[key] !== undefined ? opts[key] : schema[key].defaultValue(opts[key], opts)
```
fractalwrench: I don't understand the context to this change (although that may just be my rusty JS) - is this part of the session tracking implementation?

browser/index.js
```diff
@@ -66,19 +78,13 @@ module.exports = (opts, userPlugins = []) => {
...
-    // errors can be thrown here that prevent the lib from being in a useable state
-    bugsnag.configure(opts)
-  } catch (e) {
-    bugsnag._logger.warn(e)
-    if (e.errors) map(e.errors, bugsnag._logger.warn)
-    // rethrow. if there was an error with configuration
-    // the library is not going to work
-    throw e
-  }
+  bugsnag.transport(window.XDomainRequest ? tXDomainRequest : tXMLHttpRequest)
```
fractalwrench: Is this no longer capable of throwing an error?

browser/plugins/inline-script-content.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
       if (!frame || !frame.file || !frame.lineNumber) return frame
       if (frame.file.replace(/#.*$/, '') !== originalLocation.replace(/#.*$/, '')) return frame
       if (!DOMContentLoaded || !html) html = getHtml()
-      const htmlLines = [ '<!-- DOCUMENT START -->' ].concat(html.split('\n'))
+      const htmlLines = [ '<!-- DOC START -->' ].concat(html.split('\n'))
```
fractalwrench: Every little helps 😀

browser/plugins/interaction-breadcrumbs.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
       } catch (e) {
         targetText = '[hidden]'
         targetSelector = '[hidden]'
-        client._logger.error('Cross domain error when tracking click event. See https://docs.bugsnag.com/platforms/browsers/faq/#3-cross-origin-script-errors')
+        client._logger.error('Cross domain error when tracking click event. See https://tinyurl.com/y94fq5zm')
```
fractalwrench: How many bytes does this save? IMO it's now less clear that this is a docs link.

browser/plugins/sessions.js
```diff
@@ -16,7 +16,12 @@ const sessionDelegate = {
 
     // exit early if the reports should not be sent on the current releaseStage
     if (isArray(sessionClient.config.notifyReleaseStages) && !includes(sessionClient.config.notifyReleaseStages, releaseStage)) {
-      sessionClient._logger.warn(`Session not sent due to releaseStage/notifyReleaseStages configuration`)
+      sessionClient._logger.warn(`Session not sent due to "releaseStage/notifyReleaseStages" configuration`)
+      return sessionClient
+    }
+
+    if (!sessionClient.config.endpoints.sessions) {
+      sessionClient._logger.warn(`Session not sent because "endpoints.sessions" is not configured`)
```
fractalwrench: because -> due to should help golf messages a bit more (if that's a concern)

browser/plugins/test/network-breadcrumbs.test.js
```diff
@@ -10,25 +10,32 @@ const VALID_NOTIFIER = { name: 't', version: '0', url: 'http://' }
...
+    // setup client and plugin
+    client = new Client(VALID_NOTIFIER)
+    try {
+      client.configure({
+        apiKey: 'aaaa-aaaa-aaaa-aaaa',
+        endpoints: {
+          notify: '/echo/reports',
+          sessions: '/echo/sessions'
+        }
+      })
```
fractalwrench: Have we got an automated test to ensure that the deprecated API works as expected? I scanned but couldn't find anything (although that might reflect more on my search skills).

.codeclimate.yml
bengourley: The default threshold was 25 so it has been increased. Number of lines is a bit of a crude gauge of maintainability in my opinion. However it can be a useful indicator of when a function might be getting to big, which is why I didn't switch it off.

The function that caused me to increase this threshold was this one: https://github.com/bugsnag/bugsnag-js/blob/bengourley/sessions-default-on/browser/plugins/sessions.js

Most of the lines (16 of them) are taken up by a declarative object literal. And so rather than battle with this rule, I bumped the threshold to what I consider a more reasonable level.

base/client.js
bengourley: It's in a separate commit with its own message.

The functionality isn't related to the session tracking change, but the changes made the bundle size increase, so this change is required to land at the same time to keep the bundle size reasonable.

base/config.js
bengourley: `!` is the logical _NOT_ operator. `!!` is just the logical _NOT_ operator repeated! 

Because js has the concept of "truthyness", you can use the _NOT_ operator twice on a truthy value to get the value `true`.

for example:

```js
var str = 'hi jamie'
var strExists = !!str //=> strExists = true

var nostr = null
var nostrExists = !!nostr //=> nostrExists = false
```

Why this is useful in this situation… the logical _AND_ operator `&&` will return the result of the right hand side if all the value are truthy:

```js
var opts = { endpoints: { sessions: 'yes' } }
var result = opts.endpoints && opts.endpoints.sessions
//=> result = 'yes'
```

We can use the `!` operator once to coerce the values into booleans, but then it is `false`, so we use it again to flip it back to `true`.

```js
var opts = { endpoints: { sessions: 'yes' } }
var result = !!opts.endpoints && !!opts.endpoints.sessions
//=> result = true
```

Hope this makes sense!

base/config.js
bengourley: The `defaultValue` function previously operated on each schema item entirely independently. Now the default value for `autoCaptureSessions` is dependent on the original config object, so `opts` (the user-provided configuration object) needs to get passed in to the `defaultValue` call.

browser/index.js
bengourley: It throws errors still, and we stop swallowing them. This commit detail should explain things in more detail https://github.com/bugsnag/bugsnag-js/pull/341/commits/5778437e540cc55bfe48ab38f88c2e14f8638bb6.

browser/plugins/interaction-breadcrumbs.js
bengourley: `0.3kB`

browser/plugins/test/network-breadcrumbs.test.js
bengourley: Not automated, no. This is what I went through with Alex/Dave manually. I'll get it set up on maze though.

.codeclimate.yml
fractalwrench: 30 seems reasonable to me, particularly for that example. 👍

bengourley:
Thanks for the review @fractalwrench!

> Were there any plans to add mazerunner scenarios to cover the deprecated and new API?

Yes but maze is still a WIP branch on this repository so this may land before the maze tests do. This is why I put a lot of effort into manually testing all the permutations.

browser/plugins/interaction-breadcrumbs.js
fractalwrench: In that case maybe we could clarify the general format of shortened URLs to something like "See docs: $url"?

anikolaev:
I don't use session tracking. When I set in config:
```
  endpoints: {
    notify: bugsnagConfig.endpoint
  },
  autoCaptureSessions: false,
```
I still get the warning:
```
bugsnag.js?b390:2442 [bugsnag] notify endpoint is set but sessions endpoint is not. No sessions will be sent.
```
To remove the warning I have to add fake `sessions` value:
```
  endpoints: {
    notify: bugsnagConfig.endpoint,
    sessions: '---'
  },
  autoCaptureSessions: false,
```
I believe it should be fixed. Checked with v4.7.2.

bengourley:
This is correct behaviour according to the table of scenarios. However, I'm inclined to agree with you that it shouldn't warn in this case and don't initially see a reason that it can't be changed to do so. Let me get back to you on this.

bengourley:
When we revisited this, we established that there is a good reason for the warning in this scenario. When `autoCaptureSessions: false` you can still manually send a session with `client.startSession()`.

anikolaev:
@bengourley Maybe then it is better to display this warning (or even error) only on client.startSession() call? If I don't care about sessions I still have to use fake session endpoint to suppress this warning.

snmaynard:
You can set a fake sessions endpoint to hide the warning as you said. We don't want to remove this warning or delay it until after boot enterprise customers with an on premise bugsnag instance should set both.

PR #342, opened 2018-04-30
only pass url to original history fn if it's a string

maartenverbaarschot:
Hi :)

At our company we started using Bugsnag last week, it helped us fix quite a few bugs in our code already. But today we discovered a bug that is on your side ;-). It is a quite severe one in Internet Explorer (11) that breaks Vue-Router, making our app as good as unusable for Internet Explorer users.

The symptom is that it immediately redirects our app to '/undefined', from there on the app router will get all confused.

I found out that bugsnag-js wraps window.history functions for breadcrumb purposes. Because of the wrapper, when calling replaceState without a URL, Internet Explorer will convert the URL to the string `'undefined'`. This can be solved by only passing the URL parameter to the original function it it's a string. 

There was a similar issue with Google Tag Manager some time ago, see this comment in another project for background: https://github.com/vuejs/vue-router/issues/2023#issuecomment-364156931


bengourley:
Landed here https://github.com/bugsnag/bugsnag-js/commit/cb807890a408b22742129370765430f436004867

PR #345, opened 2018-05-03
fix(plugins): Prevent crash in IE10 when accessing history.state

### LINKED ISSUE #343: Crash when accessing history.state on IE10
### Expected behavior

No crash.

### Observed behavior

IE10 throws an 'Unspecified error' when bugsnag-js tries to access window.history.state.

### Steps to reproduce

I can't exactly tell how to reproduce this. In my case, it happens when I open the page, and then do 'refresh'.

### Version

bugsnag-js 4.5.0

### Additional information

Searching for this issue brings up other repositories the same issue, e.g. https://github.com/angular/angular.js/issues/10367 I've seen this being fixed by just wrapping the read of history.state in a try {} catch ().

(e.g. https://github.com/netman92/angular.js/commit/3bd03ff56f898cddc078791b85a127fafad3ec3c )

bengourley:
Users of various js libraries have reported occasional accounts of "Unspecified error" when
window.history.state is accessed in IE10. This change wraps the two places we access this value with
a try/catch which swallows the error (the same approach that angular.js took to solve this problem
https://github.com/netman92/angular.js/commit/3bd03ff56f898cddc078791b85a127fafad3ec3c).

Fixes #343


kattrali:
How was this change tested?

bengourley:
Other than running the existing test suite, it wasn't.

PR #348, opened 2018-05-08
fix(plugins): Use correct breadcrumb type for network breadcrumbs

bengourley:
The recognised type for this sort of breadcrumb is `"request"`, but until this change network breadcrumbs were sent with `type="network"`. This meant the breadcrumbs were received, but silently dropped by the pipeline and therefore wouldn't show in the dashboard.


bengourley:
Network breadcrumbs appearing in the dashboard with this change:

image

PR #351, opened 2018-05-18
Add Maze Runner end to end test suite

bengourley:
## Goal

This PR adds a new end-to-end test suite using [maze-runner](https://github.com/bugsnag/maze-runner). This tool factors out common tasks and assertion utilities that are needed across our notifier libraries. The main goals are:

- more complete coverage
- commonality amongst notifier testing
- facilitate simple addition of scenarios as they arise

### Added

The entire set of tests, fixtures and support scripts are added in the `features` directory, which is a cucumber/gherkin convention.

Additionally a `Gemfile` and `Gemfile.lock` are added to describe the ruby dependencies.

Tests were added for each of the following cases:

- an app containing bundled bugsnag-js using webpack
- a html page linking dist/bugsnag.js
- a library containing bundled bugsnag-js using rollup
- Syntax errors
- Thrown exceptions
- Malformed URIs
- Unhandled promise rejections
- Invoking an undefined function
- Thrown exception with a malformed stacktrace (TODO)
- Invoking notify from try/catch
- Invoking notify from a Promise.catch()
- Modifying report contents from a per-report callback
- Setting release stage
- Setting user info
- Redacting IP address
- Global callbacks
- Disabling autoNotify

### Removed

The `e2e` directory stands to be removed as it is made redundant by the new tests. At the time of writing, coverage would _decrease_ on the following scenarios, so they need to be addressed with maze tests before `e2e` can sensibly be removed:

- [x] [Sending error reports from a TypeScript app](https://github.com/bugsnag/bugsnag-js/tree/6f13b49b4e93fb107ebd99adf35e1e674b6ef800/e2e/typescript)
- [x] [Successfully detecting inline scripts after a navigation change](https://github.com/bugsnag/bugsnag-js/tree/6f13b49b4e93fb107ebd99adf35e1e674b6ef800/e2e/navigation)
- [x] [Compatibility with bluebird promises](https://github.com/bugsnag/bugsnag-js/tree/6f13b49b4e93fb107ebd99adf35e1e674b6ef800/e2e/bluebird)

### Changed

Travis configuration has been updated and __currently only runs the maze tests__. There is now a build matrix, running a job for each browser to go through all of the end to end scenerios.

- [x] The existing integration tests `browser/**/*.test.js` will need to be re-added to the travis build. The test configuration may need to be updated slightly to run this in each browser ci job.
- [x] The existing unit tests `base/**/*.test.js` will need to be added re-added to the travis build. This could be done with travis's new build stages because it needn't run for every browser.

## Review

- [ ] Commented on code changes inline explain the reasoning behind the approach
- [ ] Reviewed the test cases added for completeness and possible points for discussion
- [ ] Check the scope of the changeset - is everything in the diff required for the pull request?
  - [ ] Initial review of the intended approach, not yet feature complete
  - [ ] Structural review of the classes, functions, and properties modified
  - [x] Final review


fractalwrench:
Is there any way we could split this PR up at all into logical sections? I can have a look through, but I'm not confident that I'd see everything that needs catching in a 2.5k+ changeset.

features/lib/fast-selenium.rb
```diff
@@ -0,0 +1,19 @@
+# copy pasta from https://www.browserstack.com/automate/ruby
```
fractalwrench: Are there any licensing issues around this? It's probably fine, but IANAL

kattrali: Doubtful since it came right out of the docs

features/before_send.feature
```diff
@@ -0,0 +1,66 @@
+@before_send
+Feature: beforeSend callbacks
+
+Scenario Outline: modifying report via beforeSend in config
+  When I navigate to the URL "/before_send/<type>/a.html"
```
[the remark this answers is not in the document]
bengourley: it's a placeholder for this "Scenario Outline". That means that this scenario runs once for every item in the table below it. The `<type>` placeholder is replaced by a different value from the table each time.

features/before_send.feature
```diff
@@ -0,0 +1,66 @@
...
+  And the test should run in this browser
+  And I let the test page run for up to 10 seconds
+  And I wait for 5 seconds
+  Then I should receive 1 request
```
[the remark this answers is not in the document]
bengourley: Sounds like a good idea.

I was also considering removing the "And the test should run in this browser" from all the tests I know I don't need that check. Lots of the fixtures harcode "YES" for this purpose, which is pretty redundant.

bengourley:
> Is there any way we could split this PR up at all into logical sections?

There are individual commits for each set of features that are tested. There is a view in the PR interface where you can look at one commit at a time, e.g: https://github.com/bugsnag/bugsnag-js/pull/351/commits/ac9db8c79803e8baf41abdb3ffd985831ba6619f

.npmignore
```diff
@@ -1,2 +1,3 @@
 coverage
 .nyc_output
+features
```
fractalwrench: Should this also include entries in the .gitignore? i.e. maze_output, others?

.travis.yml
```diff
@@ -2,24 +2,39 @@ sudo: required
...
+    - BROWSER=chrome_43
+    - BROWSER=chrome_61
 addons:
-  chrome: stable
   browserstack:
     username: bugsnagplatforms1
     access_key:
      secure: "lB2aZc0HzuinyhSJyP98lzDJTbZsroEOJTrS/flnLLsHb/C096zjhwkWexHnfmNIPaY5SB2FF0G6zcnknlmsXg5TgxQiK7bdyJmkcAetGi+cSKnGaBq7GZXVhbWh++MRnddQDGV6jrSr7aqrWf85ncq3L4//9BTV6uAhDBWu2bo="
-before_script:
-- curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64
```
fractalwrench: Has this been replaced elsewhere?

features/fixtures/handled/browserify/a.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="dist/a.js"></script>
```
fractalwrench: What do a/b/c stand for in the JS + HTML files - is there a more descriptive naming scheme that would make sense to use?

features/fixtures/handled/browserify/src/a.js
```diff
@@ -0,0 +1,13 @@
...
+var config = require('./lib/config')
+
+var bugsnagClient = bugsnag(config)
+
+bugsnagClient.notify(new Error('bad things'), {
+  beforeSend: function () {
+    setTimeout(function () {
+      var el = document.getElementById('bugsnag-test-state')
+      el.textContent = el.innerText = 'DONE'
+    }, 5000)
```
fractalwrench: It would be good to extract the `5000` from these scripts to one place, in case we decide to change it in future. Is this the delay required to send a report to Bugsnag?

features/fixtures/handled/browserify/src/b.js
```diff
@@ -0,0 +1,17 @@
...
+
+try {
+  foo.bar()
+} catch (e) {
+  bugsnagClient.notify(e, {
+    beforeSend: function () {
+      setTimeout(function () {
+        var el = document.getElementById('bugsnag-test-state')
+        el.textContent = el.innerText = 'DONE'
+      }, 5000)
```
fractalwrench: This `beforeSend` function appears to be the same in a few files. Is there any way of extracting this code to one location?

features/steps/browser_steps.rb
```diff
@@ -0,0 +1,58 @@
...
+
+When("the exception matches the {string} values for the current browser") do |fixture|
+  err = get_error_message(fixture)
+  steps %Q{
+    And the exception "errorClass" equals "#{err['errorClass']}"
+    And the exception "message" equals "#{err['errorMessage']}"
+  }
+end
+
+Then(/^the request is a valid browser payload for the error reporting API$/) do
```
fractalwrench: The checks here are good, was there a reason why this needs to be separate from the standard error reporting API step in mazerunner? There may be other notifiers which don't send an API key in the headers.

features/support/env.rb
```diff
@@ -0,0 +1,91 @@
...
+      end
+    end
+  end
+end
+
+# test helpers
+
+def current_ip
+  # Parses the output of `ifconfig` to retreive the host IP for docker to talk to
+  # Breaks compatability with Windows
```
fractalwrench: Is there any simple way we can maintain compatibility with Windows here?

features/unhandled_errors.feature
```diff
@@ -0,0 +1,70 @@
...
+
+Scenario Outline: syntax errors
+  When I navigate to the URL "/unhandled/<type>/a.html"
+  And the test should run in this browser
+  And I let the test page run for up to 10 seconds
+  And I wait for 5 seconds
+  Then I should receive 1 request
+  And the request is a valid browser payload for the error reporting API
+  And the exception matches the "unhandled_syntax" values for the current browser
+  # TODO: add stacktrace assertions to this type of error
```
fractalwrench: Does this still need to be addressed?

features/release_stage.feature
```diff
@@ -0,0 +1,48 @@
+@release_stage
+Feature: Configuring releaseStage and notifyReleaseStages
```
fractalwrench: It would be good to add scenarios to cover nil/empty releaseStages as well

features/fixtures/ip_redaction/script/a.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="/node_modules/bugsnag-js/dist/bugsnag.min.js"></script>
+    <script type="text/javascript">
+      var ENDPOINT = decodeURIComponent(window.location.search.match(/ENDPOINT=(.+)/)[1])
```
fractalwrench: I think I've seen `window.location.search.match(/ENDPOINT=(.+)/)[1]` (and a couple of similar snippets) in multiple places within the fixture scripts. Is it sensible to extract these sort of snippets into separate scripts, or is that a bad idea?

fractalwrench:
Added a few more inline comments. It'd be good to get someone else to also have a look through this, given the scope of the changes.

.npmignore
bengourley: I believe .npmignore inherited rules from .gitignore (so your only need to specify the differences) but I will confirm.

.travis.yml
bengourley: No, it needs to be readded as per the "changed" section in the PR body (those two sets of tests are what generate the coverage).

features/lib/server.rb
```diff
@@ -0,0 +1,5 @@
+require 'webrick'
+
+root = File.expand_path File.join(__dir__, '..', 'fixtures')
+server = WEBrick::HTTPServer.new :Port => ENV['PORT'], :DocumentRoot => root
+server.start
```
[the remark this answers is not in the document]
bengourley: All of the static assets in `features/fixtures`. This line is resolving the path to that directory in a cross platform way:

```

features/before_send.feature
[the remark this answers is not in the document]
kattrali: "should run in this browser" seems like a nice general check, though my one concern with it would be accidentally excluding tests. Though given the implementation, that seems unlikely.

features/fixtures/handled/browserify/src/a.js
kattrali: Having it duplicated means that each e2e test fixture is self-contained, making it easier to follow and and reduces the chance that a slight change in a helper library could break/invalidate the test result.

features/fixtures/handled/browserify/src/b.js
kattrali: Having it duplicated means that each e2e test fixture is self-contained, making it easier to follow and and reduces the chance that a slight change in a helper library could break/invalidate the test result.

PR #355, opened 2018-06-04
Fix a bug with SyntaxError stacktraces in Chrome

bengourley:
## Goal

`SyntaxError`s in Chrome have an `error.stack` property which is just `${err.name}: ${err.message}` and no stack frames, but because the `err.stack` value is truthy we attempted to parse it with `error-stack-parser`.

The result of this was that the filename property of the stackframes was not correct and caused issues with grouping. This change causes Report.getStacktrace(err) to generate a backtrace instead and then the file/line/col properties get filled in with the values from `window.onerror`.

### Added

A failing end-to-end test.

### Changed

The `hasStack(obj)` function which is used to determine whether we should parse the stacktrace now returns `false` if `obj.stack === ${err.name}: ${err.message}`. This causes the notifier to extract more information from the arguments which are passed to `window.onerror`, including filename, which means the stacktrace is correct in Chrome, matching that of other browsers.

## Tests

More detailed assertions about stackframes were added to the existing unhandled SyntaxError test case.

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
  - [x] Initial review of the intended approach, not yet feature complete
  - [x] Structural review of the classes, functions, and properties modified
  - [x] Final review


features/steps/browser_steps.rb
```diff
@@ -31,6 +31,9 @@
   if err['columnNumber']
     step("the \"columnNumber\" of stack frame 0 equals #{err['columnNumber']}")
   end
+  if err['file']
+    step("the \"file\" of stack frame 0 ends with \"#{err['file']}\"")
```
bengourley: this is an "ends with" match so that the hard coded values don't need to contain the hostname/port/protocol

PR #358, opened 2018-06-07
Fix CSP bug in iOS9

### LINKED ISSUE #357: Bugsnag breaks app with CSP in place
### Expected behavior
I expect Bugsnag to load just fine with a CSP that follows the [documentation](https://docs.bugsnag.com/platforms/browsers/faq/#can-i-use-bugsnag-with-csp).

### Observed behavior
I get errors that `sessions.bugsnag.com` is not part of my CSP and the connection is refused. Even worse, on iOS 9/10, this completely breaks the loading of my app!

### Steps to reproduce
Setup Bugsnag with a CSP as directed in the FAQ. Then, try and load it.

### Version
v4 from the CDN link. Not sure which specific version it is.

### Additional information
This seemed to break last week sometime.

This is a *huge* issue, and I'm disappointed that more notice wasn't given for a change like this.

bengourley:
## Goal

In iOS9 Safari, a request that violates the Content Security Policy (CSP) rules causes a `DOMException` to be thrown. Since v4.7.0, bugsnag calls `startSession()` synchronously upon startup, which means that any thrown error in that tick of the event loop results in the libary (and anything else in that synchronous execution context) to be unusable. The behaviour of iOS Safari was fixed in version 10, but since we support 9, this needs a workaround.

### Changed

The `xml-http-request` transport mechanism already wrapped `request.send()` with a try/catch, but the error in this instance was coming from the `request.open()` call. The try/catch block was moved out out to include all of the synchronous code inside `sendReport()`/`sendSession()`.

## Tests

A test that would fail for this browser was added to the end-to-end tests. However, BrowserStack does not support any iOS9 devices via their webdriver protocol, which is what we use. The oldest version available is iOS10.3, which does not exhibit this behaviour.

It is possible to verify locally with an iOS 9 simulator:

- `npm run build`
- run the tests suite in any browser to ensure the dependencies are loaded `BROWSER=chrome_61 bundle exec maze-runner features/csp.feature`
- serve the test fixtures `PORT=9020 ruby features/lib/server.rb`
- navigate to `http://localhost:9020/csp/script/a.html?ENDPOINT=/abc`
- after 5 seconds the text `PENDING` should change. If it changes to `DONE` the test passes, `ERROR` then the test fails

Here is a screenshot of it having run in my simulator:
image

### Alternative Approaches

We could consider switch to SauceLabs, where you can use older iOS devices. This would require a lot more work.

### Linked issues

Fixes #357.

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
  - [x] Initial review of the intended approach, not yet feature complete
  - [x] Structural review of the classes, functions, and properties modified
  - [x] Final review


features/steps/browser_steps.rb
```diff
@@ -15,8 +17,13 @@
...
   wait.until {
     $driver.find_element(id: 'bugsnag-test-state') &&
-    $driver.find_element(id: 'bugsnag-test-state').text == 'DONE'
+    (
+      $driver.find_element(id: 'bugsnag-test-state').text == 'DONE' ||
+      $driver.find_element(id: 'bugsnag-test-state').text == 'ERROR'
+    )
   }
+  txt = $driver.find_element(id: 'bugsnag-test-state').text
+  assert_equal('DONE', txt, "Expected #bugsnag-test-state text to be 'DONE'. It was '#{txt}'.")
```
bengourley: This adds a way for the test fixture app to communicate that it is complete, but that an error happened.

bengourley:
Thanks! Yeah until we arrange some time to rejig the automated release process it still prompts for the changelog entry at the time of the release.

PR #360, opened 2018-06-21
Dont send stacktrace as breadcrumb metadata

snmaynard:
This is dropped by the UI and never shown as we can't render arrays in the breadcrumbs from what I can see.


bengourley:
it looks like browserstack stopped supporting simulators/emulators on their js testing product – so the karma test suite is failing now. I'll update the browser matrix and merge that into this branch to get the tests passing.

PR #368, opened 2018-07-16
feat: Implement asynchronous beforeSend callbacks

bengourley:
This was previously implemented in #314 but was not merged because of the size increase. Now this feature is essential to support code loading for stackframes in Node. This implementation supports Node-style error first callbacks and Promises, whilst retaining support for synchronous callbacks.

### DIFF
--- a/packages/core/lib/test/es-utils.test.js
+++ b/packages/core/lib/test/es-utils.test.js
@@ -2,7 +2,7 @@ const { describe, it, expect } = global
 
 const { map, reduce, filter, keys, isArray, includes } = require('../es-utils')
 
-describe('us-utils', () => {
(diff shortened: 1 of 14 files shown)


packages/core/lib/test/es-utils.test.js
```diff
@@ -2,7 +2,7 @@ const { describe, it, expect } = global
 
 const { map, reduce, filter, keys, isArray, includes } = require('../es-utils')
 
-describe('us-utils', () => {
```
kattrali:
```diff
- 🇺🇸 
+ 🇪🇸 
```

bengourley: Olé 💃

PR #369, opened 2018-07-17
Import server session implementation from bugsnag-node

bengourley:
Imports the existing implementation from bugsnag-node, and makes the necessary changes/refactors to work with the client plugin interface.

As part of this the `beforeSession` callback implementation is removed. There were a few problems with it which meant either a lot more code to fix, or its removal, which turned out to be simpler.

- `beforeSession` functions are passed the `Client` and not the `Session` object
- the `Session` data structure does not contain device/app data, but that is what the `beforeSession` callbacks were used for
- it did not work well with the node/server session tracking, which creates "child" clients where the summaries are sent from the session "tracker" – the data is needed at a per `Client` level not a per `Session` level, so it is simpler to just add the required data at the client level.

### DIFF
--- a/packages/browser/types/test/types.test.js
+++ b/packages/browser/types/test/types.test.js
@@ -15,6 +15,7 @@ const exampleValue = (k) => {
   switch (k) {
     case 'apiKey': return 'abc'
     case 'appVersion': return '1.2.3'
+    case 'appType': return 'worker'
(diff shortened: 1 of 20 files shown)


bengourley:
I believe `package-lock.json` should be checked in – it is for all other packages so far.

PR #370, opened 2018-07-18
Add node delivery mechanism

bengourley:
Adds @bugsnag/delivery-node package which sends error reports/sessions using request.

Since request is quite big (5MB on disk!) we discussed alternatives, but went with it for now as it's the most widely used and reliable for proxy support (which we need to support out of the box).

### DIFF
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -11,19 +11,14 @@ module.exports = {
   },
   logger: {
     ...schema.logger,
-    defaultValue: () =>
-      // set logger based on browser capability
-      (typeof console !== 'undefined' && typeof console.debug === 'function')
-        ? getPrefixedConsole()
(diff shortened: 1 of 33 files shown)


packages/node/src/config.js
```diff
@@ -11,19 +11,14 @@ module.exports = {
   },
   logger: {
     ...schema.logger,
-    defaultValue: () =>
-      // set logger based on browser capability
-      (typeof console !== 'undefined' && typeof console.debug === 'function')
-        ? getPrefixedConsole()
-        : undefined
+    defaultValue: () => getPrefixedConsole()
```
fractalwrench: Can the console no longer be undefined?

bengourley: No, in node it's always defined.

PR #371, opened 2018-07-25
Add surrounding code plugin for node notifier

bengourley:
Using the async support in `beforeSend` callbacks, this adds a callback which attempts to load the `file` at each stackframe and adds up to 3 lines code either side, if available, to the frame.

### DIFF
--- a/packages/core/lib/run-before-send.js
+++ b/packages/core/lib/run-before-send.js
@@ -8,11 +8,13 @@ module.exports = (report, onError) => (fn, cb) => {
       if (ret && typeof ret.then === 'function') {
         return ret.then(
           // resolve
-          val => cb(null, shouldPreventSend(report, val)),
+          val => setTimeout(() => cb(null, shouldPreventSend(report, val)), 0),
           // reject
           err => {
-            onError(err)
-            return cb(null, false)
+            setTimeout(() => {
+              onError(err)
+              return cb(null, false)
+            })
           }
         )
       }
--- a/packages/plugin-node-surrounding-code/surrounding-code.js
+++ b/packages/plugin-node-surrounding-code/surrounding-code.js
@@ -1,3 +1,4 @@
+const SURROUNDING_LINES = 3
 const { readFile } = require('fs')
 
 module.exports = {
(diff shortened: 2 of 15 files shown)


packages/core/lib/run-before-send.js
```diff
@@ -8,11 +8,13 @@ module.exports = (report, onError) => (fn, cb) => {
       if (ret && typeof ret.then === 'function') {
         return ret.then(
           // resolve
-          val => cb(null, shouldPreventSend(report, val)),
+          val => setTimeout(() => cb(null, shouldPreventSend(report, val)), 0),
```
bengourley: This ensures that any errors that are thrown in the callback `cb()` do not get swallowed by this Promise, and spat out as an unhandled rejection event.

By ensuring it executes outside of the synchronous stack of this promise call, it will throw an uncaught error.

The way this was manifesting itself was that an error in the tests was causing the tests to hang, rather than to fail the test with an error.

packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -0,0 +1,43 @@
...
+          report.stacktrace = stacktrace
+          resolve()
+        })
+        .catch(reject)
+    }))
+  }
+}
+
+const getSurroundingCode = (file, lineNumber, cb) => {
+  readFile(file, 'utf8', (err, data) => {
```
fractalwrench: Will JS files always be encoded in UTF-8?

packages/plugin-node-surrounding-code/test/surrounding-code.test.js
```diff
@@ -0,0 +1,129 @@
+const { describe, it, expect } = global
+
+const plugin = require('../')
+const { join } = require('path')
+const Report = require('@bugsnag/core/report')
+const Client = require('@bugsnag/core/client')
+const VALID_NOTIFIER = { name: 't', version: '0', url: 'http://' }
```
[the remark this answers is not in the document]
bengourley: Good point. For a more memory-safe implementation we could use streams rather than buffering the entire source file in memory. It is possible for a very large file to result in an OOM crash.

packages/plugin-node-surrounding-code/surrounding-code.js
bengourley: In node, I believe this to be the case. This SO seems to back that belief up, but not conclusively: https://stackoverflow.com/questions/10125141/what-encoding-is-expected-for-node-js-source-code

In any case, the most likely other encoding is `ascii` which is a subset of `utf8`.

The other options that node support for reading files are `binary`, `hex`, `base64` and `utf16le` – but the first three would be valid source code so wouldn't run. The latter may result in working source code and could encounter encoding issues if we interpret as `utf8`. I don't know if anyone has ever encoded a node source file in `utf16le` 🤔

fractalwrench: As I imagine this is the most common case and the previous code assumed this, I don't think any further action is required on this

PR #373, opened 2018-07-26
perf: Use streams for surrounding code loading

bengourley:
Based on @fractalwrench's comment on #371, I've thrown together a streaming implementation for the surrounding code loading. I think it's very unlikely we'd run out of memory, but this would help in the fraction of cases where we would cross that threshold. As you can see it's more complex but it passes all the same tests.

### DIFF
--- a/packages/plugin-node-surrounding-code/surrounding-code.js
+++ b/packages/plugin-node-surrounding-code/surrounding-code.js
@@ -1,5 +1,8 @@
 const SURROUNDING_LINES = 3
-const { readFile } = require('fs')
+const { createReadStream } = require('fs')
+const { Writable } = require('stream')
+const pump = require('pump')
+const byline = require('byline')
 
 module.exports = {
   init: client => {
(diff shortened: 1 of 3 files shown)


packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -28,16 +31,56 @@ module.exports = {
...
+//   '10': 'function getSquare (cb) {',
+//   '11': '  rectangles.find({',
+//   '12': '    length: 12',
+//   '13': '    width: 12',
+//   '14': '  }, err => cb)',
+//   '15': '}'
+// }
+class CodeRange extends Writable {
+  constructor (opts) {
+    super({ ...opts, decodeStrings: false })
```
bengourley: by default, writable streams convert strings into raw buffers but we don't want that here

packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -28,16 +31,56 @@ module.exports = {
...
-      }, {}))
+  const start = lineNumber - SURROUNDING_LINES
+  const end = lineNumber + SURROUNDING_LINES
+
+  const reader = createReadStream(file, { encoding: 'utf8' })
+  const splitter = new byline.LineStream({ keepEmptyLines: true })
+  const slicer = new CodeRange({ start, end })
+
+  // if the slicer has enough lines already, no need to keep reading from the file
+  slicer.on('done', () => reader.destroy())
```
bengourley: there's no way to cleanly kill a readable stream, you have to call `destroy()` on it which triggers an error. We tolerate this "premature close" error below. If we didn't do this, the entire file would be read chunk by chunk and just not operated on, which is a noop, but it does use resources for no reason.

packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -28,16 +31,56 @@ module.exports = {
...
-      .slice(start, end)
-      .reduce((accum, line, i) => {
-        accum[start + (i + 1)] = line
-        return accum
-      }, {}))
+  const start = lineNumber - SURROUNDING_LINES
+  const end = lineNumber + SURROUNDING_LINES
+
+  const reader = createReadStream(file, { encoding: 'utf8' })
+  const splitter = new byline.LineStream({ keepEmptyLines: true })
```
bengourley: byline discards empty lines by default. This took a while to track down!

packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -1,5 +1,8 @@
 const SURROUNDING_LINES = 3
-const { readFile } = require('fs')
+const { createReadStream } = require('fs')
+const { Writable } = require('stream')
+const pump = require('pump')
```
bengourley: this modules is used to pipe streams together. It removes some of the boilerplate around handling errors with more than one stream.

packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -1,5 +1,8 @@
...
+const byline = require('byline')
```
bengourley: this module is used to buffer the file read stream up to each line break and emit chunks which are single line of code

PR #374, opened 2018-07-26
feat: Add "in project" and project root stripping to stack traces

bengourley:
This adds two small plugins for the node notifier to:

- mark stackframes as `inProject: true|false`
- strip the `config.projectRoot` from `stackframe.file`


packages/plugin-node-in-project/in-project.js
```diff
@@ -0,0 +1,16 @@
+const { map } = require('@bugsnag/core/lib/es-utils')
+const normalizePath = require('@bugsnag/core/lib/path-normalizer')
+
+module.exports = {
+  init: client => client.config.beforeSend.push(report => {
+    report.stacktrace = map(report.stacktrace, stackframe => {
+      const projectRoot = normalizePath(client.config.projectRoot)
```
kattrali: Shouldn't this be outside of the loop since it won't change per frame?

bengourley: Yes, good spot, thanks!

PR #378, opened 2018-07-27
fix: Various improvements to type definitions

bengourley:
This adds the patch from #377 to the `universal` branch, along with some other type-based improvements

- Add tslint to automated tests
- Fix ambiguity around beforeSend callback signature*
- Linting fixes

 [*] without this update, the `beforeSend` callback in the "should work with breadcrumbs" test would fail with this error:

```
error TS7006: Parameter 'report' implicitly has an 'any' type.
```


packages/core/types/common.d.ts
```diff
@@ -46,11 +46,7 @@ export interface IFinalConfig extends IConfig {
   logger?: ILogger | null;
 }
 
-type SyncBeforeSend = (report: Report) => void;
-type AsyncBeforeSend = (report: Report, cb: (err: null | Error) => void) => void;
-type PromiseBeforeSend = (report: Report) => Promise<void>;
-
-export type BeforeSend = SyncBeforeSend | AsyncBeforeSend | PromiseBeforeSend;
+export type BeforeSend = (report: Report, cb?: (err: null | Error) => void) => void | Promise<void>;
```
bengourley: I'm not a big fan of this change, I liked the way it was before. However, it prevented the compiler from being able to implying the type of this function. Now this code would be allowed where it wouldn't before (allowing the mixing of callbacks and promises):

```typescript
beforeSend: (report, callback) => return new Promise(...)
```

packages/browser/types/test/fixtures/app.js
```diff
@@ -1,10 +0,0 @@
-"use strict";
-exports.__esModule = true;
-var bugsnagInstance = undefined;
-function notify(error, opts) {
-    if (bugsnagInstance === undefined) {
-        return false;
-    }
-    return bugsnagInstance.notify(error, opts);
-}
-exports.notify = notify;
```
fractalwrench: Are these two files no longer used?

bengourley: They get generated by the tests and were erroneously checked in.

PR #380, opened 2018-08-02
[wip] maze runner tests for @bugsnag/node

bengourley:


packages/node/features/support/env.rb
```diff
@@ -0,0 +1,31 @@
...
+
+After do
+# Runs after every Scenario
+end
+
+at_exit do
+# Runs when the test run is completed
+end
+
+def current_ip
```
bengourley: I copied this function from bugsnag-ruby's suite – it appears to work.

packages/node/src/config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
 
 const getPrefixedConsole = () => {
   return reduce([ 'debug', 'info', 'warn', 'error' ], (accum, method) => {
-    const consoleMethod = console[method]
+    const consoleMethod = console[method] || console.log
```
bengourley: node 4 doesn't have `console.debug()`

PR #381, opened 2018-08-07
Add maze runner tests for server session tracking

bengourley:
Adds a session tracking feature file with tests for session summary payloads.

As part of the validation for a session tracking payload, the `device` property was required, so I added the `@bugsnag/plugin-node-device` package to ensure that was populated.

This also refactors the `positiveIntIfDefined` validator, because as some point the implementation of that function and its name diverged. It is now an int range validator, which means it is more generalisable (see 7ff0f91 for more detail).

### DIFF
--- a/packages/core/lib/test/validators.test.js
+++ b/packages/core/lib/test/validators.test.js
@@ -1,17 +1,24 @@
 const { describe, it, expect } = global
 
-const { positiveIntIfDefined, stringWithLength } = require('../validators')
+const { intRange, stringWithLength } = require('../validators')
 
-describe('positiveIntIfDefined(val)', () => {
+describe('intRange(min, max)(val)', () => {
   it('work with various values', () => {
-    expect(positiveIntIfDefined(10)).toBe(true)
-    expect(positiveIntIfDefined(1e5)).toBe(true)
-    for (let i = 1; i <= 1000; i++) expect(positiveIntIfDefined(i)).toBe(true)
-    expect(positiveIntIfDefined(-10)).toBe(false)
-    expect(positiveIntIfDefined(0)).toBe(false)
-    expect(positiveIntIfDefined(1.123)).toBe(false)
-    expect(positiveIntIfDefined('')).toBe(false)
-    expect(positiveIntIfDefined('100')).toBe(false)
+    // default min/max: 1 to Infinity
+    expect(intRange()(10)).toBe(true)
+    expect(intRange()(1e5)).toBe(true)
+    for (let i = 1; i <= 1000; i++) expect(intRange()(i)).toBe(true)
+    expect(intRange()(-10)).toBe(false)
+    expect(intRange()(0)).toBe(false)
+    expect(intRange()(1.123)).toBe(false)
+    expect(intRange()('')).toBe(false)
+    expect(intRange()('100')).toBe(false)
+
+    // custom min/max
+    expect(intRange(-10, 20)(11)).toBe(true)
+    expect(intRange(-10, 20)(-13)).toBe(false)
+    expect(intRange(-10, 20)(20)).toBe(true)
+    expect(intRange(-10, 20)(21)).toBe(false)
   })
 })
 
--- a/packages/core/lib/validators.js
+++ b/packages/core/lib/validators.js
@@ -1,8 +1,6 @@
-const { includes } = require('./es-utils')
-
-exports.positiveIntIfDefined = value =>
-  includes([ 'undefined', 'number' ], typeof value) &&
+exports.intRange = (min = 1, max = Infinity) => value =>
+  typeof value === 'number' &&
   parseInt('' + value, 10) === value &&
-  value > 0
+  value >= min && value <= max
 
 exports.stringWithLength = value => typeof value === 'string' && !!value.length
--- a/packages/node/features/fixtures/sessions/scenarios/start-session-notify.js
+++ b/packages/node/features/fixtures/sessions/scenarios/start-session-notify.js
@@ -1,0 +1,17 @@
+var bugsnag = require('@bugsnag/node')
+var bugsnagClient = bugsnag({
+  apiKey: process.env.BUGSNAG_API_KEY,
+  endpoints: {
+    notify: process.env.BUGSNAG_NOTIFY_ENDPOINT,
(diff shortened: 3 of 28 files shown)


packages/plugin-simple-throttle/throttle.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
     maxEvents: {
       defaultValue: () => 10,
       message: 'should be a positive integer ≤100',
-      validate: val => positiveIntIfDefined(val) && val < 100
+      validate: val => intRange(1, 100)(val)
```
fractalwrench: I believe that <100 has changed to <= 100 here, is that intentional/does it matter?

packages/plugin-server-session/tracker.js
```diff
@@ -30,13 +30,10 @@ module.exports = class SessionTracker extends Emitter {
   }
 
   _summarize () {
-    const thisMin = dateToMsKey(new Date())
     const summary = []
     this._sessions.forEach((val, key) => {
-      if (key !== thisMin) {
```
fractalwrench: Is this check not needed anymore?

packages/plugin-node-device/test/device.test.js
```diff
@@ -0,0 +1,35 @@
...
+const ISO_8601 = /^\d{4}(-\d\d(-\d\d(T\d\d:\d\d(:\d\d)?(\.\d+)?(([+-]\d\d:\d\d)|Z)?)?)?)?$/i
+
+describe('plugin: node device', () => {
+  it('should set device = { hostname} and add a beforeSend callback which adds device time', done => {
+    const client = new Client(VALID_NOTIFIER, schema)
+    client.configure({ apiKey: 'API_KEY_YEAH' })
+    plugin.init(client)
+
+    expect(client.config.beforeSend.length).toBe(1)
+    expect(client.device.hostname).toBeDefined()
```
fractalwrench: is there an easy way of verifying the correct hostname is populated?

packages/node/features/sessions.feature
```diff
@@ -0,0 +1,64 @@
+Feature: Server side session tracking
```
fractalwrench: It may be worth considering a scenario that tests batch delivery. For example, the Java notifier deliver sessions in a batch every X seconds, and I believe that we have a test that ensures that if 100 sessions are created over that boundary, M sessions are sent in the first request, and M sessions are sent in the second. I don't know if that's already covered or if a similar mechanism is used here.

packages/node/features/fixtures/sessions/scenarios/start-session-notify.js
```diff
@@ -0,0 +1,17 @@
...
+  sessionSummaryInterval: 1000
+})
+
+var sessionClient = bugsnagClient.startSession()
+setTimeout(function () {
+  sessionClient.notify(new Error('in a session'))
+}, 1500)
+
+// keep the process open a little bit so that the session has time to get sent
+setTimeout(function () {}, 5000)
```
fractalwrench: Is there any way of extracting a constant out of 1000/5000 across these fixtures? We may want to tweak this value in future

packages/core/lib/validators.js
```diff
@@ -1,8 +1,6 @@
-const { includes } = require('./es-utils')
-
-exports.positiveIntIfDefined = value =>
-  includes([ 'undefined', 'number' ], typeof value) &&
+exports.intRange = (min = 1, max = Infinity) => value =>
```
fractalwrench: 👌

packages/core/lib/test/validators.test.js
```diff
@@ -1,17 +1,24 @@
...
+    expect(intRange()(0)).toBe(false)
+    expect(intRange()(1.123)).toBe(false)
+    expect(intRange()('')).toBe(false)
+    expect(intRange()('100')).toBe(false)
+
+    // custom min/max
+    expect(intRange(-10, 20)(11)).toBe(true)
+    expect(intRange(-10, 20)(-13)).toBe(false)
+    expect(intRange(-10, 20)(20)).toBe(true)
+    expect(intRange(-10, 20)(21)).toBe(false)
```
fractalwrench: May also be worth adding a few scenarios to ensure invalid parameters to intRange are detected, for example:

```
intRange(10, 5) // typo, meant to write 50
```

packages/plugin-simple-throttle/throttle.js
bengourley: it now matches the explanation in the message above!

packages/plugin-server-session/tracker.js
bengourley: ah I meant to explain this…

So the tests were failing to send any session summaries even though I brought the interval right down to a second. This turned out to be the thing that was preventing it – it refused to the summary for the current minute, on the premise that more sessions could be added before the minute is over. I checked with @bullmo and it turns out that it doesn't matter if we send multiple summaries for the same minute – they will be summed together in the sessions service. So tl;dr, no this check is not needed, but thanks for making sure I explained it 🦅 👀

packages/plugin-node-device/test/device.test.js
bengourley: here it is just a pretend value that is passed in by the unit test, so yeah I could check that it is equal to that rather than simply defined

packages/node/features/fixtures/sessions/scenarios/start-session-notify.js
bengourley: I think it depends on the test and I'm not keen on factoring this out unless it is duplicated at least 3x

packages/node/features/sessions.feature
bengourley: Good point, I think this would be a decent addition. Let me see if I can knock something together quickly when I address the other feedback, and if not I'll make a ticket to add this at a later point.

packages/core/lib/test/validators.test.js
bengourley: I did consider this, but if we test for that, we would then want to make the function do something sensible like `throw new Error('min must be less than max!')` – which would add more code for the browser bundle. Since all of the internals/plugins that use this helper have their own unit tests, it will be quite clear from those if you specify an unsatisfiable range so I think this would be unnecessary.

PR #382, opened 2018-08-09
refactor: Change configuration logic to better support external plugins

bengourley:
This is a pretty hefty refactor so I'll do some explaining!

### 1. Schema composition

For plugins to augment configuration the existing implementation required a two-stage process. The second argument to `new Client(notifier, configSchema)` was a complete schema definition for all configuration options. Out of necessity this schema was composed before the client was created:

https://github.com/bugsnag/bugsnag-js/blob/bd50c96cb25291c23778cc03fca3030b3790ee40/packages/browser/src/notifier.js#L72-L78

For user-defined, or external plugins, the burden was placed on the user to pass in the plugins at the moment when the `bugsnag()` initialiser was called:

https://github.com/bugsnag/bugsnag-js/blob/bd50c96cb25291c23778cc03fca3030b3790ee40/packages/browser/src/notifier.js#L50

### 2. API consistency

Users only needed pass in the plugins at initialise-time for plugins that augmented the schema. For any others, they could just call `bugsnagClient.use(plugin)` at a later point. This is inconsistent. It also has a blind spot where if a plugin needed to return something useful, but also augment the schema, this could not be done:

```js
const myPlugin = {
  init: () => {
    return () => console.log('useful!')
  },
  configSchema: { myPluginOption: ... }
}

bugsnag('api_key', [ myPlugin ])
// adds the configuration option but no way to
// access the function returned from init()

const usefulFn = bugsnagClient.use(myPlugin)
// now we have a reference to the function but
// the schema option was not added
```

### Changeset summary

Initialiser signature:

```diff
- bugsnag(opts, plugins)
+ bugsnag(opts)
```

Plugin interface:

```diff
{
+ name: 'myPlugin'
- init: (client) { }
+ init: (client, ...args) { }
  configSchema: {}
}
```

`Client` class:

```diff
+ client.setOptions (opts)
- client.configure(opts)
+ client.configure(partialSchema)
- client.use(plugin)
+ client.use(plugin, ...args)
+ client.getPlugin(name)
```

#### Public API

The _one true way_ to add a plugin now is to call:

```js
bugsnagClient.use(plugin)
```

To access something useful returned by a plugin call:

```js
const anything = bugnsagClient.getPlugin('myPlugin')
```

This takes the burden off the user to store references to something returned by a plugin throughout their application – it means that wherever they have access to `bugsnagClient` they can get at the provided utilities (such as a React `ErrorBoundary`, or an `intercept(fn)` function).

#### Internals

The `bugsnag()` initialiser for each notifier (@bugnsag/node, @bugsnag/browser) has the job of coordinating the following setup procedure on the client instance and then returning it:

```js
// create a client instance with the notifer name, version and url
const client = new Client(notifier)

// configure the delivery mechanism
client.delivery(d)

// pass in the raw user-supplied options
client.setOptions(opts)

// call configure with the schema for the core config, plus any notifier specific config.
// the job of configure() is to extract, provide default values for and validate the raw user
// supplied options passed to setOptions() and aggregate the result at client.config
client.configure(schema)

// add any built in plugins
// each call to use() may result in a call to configure() if the plugin provides
// a config schema. This results in an incremental build up the client's configuration
client.use(plugin)
```

#### Other changes

To pass arbitrary references to a plugin, the `client.use()` method has the signature, `(plugin, ...)` and it will pass anything supplied after `plugin` as arguments to `plugin.init(client, ...)`.

The organisation of the types was rejigged, so that the browser notifier-specific options (that aren't defined in inidividual plugins) extend the core `IPlugin` interface.

#### Summary

The resulting interface is such that the notifier can be incrementally configured. The entire options object can be passed in up front (if desired) but the configuration schema can be supplied by each plugin on the fly.

## Huh?

- I'm considering making the `name` property optional and allowing anonymous plugins. This is purely   a space-saving exercise for the browser bundle, but the plugins only need a name if they return something useful that the user might want later. Thoughts?
- I feel like I reached an ideal implementation with with `setOptions(opts)` -> `configure(schema)` -> `use(plugin)` flow, but it still seems a little clunky. It's abstracted from the user anyway but I wonder if it might be possible to improve it further?
- Now that the `IConfig` interface for types includes a wildcard property `[key: string]: any;` it's harder to test. At least its more permissive if we forget to define a new option, but I wonder if there's a way to improve this, or to catch a missing definition with the tests?

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -78,7 +78,6 @@ class BugsnagClient {
 
     this._configured = true
 
-    this._logger.debug(`Loaded!`)
     return this
   }
 
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -84,10 +84,12 @@ class BugsnagClient {
   use (plugin, ...args) {
     if (!this._configured) throw new Error('client not configured')
     if (plugin.configSchema) this.configure(plugin.configSchema)
+    const result = plugin.init(this, ...args)
     // JS objects are not the safest way to store arbitrarily keyed values,
     // so bookend the key with some characters that prevent tampering with
-    // stuff like __proto__ etc.
-    this._plugins[`~${plugin.name}~`] = plugin.init(this, ...args)
+    // stuff like __proto__ etc. (only store the result if the plugin had a
+    // name)
+    if (plugin.name) this._plugins[`~${plugin.name}~`] = result
     return this
   }
 
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -14,7 +14,7 @@ const REPORT_USAGE_ERR_PREFIX = `Bugsnag usage error.`
 const noop = () => {}
 
 class BugsnagClient {
-  constructor (notifier, configSchema = config.schema, session = null) {
+  constructor (notifier) {
     if (!notifier || !notifier.name || !notifier.version || !notifier.url) {
       throw new Error('`notifier` argument is required')
     }
@@ -22,20 +22,21 @@ class BugsnagClient {
     // notifier id
     this.notifier = notifier
 
-    // config
-    this.configSchema = configSchema
-
     // configure() should be called before notify()
     this._configured = false
 
-    // i/o
+    // intialise opts and config
+    this._opts = {}
+    this.config = {}
+
+    // // i/o
     this._delivery = { sendSession: noop, sendReport: noop }
     this._logger = { debug: noop, info: noop, warn: noop, error: noop }
 
     // plugins
-    this.plugins = []
+    this._plugins = {}
 
-    this.session = session
+    this._session = null
 
     this.breadcrumbs = []
 
@@ -54,24 +55,45 @@ class BugsnagClient {
     this.BugsnagSession = BugsnagSession
   }
 
-  configure (opts = {}) {
-    this.config = config.mergeDefaults({ ...this.config, ...opts }, this.configSchema)
-    const validity = config.validate(this.config, this.configSchema)
+  setOptions (opts) {
+    this._opts = { ...this._opts, ...opts }
+  }
+
+  configure (partialSchema = config.schema) {
+    const conf = config.mergeDefaults(this._opts, partialSchema)
+    const validity = config.validate(conf, partialSchema)
+
     if (!validity.valid === true) throw new Error(generateConfigErrorMessage(validity.errors))
-    if (typeof this.config.beforeSend === 'function') this.config.beforeSend = [ this.config.beforeSend ]
-    if (this.config.appVersion !== null) this.app.version = this.config.appVersion
-    if (this.config.appType !== null) this.app.type = this.config.appType
-    if (this.config.metaData) this.metaData = this.config.metaData
-    if (this.config.user) this.user = this.config.user
-    if (this.config.logger) this.logger(this.config.logger)
+
+    // update and elevate some special options if they were passed in at this point
+    if (typeof conf.beforeSend === 'function') conf.beforeSend = [ conf.beforeSend ]
+    if (conf.appVersion) this.app.version = conf.appVersion
+    if (conf.appType) this.app.type = conf.appType
+    if (conf.metaData) this.metaData = conf.metaData
+    if (conf.user) this.user = conf.user
+    if (conf.logger) this.logger(conf.logger)
+
+    // merge with existing config
+    this.config = { ...this.config, ...conf }
+
     this._configured = true
+
     this._logger.debug(`Loaded!`)
     return this
   }
 
-  use (plugin) {
-    this.plugins.push(plugin)
-    return plugin.init(this)
+  use (plugin, ...args) {
+    if (!this._configured) throw new Error('client not configured')
+    if (plugin.configSchema) this.configure(plugin.configSchema)
+    // JS objects are not the safest way to store arbitrarily keyed values,
+    // so bookend the key with some characters that prevent tampering with
+    // stuff like __proto__ etc.
+    this._plugins[`~${plugin.name}~`] = plugin.init(this, ...args)
+    return this
+  }
+
+  getPlugin (name) {
+    return this._plugins[`~${name}~`]
   }
 
   delivery (d) {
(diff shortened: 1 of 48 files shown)


packages/plugin-window-onerror/test/onerror.test.js
```diff
@@ -42,8 +44,9 @@ describe('plugin: window onerror', () => {
     // it('captures uncaught errors in DOM (level 2) event handlers', done => {
```
fractalwrench: We could delete this unless there's a reason it needs to stay commented out

packages/plugin-window-unhandled-rejection/test/unhandled-rejection.test.js
```diff
@@ -61,8 +63,9 @@ describe('plugin: unhandled rejection', () => {
     //
```
fractalwrench: Could also be deleted

packages/core/client.js
```diff
@@ -14,28 +14,29 @@ const REPORT_USAGE_ERR_PREFIX = `Bugsnag usage error.`
...
+    // // i/o
     this._delivery = { sendSession: noop, sendReport: noop }
     this._logger = { debug: noop, info: noop, warn: noop, error: noop }
 
     // plugins
-    this.plugins = []
+    this._plugins = {}
 
-    this.session = session
+    this._session = null
```
fractalwrench: do we need to worry about backward compatibility for this one at all?

packages/core/types/common.d.ts
```diff
@@ -8,49 +8,23 @@ export interface IConfig {
...
   endpoints?: { notify: string, sessions: string };
   autoCaptureSessions?: boolean;
   notifyReleaseStages?: string[];
   releaseStage?: string;
-  maxEvents?: number;
   maxBreadcrumbs?: number;
-  consoleBreadcrumbsEnabled?: boolean;
-  networkBreadcrumbsEnabled?: boolean;
-  navigationBreadcrumbsEnabled?: boolean;
-  interactionBreadcrumbsEnabled?: boolean;
```
fractalwrench: I'm assuming that these options are no longer used/are specified elsewhere

packages/core/client.js
```diff
@@ -54,24 +55,45 @@ class BugsnagClient {
...
+    if (conf.metaData) this.metaData = conf.metaData
+    if (conf.user) this.user = conf.user
+    if (conf.logger) this.logger(conf.logger)
+
+    // merge with existing config
+    this.config = { ...this.config, ...conf }
+
     this._configured = true
+
     this._logger.debug(`Loaded!`)
```
bengourley: I realised that now `configure()` is called multiple times, this message needs to go somewhere else.

packages/core/client.js
```diff
@@ -54,24 +55,45 @@ class BugsnagClient {
     this.BugsnagSession = BugsnagSession
   }
 
-  configure (opts = {}) {
-    this.config = config.mergeDefaults({ ...this.config, ...opts }, this.configSchema)
-    const validity = config.validate(this.config, this.configSchema)
+  setOptions (opts) {
+    this._opts = { ...this._opts, ...opts }
```
fractalwrench: My JS is rusty, but would calling `setOptions` combine the passed in parameters with the existing values in `this._opts`. Is there any scenario where this might not be desirable?

fractalwrench: Or some scenario where duplicate options are passed in?

packages/plugin-window-onerror/test/onerror.test.js
bengourley: Ah thanks for the nudge. All of these commented out tests need to be converted into maze tests. I'll make a ticket for that.

packages/core/client.js
bengourley: Should be fine. It's an internal API only used by us, which is why I decided to prefix with the pseudo-private (by convention) underscore to denote as such.

packages/core/types/common.d.ts
bengourley: yeah these are all browser-specific so they're added in [here](https://github.com/bugsnag/bugsnag-js/pull/382/files#diff-3ef064881d5d7b974f81782769d32e82R3)

packages/core/client.js
bengourley: this merges whatever you pass in with whatever exists on `this._opts`. Passing in the same key multiple times means the last one will take precendence. In reality I think this will only ever get called once, but I wanted to be on the defensive side of things in case it would be useful to get called again with only a subset of options.

bengourley:
I updated the plugin interface so that `name` is optional. This saved `0.1kB` in the browser bundle size.

bengourley:
Something's up with the status check here. If you click through to the "pending" build it actually passed.

@kattrali please can I get a yes/no on whether this is ok to merge? Happy to make changes at a later stage, but currently all subsequent PRs are depending on this one.

bengourley:
Sorry, I missed this for a few days because it's on a closed PR!

> I know this is merged now, but are we 100% married to the configure method name? For me `configure` and `setOptions` sound really similar, and I'd almost expect `configure` to be the one that accepts user settings.

I'm not 100% married to it.

We could also even boil it back down to a single function that takes the `schema` and more optional user-supplied `options`.

```
client.configure({ schema [, options ] })
```

I can't see a single usage in either the notifiers or in the tests where something needs to happen between supplying user data and extracting options according to a schema.

bengourley:
What do you think would be the best option then?

```js
configure(userOptions)
setSchema(schema)
```

I'm ok with `configure()` there, but `setSchema()` doesn't sound right. The job of that function is to optionally take a schema, and extract and validate some user-supplied options and upgrade them into config. I feel like the a more appropriate name would be `parseOptions()` – which actually pairs nicely with `setOptions(opts)`.

So my counter suggestion if we want the methods to be separate is:

```js
setOptions(opts)
parseOptions([ schema ])
```

OR we roll them up into a single `configure()` call which takes both `options` and an optional `schema`.

cc/ @kattrali, @martin308 if you have the capacity to think about this!

loopj:
setOptions / parseOptions I think works for me

martin308:
My initial thoughts would be a single method makes sense. What are the benefits of separating them? It seems like it makes things more confusing

PR #384, opened 2018-08-14
Add @bugnsag/plugin-express

bengourley:
~__Note__:
This branch relies on changes in `u/plugin-config-interface`, so until that is merged this PR's diff includes all of those changes too. To see the actual changeset, see [this compare link](https://github.com/bugsnag/bugsnag-js/compare/u/plugin-config-interface...u/express-plugin).~

Update: #382 has now landed.

___

## What

This PR adds a plugin which provides express support, a la `bugsnag-node`. It exports two pieces of middleware which are used in the same way and whose functionality is equivalent, the only difference being the way the middleware is accesed. Internally the middleware has been simplified too, with less reliance on domains for context.

## How

Using the new plugin interface, this functionality is decoupled from the node notifier and will be installed separately:

```js
const express = require('express')
const bugsnag = require('@bugsnag/node')
const bugsnagExpress = require('@bugsnag/plugin-express')
const bugsnagClient = bugsnag('api_key').use(bugsnagExpress)

const { requestHandler, errorHandler } = bugsnagClient.getPlugin('express')
const app = express()
app.use(requestHandler)
// app routes here
app.use(errorHandler)
app.listen(process.env.PORT)
```

The configuration option `onUnhandledError` was added to the node notifier. Because the `requestHandler` middleware uses domains to trap some errors in the context of the request that caused them, it means the process doesn't crash as it would without Bugsnag. The default behaviour of `onUnhandledError` is to print the error (and any request context available) and then kill the process. In the docs we should encourage users to override this behaviour and ensure that their servers close out all existing connections (if possible) before killing the process.

**N.B #385 renames this to `onUncaughtException` to differentiate between that and `onUnhandledRejection`, having consistency with node's built-in event names.**

To faciliate the report being sent before the process is killed, a callback was added to `notify()`.

## Tests

~The feature tests for this plugin are working locally but not on Travis. Until we figure this out the CI check will not be green. These tests differ from others in that the docker service it starts needs to expose a port (the others don't). No matter how long the test waits, the network error `ECONNREFUSED` is encountered.~

Update: I figured it out. The CI should be green shortly.

### DIFF
--- a/packages/delivery-node/delivery.js
+++ b/packages/delivery-node/delivery.js
@@ -23,6 +23,7 @@ module.exports = () => ({
       })
     } catch (e) {
       logger.error(e)
+      cb(e)
     }
   },
   sendSession: (logger, config, session, cb = () => {}) => {
@@ -44,6 +45,7 @@ module.exports = () => ({
       })
     } catch (e) {
       logger.error(e)
+      cb(e)
     }
   }
 })
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -18,6 +18,17 @@ module.exports = {
   logger: {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
+  },
+  onUnhandledError: {
+    defaultValue: () => (err, report, logger) => {
+      const context = report.request
+        ? ` at ${report.request.httpMethod} ${report.request.path || report.request.url}`
+        : ``
+      logger.error(`Encountered an unhandled error${context}, terminating…\n${err ? err.stack : err}`)
+      process.exit(1)
+    },
+    message: 'should be a function',
+    validate: value => typeof value === 'function'
   }
 }
 
--- a/packages/plugin-express/README.md
+++ b/packages/plugin-express/README.md
@@ -1,0 +1,6 @@
+# @bugsnag/plugin-express
+
+A [@bugsnag/js](https://github.com/bugsnag/bugsnag-js) plugin for capturing errors in Express (and Connect) web applications.
+
+## License
+MIT
--- a/packages/plugin-express/src/request-info.js
+++ b/packages/plugin-express/src/request-info.js
@@ -12,17 +12,9 @@ module.exports = req => {
     httpVersion: req.httpVersion
   }
 
-  if (req.params && typeof req.params === 'object' && Object.keys(req.params).length > 0) {
-    request.params = req.params
-  }
-
-  if (req.query && typeof req.query === 'object' && Object.keys(req.query).length > 0) {
-    request.query = req.query
-  }
-
-  if (req.body && typeof req.body === 'object' && Object.keys(req.body).length > 0) {
-    request.body = req.body
-  }
+  request.params = extractObject(req, 'params')
+  request.query = extractObject(req, 'query')
+  request.body = extractObject(req, 'body')
 
   if (connection) {
     request.connection = {
@@ -36,4 +28,12 @@ module.exports = req => {
     }
   }
   return request
+}
+
+const extractObject = (host, key) => {
+  if (host[key] && typeof host[key] === 'object' && Object.keys(host[key]).length > 0) {
+    return host[key]
+  } else {
+    return undefined
+  }
 }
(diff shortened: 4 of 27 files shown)


packages/plugin-express/types/bugsnag-express.d.ts
```diff
@@ -0,0 +1,3 @@
+import * as Bugsnag from "@bugsnag/core";
```
fractalwrench: are wildcard imports conventional for JS? We discourage this style in our Java repos

packages/plugin-express/README.md
```diff
@@ -0,0 +1,6 @@
+# @bugsnag/plugin-express
+
+A [@bugsnag/js](https://github.com/bugsnag/bugsnag-js) plugin for capturing errors in Express (and Connect) web applications.
```
fractalwrench: Are we planning on adding an example integration at all? It may be helpful to add some more information on how to install the plugin here (or elsewhere) if not.

packages/delivery-node/delivery.js
```diff
@@ -23,6 +23,7 @@ module.exports = () => ({
       })
     } catch (e) {
       logger.error(e)
+      cb(e)
     }
   },
   sendSession: (logger, config, session, cb = () => {}) => {
```
fractalwrench: the value of `cb` could be `noop` by default, for additional consistency

packages/node/src/config.js
```diff
@@ -18,6 +18,17 @@ module.exports = {
   logger: {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
+  },
+  onUnhandledError: {
+    defaultValue: () => (err, report, logger) => {
```
fractalwrench: is `report.request` guaranteed to be defined at this point?

packages/plugin-express/src/request-info.js
```diff
@@ -0,0 +1,39 @@
...
+
+  if (req.params && typeof req.params === 'object' && Object.keys(req.params).length > 0) {
+    request.params = req.params
+  }
+
+  if (req.query && typeof req.query === 'object' && Object.keys(req.query).length > 0) {
+    request.query = req.query
+  }
+
+  if (req.body && typeof req.body === 'object' && Object.keys(req.body).length > 0) {
```
fractalwrench: this check could be extracted out into a function

packages/plugin-express/types/bugsnag-express.d.ts
bengourley: This is typescript, and tslint doesn't seem to complain about it – but since I'm only using the `IPlugin` type from this file I will change it to be `import { IPlugin } from "@bugsnag/core";`

packages/plugin-express/README.md
bengourley: The docs will be worked on in the coming weeks. I think it'd be a good idea to add examples to this repo as a part of that work.

packages/delivery-node/delivery.js
bengourley: Yeah I see I've been inconsistent with `() => {}` and `noop`. Might be good to do this in a single pass once all of the branches have converged again?

packages/node/src/config.js
bengourley: No, it could be `undefined`, but `report` is guaranteed to be defined, so accessing `report.request` is fine.

bengourley:
@fractalwrench I've pushed up some changes in relation to you feedback. Thanks!

packages/plugin-server-session/session.js
```diff
@@ -60,12 +61,12 @@ const sendSessionSummary = client => sessionCounts => {
   req(handleRes)
 
   function handleRes (err) {
-    if (!err) return client._logger.info('Session delivered')
+    if (!err) return client._logger.debug(`${sessionCounts.length} sessions reported`)
```
kattrali: This should be reworded somewhat. I was confused at first to send a dozen requests but then the console read "1 sessions reported", since it is counting the groups rather than the sessions.

bengourley:
> What is the scope of breadcrumbs?

This was a bug/oversight which I addressed here: https://github.com/bugsnag/bugsnag-js/commit/43c264661b601f176bbf49250cce07da8f915f41

kattrali:
> This was a bug/oversight which I addressed here: 43c2646

Ah! Thank you. I missed this.

PR #385, opened 2018-08-14
Add node global error handler plugins

bengourley:
__Note__:
Like #384, this branch relies on changes in both `u/plugin-config-interface` and `u/express-plugin`, so until they are merged this PR's diff includes all of those changes too. To see the actual changeset, see [this compare link](https://github.com/bugsnag/bugsnag-js/compare/u/express-plugin...u/node-global-error-handlers).

Update: #382 is now merged, just waiting on #384 now.

___

## What

This adds the two pieces of functionality required to track unhandled errors in node:

- `@bugsnag/plugin-node-uncaught-exception`
- `@bugsnag/plugin-node-unhandled-rejection`

Both are of course bundled into the node notifier by default – they don't need to be installed by end users.

## How

When `config.autoNotify` is `true`, the two plugins hook into the `process#uncaughtException` and `process#unhandledRejection` events respectively.

When either of these functions handle an error, they'll call the `config.onUnhandledError` function, which by default will terminate the process. The option `terminateOnUnhandledRejection` can be used to prevent an unhandled promise rejection from terminating the process.

## Review

~The tests on CI will fail due to a known issue :( they work locally though!
This purpose of this PR being pushed up at this point is for manual code review.~

The test suite is now fixed and the CI should be green.

### DIFF
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -18,6 +18,17 @@ module.exports = {
   logger: {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
+  },
+  onUnhandledError: {
+    defaultValue: () => (err, report, logger) => {
+      const context = report.request
+        ? ` at ${report.request.httpMethod} ${report.request.path || report.request.url}`
+        : ``
+      logger.error(`Encountered an unhandled error${context}, terminating…\n${err ? err.stack : err}`)
+      process.exit(1)
+    },
+    message: 'should be a function',
+    validate: value => typeof value === 'function'
   }
 }
 
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -21,7 +21,7 @@ module.exports = {
   },
   onUnhandledError: {
     defaultValue: () => (err, report, logger) => {
-      const context = report.request
+      const context = report.request && Object.keys(report.request).length
         ? ` at ${report.request.httpMethod} ${report.request.path || report.request.url}`
         : ``
       logger.error(`Encountered an unhandled error${context}, terminating…\n${err ? err.stack : err}`)
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -19,13 +19,17 @@ module.exports = {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
   },
-  onUnhandledError: {
+  onUncaughtException: {
     defaultValue: () => (err, report, logger) => {
-      const context = report.request && Object.keys(report.request).length
-        ? ` at ${report.request.httpMethod} ${report.request.path || report.request.url}`
-        : ``
-      logger.error(`Encountered an unhandled error${context}, terminating…\n${err ? err.stack : err}`)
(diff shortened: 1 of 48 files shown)


packages/plugin-node-uncaught-exception/test/uncaught-exception.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+describe('plugin: node uncaught exception handler', () => {
+  it('should listen to the process#uncaughtException event', () => {
+    const before = process.listeners('uncaughtException').length
+    const c = new Client(VALID_NOTIFIER)
+    c.setOptions({ apiKey: 'api_key' })
+    c.configure()
+    c.use(plugin)
+    const after = process.listeners('uncaughtException').length
+    expect(before < after).toBe(true)
```
fractalwrench: we should verify that this has increased by 1, unless this adds multiple listeners?

packages/plugin-node-uncaught-exception/test/uncaught-exception.test.js
```diff
@@ -0,0 +1,59 @@
...
+    plugin.destroy()
```
fractalwrench: general point - do we need to verify that destroying plugins deregisters listeners?

packages/plugin-node-uncaught-exception/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "homepage": "https://www.bugsnag.com/",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:bugsnag/bugsnag-js.git"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "test": "nyc --reporter=lcov -- jasmine '!(node_modules)/**/*.test.js'"
```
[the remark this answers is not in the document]
bengourley: Nope, this package gets folded in to the node notifier so it doesn't need to standalone – thus no need for it to be compiled in isolation.

packages/plugin-node-uncaught-exception/test/uncaught-exception.test.js
bengourley: `destroy()` is not a user facing feature of plugins – it's only really there so we can run multiple tests in the same process.

packages/node/src/config.js
```diff
@@ -18,6 +18,21 @@ module.exports = {
   logger: {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
+  },
+  onUncaughtException: {
+    defaultValue: () => (err, report, logger) => {
+      logger.error(`Reported an uncaught exception${getContext(report)}, the process will now terminate…\n${err ? err.stack : err}`)
```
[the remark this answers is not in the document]
bengourley: doh – this micro-optimisation is totally not important for this code, it's node-specific!

bengourley:
By way of merging #384 / #388, the majority of these changes are already now in `master`. I believe it's just https://github.com/bugsnag/bugsnag-js/pull/385/commits/dbb5efc0d11ca99271a5aae5fe22bd054b78601d.

PR #386, opened 2018-08-15
Add "filters" option to support removing sensitive properties from payload

bengourley:
### Why?

This is an existing feature of most notifiers, including bugsnag-node, but was not yet implemented in the browser notifier.

### What?

The option can be used like so:

```js
const bugsnagClient = bugsnag({
  apiKey: 'api_key',
  filters: [ 'password', 'access_token', 'refresh_token' ]
})
```

If not supplied, the default value is `[ 'password' ]`. __Should we add anything else to the default list?__

### How?

This feature is implemented as a configurable array of property names. The content of any property in the payload whose key name matches an entry in the array will be replaced with the string `[FILTERED]`.

This implementation – rather than traversing specific parts of the payload – will replace any matching properties anywhere. 

#### Known drawback

This means that if a value is provided such as 'events' or 'stacktrace', then the resulting payload will not be valid for the error reporting api. __Is this a problem?__

The advantage of doing it this way means that we can leverage the built in [`JSON.stringify(obj, REPLACER)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter#The_replacer_parameter) functionality, meaning:
- it's a simpler implementation than writing our own
- it's more performant (using the serialisation traversal to do the filtering means the data structure is only walked once)
- it results smaller output for the browser bundle

### DIFF
--- a/packages/browser/bin/minify
+++ b/packages/browser/bin/minify
@@ -4,5 +4,5 @@ cat - | ./node_modules/.bin/uglifyjs \
   --compress \
   --mangle \
   --ie8 \
-  --source-map "content=inline,url=$(basename $1).map" \
+  --source-map "includeSources,content=inline,url=$(basename $1).map" \
 --output $1
--- a/packages/browser/features/filters.feature
+++ b/packages/browser/features/filters.feature
@@ -14,7 +14,7 @@ Scenario Outline: "password" is filtered by default
       | type       |
       | script     |
 
-Scenario Outline: User setting can override defaults`
+Scenario Outline: User setting can override defaults
   When I navigate to the URL "/filters/<type>/b.html"
   And the test should run in this browser
   Then I let the test page run for up to 10 seconds
@@ -25,6 +25,37 @@ Scenario Outline: User setting can override defaults`
   And the event "user.password" equals "123456"
   And the event "user.secret" equals "[FILTERED]"
   And the event "metaData.details.api_key" equals "[FILTERED]"
+    Examples:
+      | type       |
+      | script     |
+
+Scenario Outline: it only removes properties from specific payload subtrees
+  When I navigate to the URL "/filters/<type>/c.html"
+  And the test should run in this browser
+  Then I let the test page run for up to 10 seconds
+  And I wait for 5 seconds
+  Then I should receive 1 request
+  And the request is a valid browser payload for the error reporting API
+  And the event "user.id" equals "21"
+  And the event "user.password" equals "123456"
+  And the event "user.secret" equals "[FILTERED]"
+  And the event "metaData.details.stacktrace" equals "[FILTERED]"
+  And the "file" of stack frame 0 ends with "c.html"
+    Examples:
+      | type       |
+      | script     |
+
+Scenario Outline: it works with regexes
+  When I navigate to the URL "/filters/<type>/d.html"
+  And the test should run in this browser
+  Then I let the test page run for up to 10 seconds
+  And I wait for 5 seconds
+  Then I should receive 1 request
+  And the request is a valid browser payload for the error reporting API
+  And the event "metaData.details0" equals "[FILTERED]"
+  And the event "metaData.details1" equals "[FILTERED]"
+  And the event "metaData.details2" equals "[FILTERED]"
+  And the event "metaData.detailsA" equals "aaaa"
     Examples:
       | type       |
       | script     |
--- a/packages/browser/features/filters.feature
+++ b/packages/browser/features/filters.feature
@@ -9,7 +9,7 @@ Scenario Outline: "password" is filtered by default
   Then I should receive 1 request
   And the request is a valid browser payload for the error reporting API
   And the event "user.id" equals "21"
-  And the event "user.password" equals "[FILTERED]"
+  And the event "user.password" equals "[Filtered]"
     Examples:
       | type       |
       | script     |
@@ -23,8 +23,8 @@ Scenario Outline: User setting can override defaults
   And the request is a valid browser payload for the error reporting API
   And the event "user.id" equals "21"
   And the event "user.password" equals "123456"
-  And the event "user.secret" equals "[FILTERED]"
-  And the event "metaData.details.api_key" equals "[FILTERED]"
+  And the event "user.secret" equals "[Filtered]"
+  And the event "metaData.details.api_key" equals "[Filtered]"
     Examples:
       | type       |
       | script     |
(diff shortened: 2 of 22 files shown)


martin308:
We did a bunch of discussion around filtering the entire payload when we redid [bugsnag-dotnet](https://github.com/bugsnag/bugsnag-dotnet). The conclusion was that it should only filter from `user`, `app`, `metaData`, `breadcrumbs`, `request`, and `device` sections of the payload. I'll see if I can dig up any of the conversation from the time

packages/core/config.js
```diff
@@ -90,6 +90,11 @@ module.exports.schema = {
         (accum, method) => accum && typeof value[method] === 'function',
         true
       ))
+  },
+  filters: {
+    defaultValue: () => [ 'password' ],
+    message: 'should be an array of strings',
+    validate: value => isArray(value) && value.length === filter(value, s => typeof s === 'string').length
```
fractalwrench: it'd be nice to cover this validation with a couple of unit tests if that hasn't already been done elsewhere.

loopj:
Since the worst case of a customer adding "stacktrace" or "events" to their filters is pretty catastrophic, and will likely cause a ton of confusion and wasted time for customer and support, I'd prefer if we stuck with the notifier library specification on this one and just filter user, app, metaData, breadcrumbs, request, and device.

You could potentially still use a payload-level JSON replacer function, as long as you mark the objects we wish to filter. Eg:

```javascript
// Mark the object
this.app = {
  _filter: true
}
```

loopj:
Another quick question - will this support "string contains" as well as "regex matches" filtering? That would be a nice plus, and AFAIK would be pretty easy to implement in JS. We have this on some other platforms but is not required by the notifier library spec.

bengourley:
Ok – based on your feedback @loopj I've pushed up a different implementation. It now accepts a list of strings/regexes and they are only apploed to the desired portions of the payload. Most of the heavy lifting is done in @bugsnag/safe-json-stringify as I still wanted to avoid additional traversals if at all possible.

Check out https://github.com/bugsnag/safe-json-stringify/pull/1.

packages/delivery-node/test/delivery.test.js
```diff
@@ -32,7 +32,8 @@ describe('delivery:node', () => {
       const payload = { sample: 'payload' }
       const config = {
         apiKey: 'aaaaaaaa',
-        endpoints: { notify: `http://0.0.0.0:${server.address().port}/notify/` }
+        endpoints: { notify: `http://0.0.0.0:${server.address().port}/notify/` },
+        filters: []
```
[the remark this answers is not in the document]
bengourley: these unit tests don't go via the full config schema and so don't get any default values, so every part of config that they depend on needs to be provided

packages/browser/features/filters.feature
```diff
@@ -0,0 +1,30 @@
+@filters
+Feature: Filtering sensitive content from payload
```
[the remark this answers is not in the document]
bengourley: The properties are already fairly nested in the payload.

packages/core/package.json
```diff
@@ -19,7 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@bugsnag/cuid": "^3.0.0",
-    "@bugsnag/safe-json-stringify": "^2.1.0",
+    "@bugsnag/safe-json-stringify": "3.0.0-0",
```
bengourley: once the PR on @bugsnag/safe-json-stringify is merged and published we should update to a non prerelease version

packages/core/config.js
```diff
@@ -90,6 +90,14 @@ module.exports.schema = {
         (accum, method) => accum && typeof value[method] === 'function',
         true
       ))
+  },
+  filters: {
+    defaultValue: () => [ 'password' ],
```
bengourley: should we add any others to this array by default?

bengourley:
> You could potentially still use a payload-level JSON replacer function, as long as you mark the objects we wish to filter.

It's not possible to do this inside a replacer function – in each call of the replacer, the only context you have is the direct parent of the property you are looking at, not the entire "path".

packages/core/config.js
martin308: I think keeping it minimal by default is good, so I'd leave it as this

packages/core/types/common.d.ts
```diff
@@ -16,6 +16,7 @@ export interface IConfig {
   user?: object | null;
   metaData?: object | null;
   logger?: ILogger | null;
+  filters?: string[];
```
martin308: Should this be `filters?: (string|RegExp)[];`?

bengourley: Yeah, or it needs to allow a mixture which I'm not sure that does? I'll fix it, thanks!

bengourley: your example was correct

packages/core/lib/json-payload.js
```diff
@@ -1,7 +1,20 @@
...
+  'events.[].metaData',
+  'events.[].user',
+  'events.[].breadcrumbs',
+  'events.[].request',
+  'events.[].device',
+  // session
+  'device',
+  'app',
+  'user'
+]
```
martin308: should these just be two separate arrays? We have the two separate methods that use this array and it seems like there is no sharing between the two sets?

packages/core/lib/json-payload.js
```diff
@@ -10,8 +23,12 @@ module.exports = report => {
...
     }
-    payload = jsonStringify(report)
+    payload = jsonStringify(report, null, null, { filterPaths, filterKeys })
     if (payload.length > 10e5) throw new Error('payload exceeded 1MB limit')
   }
   return payload
 }
+
+module.exports.session = (report, filterKeys) => {
+  return jsonStringify(report, null, null, { filterPaths, filterKeys })
```
martin308: should we check the payload size here as well?

packages/browser/bin/minify
```diff
@@ -4,5 +4,5 @@ cat - | ./node_modules/.bin/uglifyjs \
   --compress \
   --mangle \
   --ie8 \
-  --source-map "content=inline,url=$(basename $1).map" \
+  --source-map "includeSources,content=inline,url=$(basename $1).map" \
```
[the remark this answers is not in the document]
bengourley: No but I noticed it was wrong when trying to debug this changeset in development.

PR #388, opened 2018-08-17
Add @bugsnag/plugin-koa

bengourley:
__Usual caveat__:
This branch depends on some things added in #384 and #385, so until they are merged this PR's diff includes all of those changes too. To see the actual changeset, see [this compare link](https://github.com/bugsnag/bugsnag-js/compare/u/node-global-error-handlers...u/koa-plugin).

___

### What?

This PR adds Koa support via the "@bugsnag/plugin-koa" package. The usage is now inline with the express plugin, exporting `requestHandler` and `errorHandler` functions:

```js
const bugsnag = require('@bugsnag/js')
const bugsnagKoa = require('@bugsnag/plugin-koa')
const Koa = require('koa')

bugsnagClient.use(bugsnagKoa)
const { requestHandler, errorHandler } = bugsnagClient.getPlugin('koa')

const app = new Koa()

app.use(requestHandler)
// app middleware goes here
app.on('error', errorHandler)

app.listen(PORT)
```

### How?

The original implementation in bugsnag-node did not have documented session support, so the plugin implementation here embellishes the existing one. This means there is now an equivalent `requestHandler` to that of `@bugsnag/plugin-express` which deals with session tracking and capturing downstream errors with request context. The `errorHandler` is used to deal with any errors that get thrown before the `requestHandler` is run. Unlike express, all of the error handling in Koa is async function / promise based, so there's no dependency on `domain`s.

The `requestHandler` attaches a bugsnag "session client" to `ctx` so `ctx.bugsnag.notify()` can be used to report handled errors with the context of the request attached.

### Tests and review

The CI should be green. The unit tests and feature tests pass locally. I have also tested this with an example koa app running locally (screenshot attached). Looking for a full/final review!

image

### DIFF
--- a/packages/plugin-koa/src/koa.js
+++ b/packages/plugin-koa/src/koa.js
@@ -1,0 +1,64 @@
+const createReportFromErr = require('@bugsnag/core/lib/report-from-error')
+const extractRequestInfo = require('./request-info')
+
+const handledState = {
+  severity: 'error',
+  unhandled: true,
+  severityReason: {
+    type: 'unhandledErrorMiddleware',
+    attributes: { framework: 'Koa' }
+  }
+}
+
+module.exports = {
+  name: 'koa',
+  init: client => {
+    const requestHandler = async (ctx, next) => {
+      // Start a session whether sessions are used or not. We use this
+      // to store request-specific info, in case of any errors.
+      const sessionClient = client.startSession()
+
+      ctx.bugsnag = sessionClient
+
+      // extract request info and pass it to the relevant bugsnag properties
+      const requestInfo = extractRequestInfo(ctx)
+      sessionClient.metaData = { ...sessionClient.metaData, request: requestInfo }
+      sessionClient.request = {
+        clientIp: requestInfo.clientIp,
+        headers: requestInfo.headers,
+        httpMethod: requestInfo.httpMethod,
+        url: requestInfo.url,
+        referer: requestInfo.referer
+      }
+
+      try {
+        await next()
+      } catch (err) {
+        ctx.bugsnag.notify(createReportFromErr(err, handledState))
+        if (!ctx.response.headerSent) ctx.response.status = 500
+      }
+    }
+
+    const errorHandler = (err, ctx) => {
+      let metaData
+      let request
+      const requestInfo = extractRequestInfo(ctx)
+      metaData = { request: requestInfo }
+      request = {
+        clientIp: requestInfo.clientIp,
+        headers: requestInfo.headers,
+        httpMethod: requestInfo.httpMethod,
+        url: requestInfo.url,
+        referer: requestInfo.referer
+      }
+
+      let c = ctx.bugsnag
+      if (!c) {
+        c = client
+      }
+      c.notify(createReportFromErr(err, handledState), { metaData, request })
+    }
+
+    return { requestHandler, errorHandler }
+  }
+}
--- a/packages/plugin-koa/src/koa.js
+++ b/packages/plugin-koa/src/koa.js
@@ -40,23 +40,20 @@ module.exports = {
     }
 
     const errorHandler = (err, ctx) => {
-      let metaData
-      let request
-      const requestInfo = extractRequestInfo(ctx)
-      metaData = { request: requestInfo }
-      request = {
-        clientIp: requestInfo.clientIp,
-        headers: requestInfo.headers,
-        httpMethod: requestInfo.httpMethod,
-        url: requestInfo.url,
-        referer: requestInfo.referer
+      if (ctx.bugsnag) {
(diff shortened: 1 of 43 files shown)


packages/plugin-koa/src/koa.js
```diff
@@ -0,0 +1,64 @@
...
+      // extract request info and pass it to the relevant bugsnag properties
+      const requestInfo = extractRequestInfo(ctx)
+      sessionClient.metaData = { ...sessionClient.metaData, request: requestInfo }
+      sessionClient.request = {
+        clientIp: requestInfo.clientIp,
+        headers: requestInfo.headers,
+        httpMethod: requestInfo.httpMethod,
+        url: requestInfo.url,
+        referer: requestInfo.referer
+      }
```
martin308: Do we need to do all this work when there will be potentially not be an error that we need to send this information for? Would it be better to gather all this in the catch below?

packages/plugin-koa/src/koa.js
```diff
@@ -0,0 +1,64 @@
...
+      metaData = { request: requestInfo }
+      request = {
+        clientIp: requestInfo.clientIp,
+        headers: requestInfo.headers,
+        httpMethod: requestInfo.httpMethod,
+        url: requestInfo.url,
+        referer: requestInfo.referer
+      }
+
+      let c = ctx.bugsnag
```
martin308: if this exists then all of the above information will already have been collected, should we do this check earlier so that we don't need to build it again?

packages/plugin-koa/src/request-info.js
```diff
@@ -0,0 +1,25 @@
...
+    path: request.url,
+    httpMethod: request.method,
+    headers: request.headers,
+    httpVersion: request.httpVersion,
+    query: ctx.request.query,
+    connection: request.connection ? {
+      remoteAddress: request.connection.remoteAddress || request.ip,
+      remotePort: request.connection.remotePort,
+      bytesRead: request.connection.bytesRead,
+      bytesWritten: request.connection.bytesWritten,
```
martin308: in the request handler is this most likely to be incorrect at the time of the error as we get the information before presumably a lot of the koa stack has been run?

packages/node/features/fixtures/express/Dockerfile
```diff
@@ -0,0 +1,16 @@
+ARG NODE_VERSION=8
+FROM bugsnag_node_test:$NODE_VERSION
+
+RUN mkdir -p /usr/src/app
```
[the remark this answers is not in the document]
bengourley: Thanks for the `Dockerfile` tips! Theses are dotted around over all the different branches so I've made a note to do this in one pass once everything has converged.

packages/plugin-koa/src/koa.js
```diff
@@ -0,0 +1,64 @@
...
+  severity: 'error',
+  unhandled: true,
+  severityReason: {
+    type: 'unhandledErrorMiddleware',
+    attributes: { framework: 'Koa' }
+  }
+}
+
+module.exports = {
+  name: 'koa',
```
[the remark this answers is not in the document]
bengourley: The convention I've used for `plugin.name` is `camelCase` – though there aren't many (any?) plugins so far that have more than one word in their name, so you'd never know!

The other usage is in the `severityReason` part of the payload, which is a human-readable string rather than an identifier – which is why it's capitalized there.

packages/plugin-koa/src/koa.js
bengourley: This is a good point and I did consider it. I left it like this because the original bugsnag-node behaved in this way. The only downside I see is that in the case of an error some data might not be available? Otherwise it seems like the better thing to do.

packages/plugin-koa/src/request-info.js
bengourley: Good point, and it ties in with your other point abvout when we collect the data. Again this is how it was in bugsnag-node.

packages/plugin-koa/src/koa.js
bengourley: Ahhh I just took a deeper look at this.

The reason the data is attached to the `client` at this point is so that the user can call `req.bugsnag.notify()` and it has all of the data already captured. We could solve this using some higher level notifier feature like `beforeNotify` (as opposed to `beforeSend`) callbacks, to only attach the data when it is required. But really this is not much work in the grand scheme of things so I don't think it is a performance concern.

packages/plugin-koa/src/koa.js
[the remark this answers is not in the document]
martin308: Ah ok I think I see now. The inconsistency I saw was in relation to this line https://github.com/bugsnag/bugsnag-js/compare/u/node-global-error-handlers...u/koa-plugin#diff-e82ce309a31234c4c5c4ee751c80a643R9

PR #389, opened 2018-08-21
Add @bugsnag/plugin-restify

bengourley:
__Usual caveat__:
This branch depends on some things added in #384 and #385, so until they are merged this PR's diff includes all of those changes too. To see the actual changeset, see [this compare link](https://github.com/bugsnag/bugsnag-js/compare/908c8cb9cb9b202c91da70b26d128e964bf52c7b...u/restify-plugin).

___

### What?

This PR adds Restify support via the "@bugsnag/plugin-restify" package. The usage is now inline with the koa and express plugins, exporting `requestHandler` and `errorHandler` functions:

```js
const bugsnag = require('@bugsnag/js')
const bugsnagRestify = require('@bugsnag/plugin-restify')
const restify = require('restify')

bugsnagClient.use(bugsnagRestify)
const { requestHandler, errorHandler } = bugsnagClient.getPlugin('restify')

const server = restify.createServer()

server.pre(requestHandler)
// middleware and routes go here
server.on('restifyError', errorHandler)

app.listen(PORT)
```

### How?

The original implementation in bugsnag-node did not have documented session support, so the plugin implementation here embellishes the existing one. This means there is now an equivalent `requestHandler` to that of `@bugsnag/plugin-express` which deals with session tracking and capturing downstream errors with request context. The `errorHandler` is used to deal with any errors that don't make it – for whatever reason to the `requestHandler`. Internally, like express, the request handler wraps all of the downstream middleware/routes etc. in a `domain`, as that's currently the best way to do it. Currently [we recommended](https://docs.bugsnag.com/platforms/nodejs/restify/#3-install-the-restify-handler) switching restify into a [mode that did this](http://restify.com/docs/server-api/#uncaughtexception), but they are trying to [deprecate that](https://github.com/restify/node-restify/blob/d90fd4fb866d0a0c52ddd48b95514c9fad6e1eaf/docs/guides/4TO5GUIDE.md#deprecations). Doing this ourselves makes it more convenient to wrap it up with session tracking too, and have consistency between support for the various node frameworks.

The `requestHandler` attaches a bugsnag "session client" to `req` so `req.bugsnag.notify()` can be used to report handled errors with the context of another route/middleware.

### Tests and review

The CI should be green. The unit tests and feature tests pass locally. I have also tested this with an example Restify app running locally. Looking for a full/final review!

### DIFF
--- a/packages/plugin-restify/src/restify.js
+++ b/packages/plugin-restify/src/restify.js
@@ -64,7 +64,7 @@ module.exports = {
         c = client
       }
       c.notify(createReportFromErr(err, handledState))
-      cb(err)
+      cb()
     }
 
     return { requestHandler, errorHandler }
--- a/packages/plugin-restify/src/restify.js
+++ b/packages/plugin-restify/src/restify.js
@@ -1,0 +1,74 @@
+const domain = require('domain') // eslint-disable-line
+const extractRequestInfo = require('./request-info')
+const createReportFromErr = require('@bugsnag/core/lib/report-from-error')
+const handledState = {
+  severity: 'error',
+  unhandled: true,
+  severityReason: {
+    type: 'unhandledErrorMiddleware',
+    attributes: { framework: 'Restify' }
+  }
+}
+
+module.exports = {
+  name: 'restify',
+  init: client => {
+    const requestHandler = (req, res, next) => {
+      const dom = domain.create()
+
+      // Start a session whether sessions are used or not. We use this
+      // to store request-specific info, in case of any errors.
+      const sessionClient = client.startSession()
+
+      // attach it to the request
+      req.bugsnag = sessionClient
+
+      // extract request info and pass it to the relevant bugsnag properties
+      const requestInfo = extractRequestInfo(req)
+      sessionClient.metaData = { ...sessionClient.metaData, request: requestInfo }
+      sessionClient.request = {
+        clientIp: requestInfo.clientIp,
+        headers: requestInfo.headers,
+        httpMethod: requestInfo.httpMethod,
+        url: requestInfo.url,
+        referer: requestInfo.referer
+      }
+
+      // unhandled errors caused by this request
+      dom.on('error', (err) => {
+        req.bugsnag.notify(createReportFromErr(err, handledState), {}, (e, report) => {
+          if (e) return client._logger('Failed to send report to Bugsnag')
+          req.bugsnag.config.onUncaughtException(err, report, client._logger)
+        })
+        if (!res.headersSent) {
+          const body = 'Internal server error'
+          res.writeHead(200, {
+            'Content-Length': Buffer.byteLength(body),
+            'Content-Type': 'text/plain'
+          })
+          res.end(body)
+        }
+      })
+
+      return dom.run(next)
+    }
+
+    const errorHandler = (req, res, err, cb) => {
+      if (err.statusCode && err.statusCode < 500) return cb()
+      let c = req.bugsnag
+      if (!c) {
+        client._logger.warn(
+          'req.bugsnag is not defined. Make sure the @bugsnag/plugin-restify requestHandler middleware is added first.',
+          'Some request information will be missing.'
+        )
+        c = client
+      }
+      c.notify(createReportFromErr(err, handledState))
+      cb(err)
+    }
+
+    return { requestHandler, errorHandler }
+  }
+}
+
+module.exports['default'] = module.exports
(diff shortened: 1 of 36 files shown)


packages/plugin-restify/src/restify.js
```diff
@@ -0,0 +1,74 @@
...
+
+      // unhandled errors caused by this request
+      dom.on('error', (err) => {
+        req.bugsnag.notify(createReportFromErr(err, handledState), {}, (e, report) => {
+          if (e) return client._logger('Failed to send report to Bugsnag')
+          req.bugsnag.config.onUncaughtException(err, report, client._logger)
+        })
+        if (!res.headersSent) {
+          const body = 'Internal server error'
+          res.writeHead(200, {
```
martin308: 200 seems like the wrong status code to use here

packages/plugin-restify/src/restify.js
```diff
@@ -0,0 +1,74 @@
...
+        referer: requestInfo.referer
+      }
+
+      // unhandled errors caused by this request
+      dom.on('error', (err) => {
+        req.bugsnag.notify(createReportFromErr(err, handledState), {}, (e, report) => {
+          if (e) return client._logger('Failed to send report to Bugsnag')
+          req.bugsnag.config.onUncaughtException(err, report, client._logger)
+        })
+        if (!res.headersSent) {
```
martin308: why do we send a response here? It seems beyond the scope of error reporting. The Koa integration only sets the response code to 500 in this scenario

```
if (!ctx.response.headerSent) ctx.response.status = 500
```

packages/plugin-restify/src/restify.js
```diff
@@ -0,0 +1,74 @@
...
+      if (err.statusCode && err.statusCode < 500) return cb()
+      let c = req.bugsnag
+      if (!c) {
+        client._logger.warn(
+          'req.bugsnag is not defined. Make sure the @bugsnag/plugin-restify requestHandler middleware is added first.',
+          'Some request information will be missing.'
+        )
+        c = client
+      }
+      c.notify(createReportFromErr(err, handledState))
```
martin308: in the Koa integration we attach the req information when we are using a client that was not on the request which seems like a better pattern than logging a warning that we couldn't add the request information. Is that possible here?

packages/plugin-restify/src/restify.js
```diff
@@ -0,0 +1,74 @@
...
+          })
+          res.end(body)
+        }
+      })
+
+      return dom.run(next)
+    }
+
+    const errorHandler = (req, res, err, cb) => {
+      if (err.statusCode && err.statusCode < 500) return cb()
```
martin308: what situation will the error handler be called where we don't want to pass the `err` to the callback?

packages/plugin-restify/src/restify.js
bengourley: Good spot, thanks, it should be `500`.

packages/plugin-restify/src/restify.js
bengourley: We've caught an error which prevented a crash. Once the error is reported the process will be killed. It makes sense to me to send an HTTP error response.

The [express plugin does](https://github.com/bugsnag/bugsnag-js/pull/384/files#diff-ac05c4d39c4020ebd563acbdc91daefdR35). Setting the status code and no body has the same effect in Koa. It will look internally for the standard HTTP error message for the provided code.

packages/plugin-restify/src/restify.js
bengourley: > what situation will the error handler be called

I don't actually know. It's not apparent from restify's docs and I couldn't glean it from the amount I looked at the code. I _think_ passing an error to the `cb()` would indicate that an error happend _within_ this function.

packages/plugin-restify/src/restify.js
bengourley: It's possible. There's certainly inconsistency here – I this approach for restify and express but not koa.

My intention was for it to be clear that omiting the `requestHandler` is a usage error. When you don't add the `requestHandler` the following bad things happen:

- there's no `req.bugsnag()` for handled errors with request context
- there's no session tracking

It's possible to make all three web frameworks take the same approach, I think we just need to decide. My suggestion is the following (y not both!!!):

```js
const requestHandler = () => {
  // start a session
  // capture request info
  // add sesstion to req/ctx.bugsnag
}

const errorHandler = () => {
  // check for req/ctx.bugsnag
  // if it exists use its notify function
  // if it doesn't exist, warn
  // collect request info
  // notify with request info
}
```

kattrali: > It's possible to make all three web frameworks take the same approach, I think we just need to decide. My suggestion is the following (y not both!!!):

Consistency here seems ideal. Not using the `requestHandler` should log a warning, but following the warning we should collect as much as possible from the available environment and notify.

PR #392, opened 2018-08-24
Ensure expected types for fileName and lineNumber

funkjedi:
Currently if an unexpected type is used for fileName or lineNumber the underlying error-stack-parser lib throws an exception. Which in turn break/prevents the expected functionality of the app sans Bugsnag.

This casts the url and lineNo arguments to ensure error-stack-parser alway gets the expected types when setting the fileName and lineNumber on the culprit.

We ran into this issue on a site using the bootstrapValidator library. When a field has errors the library creates an event `{type: 'error.validator.bv'}` when they trigger event on the field with jQuery the event is modified to be `{type: 'error', namespace: 'validator.bv'}` resulting in `window.onerror`  being triggered with the source/url argument containing an object (validation results) and the lineno argument being undefined.

```js
// example
jQuery('select.province').trigger(jQuery.Event('error.validator.bv'), { valid: false });
```
For now we've just hooked `window.onerror` to workaround the issue but having a proper fix baked in would be nice.
```js
(function(w){
  var h = w.onerror;
  w.onerror = function(a,b,c,d,e){
  	if (typeof b === 'string') h(a,b,c,d,e);
  };
})(window);
```


bengourley:
Hi. This PR was closed automatically when I switched out a new branch to `master` which didn't share any history with the version of `master` this pointed at. I'm going to take a fresh look at this issue with the new code. Thanks!

PR #395, opened 2018-09-11
Ensure manual startSession() calls result in delivered sessions

bengourley:
In the implementation of the three server-side integrations:

- koa
- express
- restify

…I used the `client.startSession()` call regardless of whether auto session tracking was enabled. This was to provide a clone of a client that was scoped to a request. In the node session tracking plugin the `autoCaptureSessions` config value was checked to decide whether to send the sessions or not. The problem with this is that manual calls to `startSession()` were not delivered. Really, I had interpreted `autoCaptureSessions` as just `captureSessions` which is incorrect.

To solve this I factored out the clone logic from the node `startSession()` implementation which the server plugin now use if auto session tracking is not enabled.

I added a failing end to end test to confirm the problem which is fixed by this change.

### DIFF
--- a/packages/plugin-server-session/session.js
+++ b/packages/plugin-server-session/session.js
@@ -1,6 +1,7 @@
 const { isArray, includes } = require('@bugsnag/core/lib/es-utils')
 const inferReleaseStage = require('@bugsnag/core/lib/infer-release-stage')
 const { intRange } = require('@bugsnag/core/lib/validators')
+const clone = require('@bugsnag/core/lib/clone-client')
 const SessionTracker = require('./tracker')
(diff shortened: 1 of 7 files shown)


packages/plugin-server-session/session.js
```diff
@@ -48,8 +31,6 @@ module.exports = {
 const sendSessionSummary = client => sessionCounts => {
   const releaseStage = inferReleaseStage(client)
 
-  if (!client.config.autoCaptureSessions) return
```
bengourley: This was the nub of the issue! At this point we don't know if the sessions we are about to report are manually or automatically captured.

PR #400, opened 2018-10-26
Generate a correct error message when null is passed to notify()

bobjflong:
## Goal

Currently, calling `notify(null)` incorrectly generates this error: `Bugsnag usage error. notify() expected error/opts parameters, got unsupported object`.

The intent from the code appears to be that this would instead generate: `Bugsnag usage error. notify() expected error/opts parameters, got nothing`

The issue is caused by `normaliseError` calling `typeof error`, which is `object` for null. So this code path was being hit: https://github.com/bugsnag/bugsnag-js/compare/master...bobjflong:BL/null?expand=1#diff-d038e8e69f5ea867f38281dc1f40df19R240

There is existing [null check](https://github.com/bugsnag/bugsnag-js/compare/master...bobjflong:BL/null?expand=1#diff-d038e8e69f5ea867f38281dc1f40df19R131) code, but it wasn't being hit as a result.

### Changed

I've added a null check to the last branch in `normaliseError`.

## Tests

I've added a specific unit test for this case.

## Review

- [?] Commented on code changes inline explain the reasoning behind the approach - necessary?
- [?] A changelog entry was added for the goal of this pull request - necessary?
  - [ ] Initial review of the intended approach, not yet feature complete
  - [ ] Structural review of the classes, functions, and properties modified
  - [x] Final review


bengourley:
Hi. This PR was closed automatically when I switched out a new branch to `master` which didn't share any history with the version of `master` this pointed at. I'm going to take a fresh look at this issue with the new code. Thanks!

PR #403, opened 2018-11-05
Add process.versions to device data tab

bengourley:
image

The version of Node was not present anywhere in the payload – what's more there are a whole bunch of other versions available in the case that Node happened to be compiled with any different binary internals.

The contents of the `process.versions` hash is now added to the device data tab verbatim.

### DIFF
--- a/packages/plugin-node-device/device.js
+++ b/packages/plugin-node-device/device.js
@@ -5,7 +5,7 @@ const { isoDate } = require('@bugsnag/core/lib/es-utils')
  */
 module.exports = {
   init: (client) => {
-    const device = { hostname: client.config.hostname }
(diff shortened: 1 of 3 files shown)


packages/plugin-node-device/device.js
```diff
@@ -5,7 +5,7 @@ const { isoDate } = require('@bugsnag/core/lib/es-utils')
  */
 module.exports = {
   init: (client) => {
-    const device = { hostname: client.config.hostname }
+    const device = { hostname: client.config.hostname, versions: process.versions }
```
duncanhewett: This should be added under `metadata.device` as it's not a documented `device` field in the API.

bengourley:
Updated based on feedback
image

PR #405, opened 2018-11-07
Add Vue example project

bengourley:
Adds an example project for demonstrating the Vue integration. The project was generated with Vue CLI so the directory structure should be familiar, and it abstracts a bunch of stuff to do with webpack and other tooling. It also has a nice parity with create-react-app.

There are a couple more buttons in the Vue example – a watch error and an error within Vue's `nextTick` callback. Both of these get extra metaData indicating where the issue arose.

(There is also a commit that fixes some untidy things in the React example that I spotted while duplicating reusable bits)


Cawllec:
Above mentioned docker issue is a failed copy to clipboard from Vue when run in docker.

PR #407, opened 2018-11-08
Add Express example app

bengourley:
This adds an example for Express. The example server renders an HTML page with buttons to click which hit different routes on the server to trigger the errors.

- unhandled error in a route
- handled error in a route
- setting user info in a route

I also spotted an issue with the `header[s]Sent` property in the express plugin itself so fixed that.


examples/express/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Usage
+
+Clone the repo and `cd` into the directory of this example:
+
+```
+git clone git@github.com:bugsnag/bugsnag-js.git
+cd bugsnag-js/examples/express
+```
+
+Use the instructions below to run the application.
```
Cawllec: Can we amend this to include visiting the `/` route in the browser

examples/express/app.js
```diff
@@ -0,0 +1,35 @@
...
+})
+
+app.post('/add-info', (req, res, next) => {
+  req.bugsnag.user = { id: '123', name: 'jim' }
+  next(new Error('Cannot load Jim’s items'))
+})
+
+app.use(errorHandler)
+
+app.listen(9871)
```
Cawllec: We should have a callback log here to display the port the app is listening on.

examples/express/app.js
```diff
@@ -0,0 +1,35 @@
...
+
+app.use('/static', express.static(`${__dirname}/static`))
+
+const index = readFileSync('./views/index.html', 'utf8')
+app.get('/', (req, res) => {
+  res.send(index)
+})
+
+app.post('/crash', (req, res) => {
+  setTimeout(() => { throw new Error('Uh oh') })
```
Cawllec: Is there a way we can either recover or notify the user they need to restart the server on visiting this route?

Cawllec: Or have some indication that this route should be visited last?  Especially for the dockerised version

examples/express/app.js
```diff
@@ -0,0 +1,35 @@
...
+})
+
+app.post('/handled', (req, res) => {
+  throw new Error('Just a wobble')
```
bengourley: I just noticed that this isn't actually "handled" by Bugsnag's definition. The server stays "up" and the process does not crash, but the error is not handled by the user. We actually want two routes here:

```suggestion
app.post('/unhandled', (req, res) => {
  throw new Error('Just a wobble')
})

app.post('/handled', (req, res) => {
  req.bugsnag.notify(new Error('Could not connect to xyz service'))
  res.sendStatus(503)
})
```

examples/express/app.js
bengourley: Good point! I reorderd the buttons and put a note in the readme.

PR #411, opened 2018-11-10
guard String() in console-breadcrumbs plugin

bathos-wistia:
In the console-breadcrumbs plugin, `String()` is used on each argument to `console.log` without a guard.

This will throw when the argument is an object which doesn’t implement toString/@@toPrimitive (most commonly this would occur when objects have null prototype).

This fix does not address the fact that various parts of Bugsnag will throw if Symbol values are logged when this plugin is enabled. I believe that may be a broader issue concerning breadcrumbs metadata, not specific to the console-breadcrumbs plugin.


bengourley:
Hey @bathos-wistia, thanks for supplying this PR. I've made this change over on another branch, as the current codebase has been rejigged for universal JS support. I'm closing this in favour of that (#429) because what's currently in master is no longer in active development (`universal` will be switched over to be the "new" `master` this week).

I explored a couple of other options, including doing `Object.prototype.toString.call(val)` but arrived back at your suggestion as the best thing to do for console breadcrumbs specifically.

With regards to your other comment:

> This fix does not address the fact that various parts of Bugsnag will throw if Symbol values are logged when this plugin is enabled. I believe that may be a broader issue concerning breadcrumbs metadata, not specific to the console-breadcrumbs plugin.

I think this is something we will need to support in the serialization module we maintain our own fork of: [@bugsnag/safe-json-stringify](https://github.com/bugsnag/safe-json-stringify). If you have a concrete example that you can create a test or quick example of, it would be great if you could post it over on that repo. Much appreciated!

PR #416, opened 2018-11-14
fix(node): Remove unused and define missing config option

bengourley:
Remove unused and define missing config option.

`terminateOnUnhandledRejection` was never implemented (`onUnhandedRejection: () => {}` was implemented instead)

### DIFF
--- a/packages/node/types/bugsnag.d.ts
+++ b/packages/node/types/bugsnag.d.ts
@@ -10,7 +10,7 @@ declare module "@bugsnag/core" {
       onUnhandledRejection?: (err: any, report: Bugsnag.Report, logger: Bugsnag.ILogger) => void;
       proxy?: string;
       projectRoot?: string;
-      terminateOnUnhandledRejection?: boolean;
+      sendCode?: string;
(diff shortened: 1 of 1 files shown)


packages/node/types/bugsnag.d.ts
```diff
@@ -10,7 +10,7 @@ declare module "@bugsnag/core" {
       onUnhandledRejection?: (err: any, report: Bugsnag.Report, logger: Bugsnag.ILogger) => void;
       proxy?: string;
       projectRoot?: string;
-      terminateOnUnhandledRejection?: boolean;
+      sendCode?: string;
```
JacobMarshallPP: Shouldn't `sendCode` be a boolean?

https://github.com/bugsnag/bugsnag-js/blob/2863a379a6bdbffdbd857a2c9f534a07d49d1de3/packages/plugin-node-surrounding-code/surrounding-code.js#L34-L38

bengourley: Thanks JM :)

PR #417, opened 2018-11-14
feat(core): Allow client.notify to be passed directly to callbacks

bengourley:
This allows the notify function to work even when called without the bugsnag instance as the callee:

```js
var bugsnagClient = bugsnag('API_KEY')
var notify = bugsnag.notify

notify(new Error('derp'))
```

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -51,6 +51,12 @@ class BugsnagClient {
     this.BugsnagReport = BugsnagReport
     this.BugsnagBreadcrumb = BugsnagBreadcrumb
     this.BugsnagSession = BugsnagSession
+
+    var self = this
+    var notify = this.notify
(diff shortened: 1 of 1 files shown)


packages/core/client.js
```diff
@@ -51,6 +51,12 @@ class BugsnagClient {
     this.BugsnagReport = BugsnagReport
     this.BugsnagBreadcrumb = BugsnagBreadcrumb
     this.BugsnagSession = BugsnagSession
+
+    var self = this
+    var notify = this.notify
+    this.notify = function () {
+      return notify.apply(self, arguments)
```
bengourley: We could use `fn.bind()` if we didn't support IE8 😬

bengourley:
cc @bullmo!

PR #418, opened 2018-11-14
feat(node): Use NODE_ENV as releaseStage if set

bengourley:
Use the NODE_ENV environment variable as the release stage if it is set.


bengourley:
cc @CodeHex

PR #421, opened 2018-11-16
feat(plugin-express): Ensure express plugin supports connect too

bengourley:
When putting together the Nuxt example I noticed we used a couple of features of Express that Connect doesn't have:

- `res.sendStatus(n)`
- `req.host` or `req.hostname`

This PR keeps the same functionality but using Connect-compatible APIs.

### DIFF
--- a/packages/plugin-express/src/request-info.js
+++ b/packages/plugin-express/src/request-info.js
@@ -5,7 +5,9 @@ module.exports = req => {
   const address = connection && connection.address && connection.address()
   const portNumber = address && address.port
   const port = (!portNumber || portNumber === 80 || portNumber === 443) ? '' : `:${portNumber}`
-  const url = `${req.protocol}://${req.hostname || req.host}${port}${req.url}`
+  const protocol = typeof req.protocol !== 'undefined' ? req.protocol : (req.connection.encrypted ? 'https' : 'http')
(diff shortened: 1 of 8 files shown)


bengourley:
This CI job is flaking. The changeset is for server code but it's the browser tests on Opera which are flaking out! I ran the test locally on the same browser and they pass :/

kattrali:
Is there an integration test for connect?

bengourley:
Not an automated one yet. Only the Nuxt project I manually tried it on.

packages/plugin-express/src/request-info.js
```diff
@@ -5,7 +5,9 @@ module.exports = req => {
   const address = connection && connection.address && connection.address()
   const portNumber = address && address.port
   const port = (!portNumber || portNumber === 80 || portNumber === 443) ? '' : `:${portNumber}`
-  const url = `${req.protocol}://${req.hostname || req.host}${port}${req.url}`
+  const protocol = typeof req.protocol !== 'undefined' ? req.protocol : (req.connection.encrypted ? 'https' : 'http')
+  const hostname = (req.hostname || req.host || req.headers.host || '').replace(/:\d+$/, '')
```
snmaynard: does req.headers definitely exist?

bengourley: yeah it's defined on the node incoming http request object: https://nodejs.org/docs/latest/api/http.html#http_message_headers

PR #423, opened 2018-11-20
Types revamp

bengourley:
The changes in this PR mean that TypeScript users _only_ have to install `@bugsnag/js` – nothing extra – to get the JS and the types.

This is achieved by a script which copies the `@bugsnag/core` types into each of `@bugsnag/browser` and `@bugsnag/node`, with an update to the import paths for the `.d.ts` files in those modules. This happens at build time.

This PR also fixes the type def for the `client.use()` method.


packages/js/package.json
```diff
@@ -30,14 +30,11 @@
     "node",
     "types.d.ts"
   ],
-  "scripts": {
-    "postversion": "npm run build"
```
bengourley: this was on the wrong package, it should have been on node.

packages/node/package.json
```diff
@@ -13,29 +13,30 @@
...
   ],
   "scripts": {
     "clean": "rm -fr dist && mkdir dist",
-    "build": "npm run clean && npm run build:dist",
+    "bundle-types": "../../bin/bundle-types",
+    "build": "npm run clean && npm run build:dist && npm run bundle-types",
     "build:dist": "../../bin/bundle src/notifier.js --node --exclude=iserror,stack-generator,error-stack-parser,request,pump,byline --standalone=bugsnag | ../../bin/extract-source-map dist/bugsnag.js",
-    "test": "bundle exec maze-runner"
+    "test": "bundle exec maze-runner",
+    "postversion": "npm run build"
```
bengourley: The node assets were not built when each version was created. This was the cause.

PR #424, opened 2018-11-20
fix(node): Disable breadcrumbs on Node

bengourley:
This disables breacrumbs on Node until we have properly spec'd the automatic breadcrumbs, and how to scope them to requests/sessions.

### DIFF
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -19,6 +19,10 @@ module.exports = {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
   },
+  autoBreadcrumbs: {
+    ...schema.autoBreadcrumbs,
+    defaultValue: () => false
+  },
   releaseStage: {
     ...schema.releaseStage,
     defaultValue: () => process.env.NODE_ENV || 'production'
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -19,10 +19,6 @@ module.exports = {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
   },
-  autoBreadcrumbs: {
-    ...schema.autoBreadcrumbs,
-    defaultValue: () => false
-  },
   releaseStage: {
     ...schema.releaseStage,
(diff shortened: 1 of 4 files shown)


packages/node/src/notifier.js
```diff
@@ -45,6 +45,11 @@ module.exports = (opts, userPlugins = []) => {
 
   bugsnag._logger.debug(`Loaded!`)
 
+  bugsnag.leaveBreadcrumb = function () {
```
bengourley: I figured this would be the simplest approach, monkey patching the client method to log a warning.

packages/node/src/config.js
```diff
@@ -19,6 +19,10 @@ module.exports = {
     ...schema.logger,
     defaultValue: () => getPrefixedConsole()
   },
+  autoBreadcrumbs: {
+    ...schema.autoBreadcrumbs,
+    defaultValue: () => false
```
bengourley: This change is required so that _every_ notify() doesn't also cause the warning log due to this:
https://github.com/bugsnag/bugsnag-js/blob/b089283dcc27ffd18871fa308b465cbe87db06c5/packages/core/client.js#L215-L222

kattrali: Can autoBreadcrumbs be removed as a config option instead?

bengourley: The only drawback of that is it doesn't look like it's possible to remove that property from the TypeScript definition of the options interface

bengourley:
From checking how to remove/disable this in the types, I realised that the config augmentation was not working at all … it didn't complain because of this catch all line: https://github.com/bugsnag/bugsnag-js/blob/de7e4cc2c952947236cb71356c9461bae4c4ddd4/packages/core/types/common.d.ts#L20

I fiddled for some time trying to get a DRY solution, but there doesn't appear to be a way to merge properties of an imported interface. Nor when you extend can you remove or change the type signature of a property, only add.

I'm going to land this on the `u/types/revamp` branch, do some QA and figure out the CI failures.

PR #429, opened 2018-11-26
fix(plugin-console-breadcrumbs): String(val) can throw for some values

bengourley:
Wrap a try/catch around the call to `String(val)` in console breadcrumbs. This fix was supplied by
@bathos-wistia in a PR that targeted a different branch, thanks!

### DIFF
--- a/packages/plugin-console-breadcrumbs/console-breadcrumbs.js
+++ b/packages/plugin-console-breadcrumbs/console-breadcrumbs.js
@@ -9,7 +9,11 @@ exports.init = (client) => {
     console[method] = (...args) => {
       client.leaveBreadcrumb('Console output', reduce(args, (accum, arg, i) => {
         // do the best/simplest stringification of each argument
-        let stringified = String(arg)
+        let stringified = '[Unknown value]'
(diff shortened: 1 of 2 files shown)


packages/plugin-console-breadcrumbs/console-breadcrumbs.js
```diff
@@ -9,7 +9,11 @@ exports.init = (client) => {
     console[method] = (...args) => {
       client.leaveBreadcrumb('Console output', reduce(args, (accum, arg, i) => {
         // do the best/simplest stringification of each argument
-        let stringified = String(arg)
+        let stringified = '[Unknown value]'
```
Cawllec: Can this be set in the catch block?

bengourley: It can, yeah. What are the advantages of doing so?

Cawllec: It just seems more readable to me, rather than having a default and trying to assign, it would be trying to assign and having a fallback, which fits the intended purpose better.

PR #439, opened 2018-11-29
Fix README for accessing ErrorBoundary

jorilallo:
Not sure if the code in the README works as is (didn't test) but it's at least against typescript definitions and differs from the documented usage: https://docs.bugsnag.com/platforms/javascript/react/


bengourley:
Thanks!

PR #443, opened 2018-12-05
window.onerror handler improvements

### LINKED ISSUE #393: Error in Microsoft Edge 18+ "lineNumber must be a Number"
### Expected behavior
Correctly log error w/ stacktrace.

### Observed behavior
Errors out at the line in the below screenshot. 

### Steps to reproduce
This seems to only be happening on Microsoft Edge 18+ versions and no other browsers.

### Version
4.7.2

### Additional information
screen shot 2018-08-29 at 11 19 14 am

bengourley:
A couple of related fixes:

c8ace7c: fix(plugin-window-onerror): Make decorateStack more robust

Only supply values that are of the expected type, as the stacktrace module can throw if given input
of the wrong kind.

Fixes #393.

---

7087fe9: fix(plugin-window-onerror): Support additional onerror call signature

For some events triggered by jQuery, the second argument to onerror can be an object. This
previously caused the onerror handler to misidentify the kind of error that had happened. This fix
amends the check on the "url" argument, only determining it to be a url if it is defined and is a
string.

Fixes #392.

### DIFF
--- a/packages/plugin-window-onerror/onerror.js
+++ b/packages/plugin-window-onerror/onerror.js
@@ -44,14 +44,16 @@ module.exports = {
 const decorateStack = (stack, url, lineNo, charNo) => {
   const culprit = stack[0]
   if (!culprit) return stack
-  if (!culprit.fileName) culprit.setFileName(url)
-  if (!culprit.lineNumber) culprit.setLineNumber(lineNo)
+  if (!culprit.fileName && typeof url === 'string') culprit.setFileName(url)
+  if (!culprit.lineNumber && isActualNumber(lineNo)) culprit.setLineNumber(lineNo)
   if (!culprit.columnNumber) {
-    if (charNo !== undefined) {
+    if (isActualNumber(charNo)) {
       culprit.setColumnNumber(charNo)
-    } else if (window.event && window.event.errorCharacter) {
-      culprit.setColumnNumber(window.event && window.event.errorCharacter)
+    } else if (window.event && isActualNumber(window.event.errorCharacter)) {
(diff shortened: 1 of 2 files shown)


packages/plugin-window-onerror/onerror.js
```diff
@@ -20,11 +20,11 @@ module.exports = {
             report = new client.BugsnagReport('window.onerror', String(error), decorateStack(client.BugsnagReport.getStacktrace(error, 1), url, lineNo, charNo), handledState)
             report.updateMetaData('window onerror', { error })
           }
-        } else if ((typeof messageOrEvent === 'object' && messageOrEvent !== null) && !url && !lineNo && !charNo && !error) {
+        } else if ((typeof messageOrEvent === 'object' && messageOrEvent !== null) && (!url || typeof url !== 'string') && !lineNo && !charNo && !error) {
```
snmaynard: The line length in this file is getting pretty long. Do we not have coding standards for that in the js repo? Should we use prettier or something?

This boolean logic is also pretty complex, and there is very little commenting in the file. Why do we just do `!url` and `messageOrEvent is !== null` explicitly?

snmaynard: It seems like we should comment saying why we are checking urls type here (jquery). We probably also want to link to https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

bengourley: It's certainly not a pretty file. We use [standard](https://standardjs.com/rules.html) for style rules which is not opinionated about line length.

I'll add some comments and see what I can do about the line length without having an adverse effect on output file size.

PR #450, opened 2018-12-18
fix(node): Flatten nested scopes in output to support bundling with webpack

### LINKED ISSUE #441: node.js module (@bugsnag/node) does not allow to be bundled
### Reproducible code:

https://github.com/kwonoj/bugsnag-node-webpack

`npm install && npm start`

will raise exception around module resolution:

```
Error: Cannot find module 'error-stack-parser'
    at o (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:292)
    at /Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:449
    at Object.20../lib/es-utils (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:1229:24)
    at o (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:398)
    at /Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:449
    at Object.3../breadcrumb (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:164:21)
    at o (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:398)
    at /Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:449
    at Object.24../config (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:1638:14)
    at o (/Users/ojkwon/github/bugsnag-node-webpack/dist/index.bundle.js:109:398)
```

This looks like distributed module `@bugsnag/node` is already bundled somehow, cannot resolve module by consumer application's bundler configuration.

### Motivation

Yes, typically node.js side module is not intended to be bundled. While reproducible code is provided as targeting `node` in example, actual usecases for our application is targeting `electron-main`, which inherits node.js nature blocks to create application bundle including bugsnag but have to specify as external. We would like to remove externals as it defeats our purpose of bundling.

### Possible approaches

If it's possible, I'd like to suggest to publish module with non-bundled code. This can be non-breaking changes if publishing esm module for those purpose (using `package.json`'s `module` field) while preserving existing main field with current bundled modules.

bengourley:
The dist node bundle format did not support being consumed in a webpack bundle, because there was too much nesting/dynamicism in the require() calls. Applying the browser-pack-flat plugin results in fewer nested scopes such that only the external require()s are left – and these get properly picked up by webpack.

Fixes #441.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -85,7 +85,7 @@ steps:
     env:
       COMPOSE_PROJECT_NAME: node4
       NODE_VERSION: "4"
-    command: '-e koa.feature -e koa-1x.feature'
+    command: '-e koa.feature -e koa-1x.feature -e webpack.feature'
 
   - label: ':node: Node 6'
     plugins:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.x.x (TBD)
+
+### Fixed
+
+- Support @bugsnag/node being consumed in a Webpack bundle for Node (#450, #441)
+
 ## 5.0.2 (2018-12-06)
 
 ### Fixed
--- a/bin/bundle
+++ b/bin/bundle
@@ -14,10 +14,10 @@ const standalone = (args.find(arg => /^--standalone=/.test(arg)) || '').replace(
 const b = browserify(entry, { debug: true, standalone, node: forNode })
   .transform('babelify', { global: true, ...babelConfig })
   .transform('browserify-versionify')
+  .plugin('browser-pack-flat/plugin')
 
 if (!forNode) {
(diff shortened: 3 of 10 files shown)


.buildkite/pipeline.yml
```diff
@@ -85,7 +85,7 @@ steps:
     env:
       COMPOSE_PROJECT_NAME: node4
       NODE_VERSION: "4"
-    command: '-e koa.feature -e koa-1x.feature'
+    command: '-e koa.feature -e koa-1x.feature -e webpack.feature'
```
bengourley: [Webpack does not support node v4](https://github.com/webpack/webpack/wiki/Changelog-WIP#big-changes)

PR #462, opened 2019-01-15
feat(delivery-node): Remove dependency on request 

bengourley:
Request is a huge dependency, both in terms of size and number of transient dependencies, compared with the rest of the node notifier.

This PR replaces the sending of error reports via request with the built-in http/https modules, and exposes the `agent` option to faciliate sending via proxy (which is the only feature of request that we relied on).

Still todo:
- [ ] more unit tests on the `delivery-node` package to test the multiple code paths that have been introduced
- [ ] changelog entry


bengourley:
buildkite doesn't appear to like `/` in the branch name :'(

PR #464, opened 2019-01-16
feat(delivery-node): Remove dependency on request 

bengourley:
Request is a huge dependency, both in terms of size and number of transient dependencies, compared with the rest of the node notifier.

This PR replaces the sending of error reports via request with the built-in http/https modules, and exposes the `agent` option to faciliate sending via proxy (which is the only feature of request that we relied on).

Still todo:
- [x] more unit tests on the `delivery-node` package to test the multiple code paths that have been introduced
- [x] changelog entry


bengourley:
This is now ready for review.

Question… do we want to add some code to warn people if they use the – now removed – `proxy` option? At the moment it will silently fail.

packages/delivery-node/request.js
```diff
@@ -0,0 +1,43 @@
...
+    hostname: parsedUrl.hostname,
+    port: parsedUrl.port,
+    path: parsedUrl.path,
+    headers
+  })
+  req.on('error', onError)
+  req.on('response', res => {
+    bufferResponse(res, (err, body) => {
+      if (err) return onError(err)
+      if (res.statusCode !== 200) {
```
bullmo: The session api returns 202 on success rather than 200 - https://bugsnagsessiontrackingapi.docs.apiary.io/#reference/0/session/report-a-session-starting

snmaynard: we should check if its 2xx

bengourley:
Closing this for a rebased branch.

PR #470, opened 2019-01-24
fix(browser): Dedupe external deps from bundle

bengourley:
`@bugsnag/core` and `@bugsnag/plugin-window-unhandled-error` both had local copies of the same external modules in their node_modules.

This meant `error-stack-parser` and `iserror` got included twice in the output bundle. This commit exposes the modules from inside `@bugsnag/core` and the unhandled rejection plugin references it from there, so they are sure to get only one and the same module.

|   |     Before    |  After     |  Diff |
|------|---------|--------|----|
| min | 44.01kB | 39.15kB |  -11.0%  |
| min + gz | 11.83kB | 11.68kB |  -1.2%  |


bengourley:
I'm going to reopen this with a different branch name. The `/` doesn't work with the test container registry.

PR #478, opened 2019-02-07
feat(delivery-node): Remove dependency on request (rebased)

bengourley:
Same as #464 but rebased now that the CI overhaul is merged.


Cawllec:
Tested this, and like the previous PR works fine :+1:

PR #485, opened 2019-02-22
Add React native error handler plugin

bengourley:
This adds the `@bugsnag/plugin-react-native-global-error-handler` and includes it by default in the `@bugsnag/expo` notifier.

Along with this addition, there's a sample Expo app in the top level `test` directory which I'm using to test things out manually, and the necessary scripts to get a dev version of the notifier into that app.

The guts of the plugin are lifted from [bugsnag-react-native](https://github.com/bugsnag/bugsnag-react-native/blob/98045e152397295ebca610f21fc872d802af764d/src/Bugsnag.js#L47-L63). Due to fact that we'll be sending errors from the JS layer I didn't bring the arbitrary timeout, but I did chose to only call through to the previous handler once the notify() method has called back.

There are unit tests for this functionality which mock the global `ErrorUtils` object, and I used the sample app to manually test it aswell.


packages/plugin-react-native-global-error-handler/error-handler.js
```diff
@@ -0,0 +1,22 @@
+/*
+* Automatically notifies Bugsnag when React Native's global error handler is called
+*/
+
+const createReportFromErr = require('@bugsnag/core/lib/report-from-error')
+
+module.exports = {
+  init: (client, ErrorUtils = global.ErrorUtils) => {
+    if (!client.config.autoNotify) return
```
kattrali: Would `ErrorUtils` be present in end users' test environments? Is it something that would require a null check?  I ask because I had this check previously in `bugsnag-react-native` and don't remember why. 😅

packages/expo/package.json
```diff
@@ -47,7 +47,8 @@
     "typescript": "^3.3.3"
   },
   "dependencies": {
-    "@bugsnag/core": "^5.2.0",
+    "@bugsnag/core": "^6.0.0",
+    "@bugsnag/plugin-react-native-global-error-handler": "^5.2.0",
```
kattrali: What's up with the versioning differing from core?

bengourley: Hmm, just a hangover from starting this branch before 6.0.0 went out I think. It doesn't really matter because it's just linking stuff up locally and upon publish it'll get a proper version, but I'll sync it up now for tidiness.

packages/plugin-react-native-global-error-handler/error-handler.js
bengourley: > Would `ErrorUtils` be present in end users' test environments?
I'm not sure.

I don't see a reason not to add a check and a log warning if it's not defined so I'll do that.

PR #493, opened 2019-03-12
feat(plugin-expo-device): Add plugin to capture Expo-specific device data for the payload

bengourley:
This PR adds a plugin that collects the following `device` info:

- `manufacturer`¹
- `modelName`¹
- `modelNumber`¹
- `osName`
- `osVersion`
- `time`
- [`id`](https://docs.expo.io/versions/v32.0.0/sdk/constants/#constantsinstallationid)
- `orientation`

<small>¹ only on Apple devices</small>

The following `metaData.device` info:

- [`isDevice`](https://docs.expo.io/versions/v32.0.0/sdk/constants/#constantsisdevice)
- [`appOwnership`](https://docs.expo.io/versions/v32.0.0/sdk/constants/#constantsappownership)

And the following `runtimeVersions`:

- `reactNative`
- [`expoApp`](https://docs.expo.io/versions/v32.0.0/sdk/constants/#constantsexpoversion)
- [`expoSdk`](https://docs.expo.io/versions/v32.0.0/workflow/how-expo-works/#sdk-versions)
- `androidApiLevel`

### DIFF
--- a/packages/plugin-expo-device/LICENSE.txt
+++ b/packages/plugin-expo-device/LICENSE.txt
@@ -1,0 +1,19 @@
+Copyright (c) 2018 Bugsnag, https://www.bugsnag.com/
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the Software
+is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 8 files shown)


packages/plugin-expo-device/LICENSE.txt
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2018 Bugsnag, https://www.bugsnag.com/
```
Cawllec: It probably doesn't matter too much, but 2019?

packages/plugin-expo-device/device.js
```diff
@@ -0,0 +1,52 @@
...
+    updateOrientation()
+
+    client.device = {
+      ...client.device,
+      id: Constants.installationId,
+      manufacturer: Constants.platform.ios ? 'Apple' : undefined,
+      modelName: Constants.platform.ios ? Constants.platform.ios.model : undefined,
+      modelNumber: Constants.platform.ios ? Constants.platform.ios.platform : undefined,
+      osName: Platform.OS,
+      osVersion: Constants.platform.ios ? Constants.platform.ios.system : Constants.systemVersion,
```
Cawllec: Is this correct?  Docs wise it seems the calls should be `Constants.platform.ios.systemVersion` and `Platform.Version`

packages/plugin-expo-device/LICENSE.txt
bengourley: Good spot, thanks!

packages/plugin-expo-device/device.js
bengourley: Ah, yeah you're right. I'll update this.

Did you also run it and log out payloads? This is unit tested, but all the external stuff is mocked, so like with this example if both the tests and impl are wrong in the same way we get a false positive. It would be good to have some payloads from different devices (even if they are simulators) eyeballed until we have proper e2e tests.

PR #500, opened 2019-03-19
feat(expo): Allow apiKey to be set in app.json rather than in code

bengourley:
This PR adds support for configuring the `apiKey` using the [`extra`](https://docs.expo.io/versions/v32.0.0/workflow/configuration/#extra) key in `app.json`, for example:

```json
{
  "extra": {
    "bugsnag": {
      "apiKey": "API_KEY"
    }
}
```

There are no unit tests for the standalone notifiers, as they pull together everything and would require an absurd amount of mocking. This one needs manually testing for now.

### DIFF
--- a/packages/expo/src/notifier.js
+++ b/packages/expo/src/notifier.js
@@ -22,6 +22,7 @@ module.exports = (opts) => {
 
   // attempt to fetch apiKey from app.json if we didn't get one explicitly passed
   if (!opts.apiKey &&
+    Constants.manifest &&
     Constants.manifest.extra &&
     Constants.manifest.extra.bugsnag &&
     Constants.manifest.extra.bugsnag.apiKey) {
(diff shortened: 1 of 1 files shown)


packages/expo/src/notifier.js
```diff
@@ -15,6 +16,18 @@ module.exports = (opts) => {
   // handle very simple use case where user supplies just the api key as a string
   if (typeof opts === 'string') opts = { apiKey: opts }
 
+  // ensure opts is actually an object (at this point it
+  // could be null, undefined, a number, a boolean etc.)
+  opts = { ...opts }
+
+  // attempt to fetch apiKey from app.json if we didn't get one explicitly passed
+  if (!opts.apiKey &&
+    Constants.manifest.extra &&
```
fractalwrench: this may be overly paranoid, but can we trust that Constants.manifest will be non-null?

Cawllec: An expo app should always have a `app.json` manifest, but I can't see it hurting having a null check

bengourley:
I've added the check. If `Constants.manifest` isn't defined then there are probably way bigger issues, but at least we wouldn't be the thing surfacing that problem.

PR #504, opened 2019-03-25
Add Expo postPublish hook

bengourley:
`./hooks/post-publish.js` will:

- report a build
- upload iOS and Android source maps

The `postPublish` hook is run in the following situations:

- `expo publish` (publishing a new JS bundle)
- `expo build:[ios|android]` (building a new standalone app for a particular platform)

The following config needs to be added to `app.json` to get Expo to run this hook:

```js
{
  "hooks": {
    "postPublish": [
      {
        "file": "@bugsnag/expo/hooks/post-publish",
        "config": {
          "apiKey": "________" // <optional>,
          "reportBuild": true // <optional default=true>,
          "uploadSourceMaps": true // <optional default=true>
        }
      }
    ]
  }
}
```

Adding this config will be automated by a subsequent cli PR. If `apiKey` is not present in the hook config, it will look in the`"extra"` portion of `app.json`. This is the preferred place to put it, as the notifier can find it from this location too:

```js
{
  "extra": {
    "bugsnag": {
      "apiKey": "_________"
    }
  }
}
```

### How we treat native `versionCode` and `bundleVersion` in Expo

When reporting a build, we don't know in the `postPublish` hook if we're building a standalone version or a just a new JS bundle. What's more, from the user's perspective, they are only really creating one release. For this reason, we don't include `versionCode` or `bundleVersion` in the build payload.

Subsequently when the notifier is running, to prevent those builds (appVersion + native version) from being automatically created, we move the `versionCode|bundleVersion` data into `app.metaData` prefixed with `native*` so that it doesn't get migrated by the pipeline.

### Source map lookup

Source maps are matched on stack frame location and `codeBundleId` (which we use Expo's `revisionId` for). On iOS, the stack frame location looks like:

`https://d1wp6m56sqw74a.cloudfront.net/%40bengourley-bugsnag%2Fmy-expo-project%2F2.0.0%2F31439685b360e74216b03f3c164d1bba-32.0.0-ios.js`

but on Android it looks like:

`/data/user/0/host.exp.exponent/files/32.0.0/cached-bundle-experience-%40bengourley-bugsnag%2Fmy-expo-project-1239374583-32.0.0`

Because it doesn't have a file extension, we're waiting on a pipeline update to ensure that it looks for a source map (currently it will only do this for frames with a `.js`, `.bundle` or `.jsbundle` extension – which means it's only possible to test Expo source maps on iOS at this point.


bengourley:
One more thing of note – I started publishing super unstable pre alpha releases to npm as a way of installing the development notifier inside of an app (it's quicker than attempting to do it any other way). You can install `@bugsnag/expo@v7.0.0-pre-alpha-ben.1` to test out this PR (excluding one tiny subsequent chore commit).

packages/expo/hooks/lib/report-build.js
```diff
@@ -0,0 +1,10 @@
+const reportBuild = require('bugsnag-build-reporter')
+
+module.exports = async (apiKey, manifest, projectRoot) => {
+  const { revisionId, version } = manifest
+  await reportBuild({
+    apiKey,
+    appVersion: version,
+    metadata: { [`bundle@${new Date().toISOString()}`]: revisionId }
```
bengourley: This means you get a list of `revisionId`s that were shipped under this app version on the release page:

image

packages/expo/hooks/lib/upload-source-maps.js
```diff
@@ -0,0 +1,54 @@
...
+
+  await promisify(writeFile)(androidSourceMapPath, androidSourceMap, 'utf-8')
+  await promisify(writeFile)(androidBundlePath, androidBundle, 'utf-8')
+
+  // android
+  console.log(`Uploading Android source map`)
+  await upload({
+    apiKey,
+    appVersion: androidManifest.version,
+    minifiedUrl: '*/cached-bundle-experience-*',
```
bengourley: Android stack frames don't contain the CDN url, as the bundle is downloaded and saved to disk at a path which includes `/cached-bundle-experience`.

packages/expo/hooks/lib/upload-source-maps.js
```diff
@@ -0,0 +1,54 @@
...
+) => {
+  const dir = await makeTmpDir()
+
+  const androidBundlePath = join(dir, 'android.bundle.js')
+  const androidSourceMapPath = join(dir, 'android.bundle.js.map')
+
+  const iosBundlePath = join(dir, 'ios.bundle.js')
+  const iosSourceMapPath = join(dir, 'ios.bundle.js.map')
+
+  await promisify(writeFile)(iosSourceMapPath, iosSourceMap, 'utf-8')
```
kattrali: More of a question than a critique, but is `promisify(f)` impure  in some way? Reading this, I wasn't certain if these two were equivalent:

```js
await promisify(writeFile)(iosSourceMapPath, iosSourceMap, 'utf-8')
await promisify(writeFile)(iosBundlePath, iosBundle, 'utf-8')
```

```js
const asyncWriteFile = promisify(writeFile)
await asyncWriteFile(iosSourceMapPath, iosSourceMap, 'utf-8')
await asyncWriteFile(iosBundlePath, iosBundle, 'utf-8')
```

packages/expo/hooks/post-publish.js
```diff
@@ -15,9 +15,29 @@ module.exports = async ({
   exp
 }) => {
   try {
-    log('beep boop')
-    log(url)
-    log(exp)
+    let apiKey
+    if (config && config.apiKey) {
+      apiKey = config.apiKey
```
kattrali: Discussed this further in person, but having the API key potentially in two places in `app.json` seems confusing, particularly when we can look up the key (and write it, via the CLI) in the default location. There should be just one place for the API key and the location in extras is already recommended/in use.

packages/expo/hooks/lib/upload-source-maps.js
bengourley: Those examples would be functionally equivalent. Want me to change it over so we're not repeatedly calling `promisify()` on the same function?

kattrali: Yeah, its a bit more readable to me 👍

Issue #449, opened 2018-12-17
notifyReleaseStages prevents exception logging to console


jgentes:
If I use `notifyReleaseStages` to prevent sending reports when we're running in dev, it also swallows the errors in the console. All I see is:

```
bugsnag.js:1581 Loaded!
bugsnag.js:281 Report not sent due to releaseStage/notifyReleaseStages configuration
```

For now we're reporting all exceptions, even in dev as a workaround, and we're overloading our quota. Could really use a fix for this one.

snmaynard:
You can use a beforeNotify function that returns false in dev after printing the error in the console (as an interim measure).

mfouad:
I think the problem is that the default onUncoughtExpection is not called in that case
or any callback sent to notify() in that case

```
if (isArray(this.config.notifyReleaseStages) && !includes(this.config.notifyReleaseStages, releaseStage)) {
      this._logger.warn("Report not sent due to releaseStage/notifyReleaseStages configuration");
      return false; // ISSUE: should call cb() here 
}
```

snmaynard:
I wouldnt use notifyReleaseStages in that case. You can just return false in dev yourself from the callback

mfouad:
> You can use a beforeNotify function that returns false in dev after printing the error in the console (as an interim measure).
I think you mean beforeSend in the new library. From the code it seems wont be called in this case too. so no luck :disappointed:

mfouad:
> I wouldnt use notifyReleaseStages in that case. You can just return false in dev yourself from the callback

I think you mean beforeSend in the new library. From the code it seems wont be called in this case too. so no luck :disappointed:

jgentes:
I've tried to get this to work, but the bugsnag report isn't great if just logged to the console. I've tried creating a real Error object out of it, but haven't had much success:

```
beforeSend: function (report) {
    if (Env.DEV) {
      // create an object that has Error properties
      report = {
        message: report.errorMessage,
        stack: report.stacktrace
      }
      // try to log this to console as a real exception
      try {
        throw(report);
      } catch(e) {
        return false;
      }
    }
```

Unfortunately the throw never makes it to the console, perhaps because it ends up getting swallowed in the same loop? Not sure, but any ideas on how to convert the bugsnag report object into a real error would be appreciated.

snmaynard:
[...]

PR #515, opened 2019-04-04
Improve logging and available information when error reports are not sent

### LINKED ISSUE #420: v4 - access to original exception in beforeSend
### Desired Behavior & Use Case

Currently in `v4.x`, the `beforeSend` callback receives a [report object](https://docs.bugsnag.com/platforms/browsers/js/customizing-error-reports/#the-report-object ) that does not include the original exception objects that are associated w/ the report. Having the original error instance included in the report would allow clients to leverage more sophisticated logic when determining whether or not to send the report along to the server. This would enable ember applications to leverage things like [`isAjaxError`](https://github.com/ember-cli/ember-ajax/blob/101c357645cff3ac7f5a6b0d4315275ddc882e62/addon/errors.ts#L82-L84).

Related to #43.

bengourley:
This PR contains a few small improvements for the following situations:

- `releaseStage` is not in the list of `notifyReleaseStages`, therefore reports are not send
- a `beforeSend` callback prevents a report from being sent

### The problem

As described in #449, Bugsnag shouldn't silently consume the errors if it's not going to send them anywhere, they should be logged out.

### Unhandled errors

In both of these cases, when the error is __unhandled__ the following log message would be written and the `notify(err, opts, cb)` callback would not run:

```
[bugsnag] Report not sent due to releaseStage/notifyReleaseStages configuration
```

By ensuring that the `notify(err, opts, cb)` callback is called, in Node this means the defautlt `onUncaughtException` (and `onUnhandledRejection`) now correctly get called so the error is printed and the process exits (using the example from #449):

```
[bugsnag] Report not sent due to releaseStage/notifyReleaseStages configuration
[bugsnag] Uncaught exception, the process will now terminate…
/Users/bengourley/scratch/bad.js:2
  bad: 123;
          ^

SyntaxError: Unexpected token ;
    at new Script (vm.js:74:7)
    at createScript (vm.js:246:10)
    at Object.runInThisContext (vm.js:298:10)
    at Module._compile (internal/modules/cjs/loader.js:670:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
```

Unit tests were added to ensure the behaviour was correct for various permutations.

### Handled errors

The situation for handled errors is a little different. We can't assume that a log is desirable, but we can provide the means for users to log it themselves.

I added a property `originalError` to the `Report` class which contains whatever the raw thing was that the report was created from. This also resolves #420, meaning that users who want to do something such as 

- log out the stack
- inspect the error

can do so with the following:

```js
// it's available in beforeSend, however if the beforeSend callback was
// added after one which can return false, this one won't be run
bugsnagClient.notify(new Error, {
  beforeSend: report => console.log(report.originalError)
})

// so it's also in the notify callback which runs after the last
// beforeSend, even if one returned false
bugsnagClient.notify(new Error, { /* opts */ }, (err, report) => {
  console.log(report.originalError)
})
```

## Tests

Added unit tests to ensure the `client.notify(err, opts, cb)` callback was always called. The end-to-end tests aren't set up to make assertions about the the process's output, so I manually tested that the expected content was seen when `releaseStage` was not in `notifyReleaseStages` and when a `beforeSend` prevented a report from sending.

## Bundle size discussion time

Since this makes changes to `@bugsnag/core` it stands to affect browser bundle size. Here's the change:

```diff
- 11.68kB
+ 11.69kB
```

Net increase of `0.01kB`. I think this is 🆗.

### DIFF
--- a/packages/core/types/report.d.ts
+++ b/packages/core/types/report.d.ts
@@ -33,7 +33,14 @@ declare class Report {
     url: string;
   };
 
-  constructor(errorClass: string, errorMessage: string, stacktrace?: any[], handledState?: IHandledState);
+  constructor(
+    errorClass: string,
+    errorMessage: string,
+    stacktrace?: any[],
+    handledState?: IHandledState,
+    originalError?: any,
+  );
+
   public isIgnored(): boolean;
   public ignore(): void;
   public updateMetaData(section: string, value: object): Report;
--- a/packages/core/types/report.d.ts
+++ b/packages/core/types/report.d.ts
@@ -32,6 +32,7 @@ declare class Report {
   public request: {
     url: string;
   };
+  public originalError: any;
 
   constructor(
     errorClass: string,
--- a/packages/browser/types/test/types.test.js
+++ b/packages/browser/types/test/types.test.js
@@ -51,9 +51,9 @@ bugsnag({
     const program = `
 import { Bugsnag } from "../../..";
 let bugsnagInstance: Bugsnag.Client | undefined = undefined;
-export function notify(error: Bugsnag.NotifiableError, opts?: Bugsnag.INotifyOpts): boolean {
+export function notify(error: Bugsnag.NotifiableError, opts?: Bugsnag.INotifyOpts): void {
   if (bugsnagInstance === undefined) {
-    return false
+    return
   }
   return bugsnagInstance.notify(error, opts)
 }
--- a/packages/browser/types/test/types.test.js
+++ b/packages/browser/types/test/types.test.js
@@ -106,4 +106,22 @@ console.log(bugsnagClient.getPlugin('foo') === 10)
     ])
     expect(stdout.toString()).toBe('')
   })
+
+  it('should work with the notify() callback', () => {
+    const program = `
+import bugsnag from "../../..";
+const bugsnagClient = bugsnag('api_key');
+bugsnagClient.notify(new Error('123'), {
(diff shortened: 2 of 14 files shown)


packages/core/types/report.d.ts
```diff
@@ -33,7 +33,14 @@ declare class Report {
     url: string;
   };
 
-  constructor(errorClass: string, errorMessage: string, stacktrace?: any[], handledState?: IHandledState);
+  constructor(
+    errorClass: string,
+    errorMessage: string,
+    stacktrace?: any[],
+    handledState?: IHandledState,
+    originalError?: any,
```
kattrali: Is there some docs or usage/tests to go along with this change?

bengourley: I figured I would write the docs once the implementation was ratified. I don't have a good reason for not adding for that aspect of this PR though so I will add some.

jordyvandomselaar:
Is there any chance this could be added to the JS docs? I couldn't find it in there =), thanks for the originalError addition! It has helped us a lot with grouping errors etc.

PR #516, opened 2019-04-04
Improve normaliseError() utility in Client class

bengourley:
This PR makes a couple of improvements to the `normaliseError()` function which is used by `notify()` to coerce its arguments into a reasonable error, or generate an error alerting the developer to incorrect usage:

- When https://github.com/bugsnag/bugsnag-js/commit/dd4030b505a5e5f875871de2d05f9537ec00a0af was added it meant every `notify()` call that ended up generating a stacktrace included a frame from this codebase. The `errorFramesToSkip` count has been bump to mitigate that.
- The logic for a missing error argument has been moved so that it is all encapsulated inside this function, and in the process fixed an issue where `null` ended up in the "object" case block (see #400 for original report/potential fix)

## Testing

Unit tests and end-to-end tests added to cover these changes.

## Bundle size discussion time

Since this makes changes to `@bugsnag/core` it stands to affect browser bundle size. Here's the change:

```diff
- 11.68kB
+ 11.64kB
```

Net decrease of `0.04kB`. I think this is 🆒.

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -157,13 +157,6 @@ class BugsnagClient {
     let { err, errorFramesToSkip, _opts } = normaliseError(error, opts, this._logger)
     if (_opts) opts = _opts
 
-    // if we have something falsey at this point, report usage error
-    if (!err) {
-      const msg = generateNotifyUsageMessage('nothing')
-      this._logger.warn(`${LOG_USAGE_ERR_PREFIX} ${msg}`)
-      err = new Error(`${REPORT_USAGE_ERR_PREFIX} ${msg}`)
-    }
-
     // ensure opts is an object
     if (typeof opts !== 'object' || opts === null) opts = {}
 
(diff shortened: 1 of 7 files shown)


test/browser/features/handled_errors.feature
```diff
@@ -51,3 +51,35 @@ Scenario Outline: calling notify() with Error within Promise catch
     | browserify |
     | rollup     |
     | typescript |
+
+Scenario Outline: calling notify() with an object, getting a generated a stacktrace
```
snmaynard: Theres no need for this to be a scenario outline as it only has one example

packages/core/client.js
```diff
@@ -243,37 +242,33 @@ const normaliseError = (error, opts, logger) => {
...
         // report usage/deprecation errors if this function is called like that
-        const msg = generateNotifyUsageMessage('string/string')
-        logger.warn(`${LOG_USAGE_ERR_PREFIX} ${msg}`)
-        err = new Error(`${REPORT_USAGE_ERR_PREFIX} ${msg}`)
+        err = createAndLogUsageError('string/string')
         _opts = { metaData: { notifier: { notifyArgs: [ error, opts ] } } }
       } else {
         err = new Error(String(error))
-        errorFramesToSkip += 2
+        errorFramesToSkip += 3
```
snmaynard: Should this be a const because the same value is also used below

PR #518, opened 2019-04-08
Add notify's 3rd argument type

deanmraz:
# Issue 
- Typescript is unaware of 3rd argument 

# Goal 
- Add 3rd argument to typescript definition

# Code 
- https://github.com/bugsnag/bugsnag-js/blob/next/packages/core/client.js#L150

```
notify (error, opts = {}, cb = () => {}) {
```


bengourley:
This is fixed in #515 which is due to land soon. Thanks!

PR #525, opened 2019-04-23
v6.2.0 (@bugsnag/expo release)

bengourley:
This release adds the Expo notifier (`@busgnag/expo`) and all of its associated packages (`@bugsnag/delivery-expo`, `@bugsnag/plugin-(react-native|expo)-*`)

All of the functionality has been PR'd and QA'd before landing on the Expo branch itself. This PR needs just needs review on the following updates:

- changelog
- readme

After that a release can be cut, and the publish will be coordinated with the docs.


README.md
```diff
@@ -2,19 +2,19 @@
...
 
-This is a monorepo (managed with [Lerna](https://lernajs.io/)) containing our universal error reporting client: [`@bugsnag/js`](/packages/js), along with:
+This is a monorepo (managed with [Lerna](https://lernajs.io/)) containing our universal error reporting client [`@bugsnag/js`](/packages/js) and our Expo client [`@bugsnag/expo`](/packages/expo), along with:
 
 - the core Bugsnag libraries for reporting errors ([`@bugsnag/core`](/packages/core))
 - plugins for supporting various frameworks (e.g. [`@bugsnag/plugin-react`](/packages/plugin-react))
 - plugins for internal functionality (e.g. [`@bugsnag/plugin-simple-throttle`](/packages/plugin-simple-throttle))
 
-See [packages](/packages) for a full list of contents.
+Etc. See [packages](/packages) for a full list of contents.
```
fractalwrench: "Etc." can be removed here

README.md
```diff
@@ -2,19 +2,19 @@
...
 
 ## Getting started
```
fractalwrench: The 'Integrating with Frameworks' section is missing an entry for the Expo plugin/docs

README.md
bengourley: I wanted to imply that there were many more than just the three mentioned above. Does

> See packages for a full list of contents.

Imply that strongly enough for you?

README.md
bengourley: There's nothing to install so the entire table would be something like:

### Expo

| Framework  | Bugsnag plugin | Documentation |
| ---------- | -------------- | --------------|
| React | [@bugsnag/plugin-react](packages/plugin-react) (already bundled with the Expo notifier) | [Capturing react render errors docs](https://deploy-preview-906--docs-bugsnag.netlify.com/platforms/react-native/expo/#capturing-react-render-errors)

It doesn't really fit this format which is why I left it off. I could add an Expo section that explains this in copy instead?

README.md
fractalwrench: I think that works fine

PR #531, opened 2019-04-26
fix(plugin-node-surrounding-code): Ensure surrounding code is truncated to sensible length

bengourley:
For minified code a single line can be unreasonably long, causing the payload to be rejected because
it's over the 1MB size limit. This change just takes the first 200 characters from the line.


packages/plugin-node-surrounding-code/surrounding-code.js
```diff
@@ -86,7 +86,7 @@ class CodeRange extends Writable {
     this._n++
     if (this._n < this._start) return cb(null)
     if (this._n <= this._end) {
-      this._code[String(this._n)] = chunk
```
snmaynard: Is the substring centered on the column provided if we get one or is it just the start of line?

bengourley: It's just the start of the line. We could center around the column but only for one of the lines.

snmaynard: yeah - might get confusing. ok!

packages/plugin-node-surrounding-code/test/surrounding-code.test.js
```diff
@@ -207,4 +207,32 @@ describe('plugin: node surrounding code', () => {
       }
     ]))
   })
+
+  it('truncates lines to 150 characters', done => {
```
fractalwrench: should this say 200 characters?

bengourley: yes – or better still only have the constant in the assertion and make the scenario description generic

PR #532, opened 2019-05-02
Expo end-to-end tests

Cawllec:
### Goal
Adds e2e tests for Expo running as a standalone app on Android devices in Browserstack in both CI and when running locally.

A second PR adding iOS functionality to this change will be coming soon.


dockerfiles/Dockerfile.ci
```diff
@@ -1,5 +1,5 @@
 # CI test image for unit/lint/type tests
-FROM node:10-alpine
```
snmaynard: We should only use even numbered node releases. We may want to jump to 12 now though.

snmaynard: https://medium.com/@nodejs/october-brings-node-js-10-x-to-lts-and-node-js-11-to-current-ae19f8f12b51

test/expo/features/support/env.rb
```diff
@@ -0,0 +1,130 @@
...
+def click_element(element)
+  unless $bs_driver.nil?
+    $bs_driver.click_element(element)
+  end
+end
+
+def set_dropdown_value(element, value)
+  unless $bs_driver.nil?
+    $bs_driver.click_element(element)
+    sleep(1)
```
snmaynard: do we really need the sleep?

Cawllec: We actually don't need this at all, so I'll remove it.

Cawllec:
@bengourley Everything should be up to date now.  If you could verify this addition is running properly by giving it a once over, and by running the Android tests using the instruction in the contributing guide that'd be great.

dockerfiles/Dockerfile.expo-android-builder
```diff
@@ -0,0 +1,28 @@
...
+
+RUN apt-get update
+RUN apt-get install -y nodejs make g++ yarn
+
+WORKDIR /app/test/expo/features/fixtures/test-app
+
+RUN mkdir -p /app/test/expo/features/fixures/build
+
+RUN npm i -g gulp-cli node-gyp
+RUN yarn add bunyan turtle-cli
```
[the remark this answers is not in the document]
Cawllec: 1. It complains and fails if bunyan isn't explicitly installed for some reason
2. We need to alter part of the turtle-cli installation to remove the push-notification entitlement.  This makes using the local installation a requirement
3. Honestly, no real idea.  If `turtle-cli` is installed with npm it fails to install & build around 60% of the time.  As it uses Yarn internally, it makes more sense to install it with yarn.  Is there a way to use yarn to install global packages?  It seemed they'd deprecated that option

test/expo/features/fixtures/test-app/App.js
```diff
@@ -0,0 +1,113 @@
...
+import Sessions from './app/sessions'
+import NetworkBreadcrumbs from './app/network_breadcrumbs'
+
+export default class App extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      scenario: null,
+      scenarios: [
+        '',
```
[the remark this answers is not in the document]
Cawllec: Yes, but probably not necessary, I'll look at refactoring

dockerfiles/Dockerfile.expo-android-builder
```diff
@@ -0,0 +1,28 @@
...
+
+RUN node_modules/.bin/turtle setup:android
```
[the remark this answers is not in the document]
Cawllec: See my about comment

dockerfiles/Dockerfile.expo-android-builder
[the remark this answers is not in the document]
bengourley: Re 3: looks it was deprecated it as a flag and added it as a command instead:
https://yarnpkg.com/lang/en/docs/cli/global/

So you could try installing everything with yarn:

```
yarn global add turtle-cli gulp-cli node-gyp
```

And then add bunyan if it still doesn't work – but it seems like it should.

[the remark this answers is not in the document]
bengourley: Re 2: where does this alteration take place?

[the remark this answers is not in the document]
Cawllec: It's actually still in the iOS branch as that's what it's required for.  Should have made that clear!

Cawllec:
I've messaged BrowserStack re `reset-device` timings on Android 7 & 8, but aside from that I'm not sure there's much more I can add/remove to improve speed atm.  I'd definitely appreciate a second review.

.buildkite/pipeline.yml
```diff
@@ -51,8 +51,44 @@ steps:
...
+          cache-from:
+            - expo-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:ci-${BUILDKITE_BRANCH}
+      - docker-compose#v2.6.0:
+          push:
+            - expo-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:ci-${BUILDKITE_BRANCH}
+            - expo-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:ci-latest
+
   - wait
 
+  - label:  ':docker: Publish expo app'
```
snmaynard: This should have its own build step that caches properly as it takes 15 minutes! Currently you are caching from ci but it doesnt share any layers from ci because the first layer has differences - one is `RUN apk add --update bash git` and the other is `RUN apk add --update bash`. We should try to make the first few steps in expo-publisher the same as ci to make that cache reusable and only fall back to that if we dont have a pre-existing cache for the expo-publisher image on this `BUILDKITE_BRANCH`

.buildkite/pipeline.yml
```diff
@@ -51,8 +51,44 @@ steps:
...
+
+  - label:  ':docker: Build expo APK builder'
+    env:
+      EXPO_RELEASE_CHANNEL: ${BUILDKITE_BUILD_ID}
+    artifact_paths: build/output.apk
+    plugins:
+      - docker-compose#v2.6.0:
+          build: expo-android-builder
+          cache-from:
+            - expo-android-builder:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:expo-android-builder-latest
```
snmaynard: This should have the branch name as well as a latest cache so it doesnt matter if we have multiple branches going on at the same time.

.buildkite/pipeline.yml
Cawllec: I've split it between a build step that runs in the same stage as building the maze images (that should run in approximately the same time as the node-maze image), and a run step in the same place as previous.  I'll run it a couple of times to check the caching, and see if there's a speed increase there.

bengourley:
Having looked at the length of the CI jobs on this branch vs. those on next/master, the average time to complete pretty much doubles. It's not ideal, but I don't have anything constructive to suggest.

Cawllec:
Do we think this is mergeable now?  Shall we wait for @bengourley before doing so?

.buildkite/pipeline.yml
```diff
@@ -60,6 +68,7 @@ steps:
         use-aliases: true
     env:
       BROWSER: "chrome_43"
+      VERBOSE: 'true'
```
snmaynard: Remove this

PR #538, opened 2019-05-16
test(node): add MazeRunner tests for node runtime version tracking

tomlongridge:


test/node/features/runtime_versions.feature
```diff
@@ -0,0 +1,29 @@
...
+  Given I store the api key in the environment variable "BUGSNAG_API_KEY"
+  And I store the endpoint in the environment variable "BUGSNAG_NOTIFY_ENDPOINT"
+  And I store the endpoint in the environment variable "BUGSNAG_SESSIONS_ENDPOINT"
+
+Scenario: report for unhandled event contains runtime version information
+  And I run the service "unhandled" with the command "node scenarios/thrown-error-not-caught"
+  And I wait to receive a request
+  Then the request is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
+  And the event "unhandled" is true
+  And the event "device.runtimeVersions.node" is not null
```
snmaynard: Is it possible to have a better check for this? I assume we could assert the type is string, but could we have a version expection or a regex that it should look like?

bengourley: Yeah or even further than that – does the test process know what major version of Node is being used?

If so it could check that the version starts with that, e.g. something like

`And the event "device.runtimeVersions.node" starts with env.NODE_VERSION`

Cawllec: A useful addition may be being able to define a Hash of variables to test against in the `env.rb`, and accompanying steps.  Something like:

```
#env.rb
$test_variables = {
  :notifier_name => 'Bugsnag Node',
  :notifier_version => 'x.x.x',
  :node_version => ENV['NODE_VERSION']
}
```

```
#Steps
Then('the event {string} equals the test variable {string}') do
...
end
```

Obviously that's not something that can be done right now, so using a regex is probably the best current option, something like:

`And the event "device.runtimeVersions.node" matches "(\d+\.){2}\d+"`

tomlongridge: Added the regex tests; @Cawllec is adding the above MazeRunner enhancement to Trello.

snmaynard: I think adding something that can check against an environment variable might be useful. Having it check the specific version would be useful

PR #559, opened 2019-06-20
Add a check for the case when there's no stacktrace at all

andvla:
Getting this error message when there's no stack-trace. This PR fixes that issue by first checking if `frame` exists.

Screenshot 2019-06-20 at 10 59 45


bengourley:
Hey @andvla, thanks for the fix. I'll land this with a test and get it published. Cheers!

PR #560, opened 2019-06-24
Add React Native example app

fractalwrench:
Adds a React Native example app to the examples directory. This is mostly a verbatim copy of the existing project, created from scratch to use the latest dependencies/build tools etc.


fractalwrench:
I've removed the auto-generated tests and added `.npmrc` to `.gitignore` to avoid it being committed to VCS.

The `gradle-wrapper.jar` needs to be included in source control to make local development easier - we could remove it from the final artefact however.

PR #561, opened 2019-06-24
Flesh out JS->Native comms design

bengourley:
This spec should contain everything in the `BugsnagReactNative` class that the JS layer needs.

Please review the doc in its entirety rather than the changeset as the commit before was just a wip.

### DIFF
--- a/packages/react-native/native-js-interface.md
+++ b/packages/react-native/native-js-interface.md
@@ -4,7 +4,7 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
 
 ## Intro
 
-`BugsnagReactNative` is a class which will be instantiated once for the entire runtime and accessible via JS. It's responsible for managing the instantiation of the native client (either `bugsnag-android` or `bugsnag-cocoa`) and providing forwarding on data and invocations from JS to the native client.
+`BugsnagReactNative` is a class which will be instantiated once for the entire runtime and accessible via JS. It's responsible for managing the communicating with the native client (either `bugsnag-android` or `bugsnag-cocoa`) and providing data and invocations from JS to the native client.
 
 ## Accessing `BugsnagReactNative` from JavaScript
 
@@ -12,19 +12,135 @@ Access is set up in the standard React Native way<sup>[1](https://facebook.githu
 
 ```js
 const { NativeModules } = require('react-native')
-const NativeClient = NativeModules.BugsnagReactNative
+const { BugsnagReactNative } = NativeModules
 ```
 
 ## API
 
-### `BugsnagReactNative.leaveBreadcrumb(breadcrumb): void`
+### Methods
+
+#### `BugsnagReactNative.getConfig(): configMap`
+
+Retrieves a data structure representing the configuration which the native client was initialised with. The data structure needn't be immutable since it will be passed over the bridge anyway, but mutations should not have any effect.
+
+An "empty" map will indicate to the JS layer that the native client is not configured, which is a problem.
+
+This method is required to be synchronous so it should be annotated as such. This is because during its synchronous configuration, the JS layer needs to obtain config from the native layer.
+
+###### Android
+
+```java
+@ReactMethod(isBlockingSynchronousMethod = true)
+public ReadableMap getConfig()
+```
+
+###### iOS
+```objc
+(NSDictionary *)getConfig;
+
+// annotation goes in BugsnagReactNative.m
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getConfig);
+```
+
+#### `BugsnagReactNative.updateClientState(key, updates): void`
+
+Updates to mutable client state in JS (e.g. `metaData`, `user`) should be replicated and stored in the native layer using this method.
+
+###### Android
+```java
+void updateClientState(String key, ReadableMap update)
+```
+
+###### iOS
+```objc
+- (void)updateClientState:(NSString * key)
+                         :newValue (NSDictionary *)update;
+```
+
+The keys in the updates map are the names of the properties to be updated and the values are encoded objects containing the new value for that property. The list of allowed keys is:
+
+- `user`: a Map<String, String> with `id`, `name` and `email`
+- `metaData`: a Map<String, Map> of metadata tabs
+
+#### `BugsnagReactNative.deliver(payload): Promise<boolean>`
+
+Send a report that was created in JS-land. The boolean return value denotes the following:
+
+- `true`: the report was sent, or was enqueued with the intention of sending later
+- `false`: the report was not sent and will never be sent
+
+The `report` contains this subset of the error reporting payload structure:
+
+| Key                        | Type                  |
+|----------------------------|-----------------------|
+| `exceptions`               | `Array[Exception]`    |
+| `severity`                 | `String`              |
+| `unhandled`                | `boolean`             |
+| `app`                      | `Map<String, String>` |
+| `device`                   | `Map<String, String>` |
+| `breadcrumbs`              | `Array<Breadcrumb>`   |
+| `context`                  | `String`              |
+| `user`                     | `Map`                 |
+| `metaData`                 | `Map<String, Map>`    |
+| `groupingHash`             | `String`              |
+
+The native layer will be responsible for tracking the handled/unhandled count based on the handled state in the provided payload, and for using this payload as a seed for constructing the actual payload it will send.
+
+###### Android
+```java
+public void deliver(ReadableMap payload, Promise promise)
+```
+
+###### iOS
+```objc
+- (void)deliver:(NSDictionary *)payload
+        resolve:(RCTPromiseResolveBlock)resolve
+         reject:(RCTPromiseRejectBlock)reject;
+```
+
+#### `BugsnagReactNative.getPayloadInfo(): Promise<info>`
+
+Retrieves info from the native layer so that JS `beforeSend` callbacks can access it.
+
+###### Android
+```java
+public void getPayloadInfo(Promise promise)
+```
+
+###### iOS
+```objc
+- (void)getPayloadInfo:(RCTPromiseResolveBlock)resolve
+                reject:(RCTPromiseRejectBlock)reject;
+```
+
+`info` should be a `Map<String, any>` containing the following keys:
+
+| Keys          |
+|---------------|
+| `threads`     |
+| `breadcrumbs` |
+| `app`         |
+| `device`      |
+| `metaData`    |
+| `user`        |
+| `context`     |
+
+#### BugsnagReactNative.leaveBreadcrumb(breadcrumb): void`
 
 Breadcrumbs collected from JS should be sent to the native layer to be stored.
 
-- Android: `void leaveBreadcrumb(ReadableMap breadcrumb)`
-- iOS `- (void)leaveBreadcrumb:(NSDictionary *)breadcrumb;`
+###### Android
 
-The `breadcrumb` map/dictionary must have the following structure in order to be valid<sup>1</sup>:
+```java
+void leaveBreadcrumb(ReadableMap breadcrumb)
+```
+
+###### iOS
+```objc
+- (void)leaveBreadcrumb:(NSDictionary *)breadcrumb;
+```
+
+The `breadcrumb` map/dictionary must have the following structure in order to be valid:
 
 | Key        | Type                | Required  |
 |------------|---------------------|-----------|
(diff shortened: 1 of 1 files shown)


packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
-### `BugsnagReactNative.leaveBreadcrumb(breadcrumb): void`
+### Methods
+
+#### `BugsnagReactNative.getConfig(): configMap`
+
+Retrieves a data structure representing the configuration which the native client was initialised with. The data structure needn't be immutable since it will be passed over the bridge anyway, but mutations should not have any effect.
+
+An "empty" map will indicate to the JS layer that the native client is not configured, which is a problem.
+
+This method is required to be synchronous so it should be annotated as such. This is because during its synchronous configuration, the JS layer needs to obtain config from the native layer.
```
bengourley: I'm concerned about the ramifications of making this method call synchronous. The only reason we need it to be synchronous is that the interface of the JS notifier is synchronous. It would be a big change and departure from the current interface to make it async, but is that a good enough reason?

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+
+###### iOS
+```objc
+(NSDictionary *)getConfig;
+
+// annotation goes in BugsnagReactNative.m
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getConfig);
+```
+
+#### `BugsnagReactNative.updateClientState(key, updates): void`
```
bengourley: Since we boiled down clientState to just two updateable properties (that can go over the RN bridge), would it instead make sense to have to separate methods?

- `updateMetaData(key, val)`
- `updateUser(id, name, email)`

The reason I kept it like this is to future-proof – new updateable properties result in changes to implementation but not interface.

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+| `unhandled`                | `boolean`             |
+| `app`                      | `Map<String, String>` |
+| `device`                   | `Map<String, String>` |
+| `breadcrumbs`              | `Array<Breadcrumb>`   |
+| `context`                  | `String`              |
+| `user`                     | `Map`                 |
+| `metaData`                 | `Map<String, Map>`    |
+| `groupingHash`             | `String`              |
+
+The native layer will be responsible for tracking the handled/unhandled count based on the handled state in the provided payload, and for using this payload as a seed for constructing the actual payload it will send.
```
bengourley: The caveat of this is that in a JS beforeSend callback, we _don't_ have a correct handled/unhandled count in the session, but it means the data is maintained in a single location.

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+
+| Keys          |
+|---------------|
+| `threads`     |
+| `breadcrumbs` |
+| `app`         |
+| `device`      |
+| `metaData`    |
+| `user`        |
+| `context`     |
```
bengourley: @kattrali's prototype had `session` in here but for it to be correct, we would have to be manually augment the handled/unhandled count, and then the native layer would have to throw this away anyway.

packages/react-native/native-js-interface.md
```diff
@@ -33,85 +149,105 @@ The `breadcrumb` map/dictionary must have the following structure in order to be
...
 
-Updates to mutable client state in JS (e.g. `metaData`, `user`) should be replicated and stored in the native layer.
+#### `BugsnagReactNative.stopSession(): void`
 
-- Android: `void updateClientState(ReadableMap updates)`
-- iOS `- (void)updateClientState:(NSDictionary *)updates;`
+Calls straight through to the native `stopSession()` method.
 
-The keys in the updates map are the names of the properties to be updated and the values are encoded objects containing the new value for that property. The list of allowed keys is:
+#### `BugsnagReactNative.resumeSession(): Promise<boolean>`
```
bengourley: This one's gotta be async if we want a return value, but thankfully we didn't create this method in JS yet.

bengourley:
One thing missing from this interface is the events that communicate back from native -> JS. TODO!

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+###### iOS
+```objc
+- (void)deliver:(NSDictionary *)payload
+        resolve:(RCTPromiseResolveBlock)resolve
+         reject:(RCTPromiseRejectBlock)reject;
+```
+
+#### `BugsnagReactNative.getPayloadInfo(): Promise<info>`
+
+Retrieves info from the native layer so that JS `beforeSend` callbacks can access it.
```
fractalwrench: Will this native info be combined with the JS info before being exposed in a callback?

packages/react-native/native-js-interface.md
fractalwrench: I think it makes sense to omit session as we don't want users to alter it directly anyway.

packages/react-native/native-js-interface.md
```diff
@@ -33,85 +149,105 @@ The `breadcrumb` map/dictionary must have the following structure in order to be
...
-}
+### Android
 
-MAP = {
-  "type": "map",
-  "value": {
-    [key: string]: VALUE
-  }
-}
+RN's `ReadableMap` type in Java provides the means to get the type of each entry `ReadableType getType (String key)`:
```
fractalwrench: This should work very nicely compared to our current solution 👌

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
 ## API
 
-### `BugsnagReactNative.leaveBreadcrumb(breadcrumb): void`
+### Methods
+
+#### `BugsnagReactNative.getConfig(): configMap`
+
+Retrieves a data structure representing the configuration which the native client was initialised with. The data structure needn't be immutable since it will be passed over the bridge anyway, but mutations should not have any effect.
+
+An "empty" map will indicate to the JS layer that the native client is not configured, which is a problem.
```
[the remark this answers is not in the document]
bengourley: `BugsnagReactNative` will always be initialised by the system as a singleton. So the implementation of this method will know if the native client has been been configured or not.

Thinking about it, I'm not sure why I thought an empty map would be the best thing to return. What about `null` instead?

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+###### iOS
+```objc
+- (void)updateClientState:(NSString * key)
+                         :newValue (NSDictionary *)update;
+```
+
+The keys in the updates map are the names of the properties to be updated and the values are encoded objects containing the new value for that property. The list of allowed keys is:
+
+- `user`: a Map<String, String> with `id`, `name` and `email`
+- `metaData`: a Map<String, Map> of metadata tabs
```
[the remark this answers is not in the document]
bengourley: If I understand what you're getting at… yes, this kind of payload is valid:

```js
"metaData": {
  "tabOne": { "someData": { "nested": { "quite": "deeply" } } }
}
```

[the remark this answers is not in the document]
bengourley: yeah this was in the back of my mind – the reason I went with the whole object initially was

- sending the entire state of the object avoids synchronisation issues you might get with lots of little updates flying around
- have to create an interface for deleting properties

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+
+#### `BugsnagReactNative.deliver(payload): Promise<boolean>`
```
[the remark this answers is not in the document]
bengourley: There are two things on my mind here:

1. we want to avoid synchronous bridge calls if at all possible (they block the main thread)
2. in handling a fatal JS error, my understanding is that the process remains alive until there is no work left to do – so any async stuff in an `onerror` callback is allowed to complete.

So I think this ok – but we can and should revisit when implementing and testing.

packages/react-native/native-js-interface.md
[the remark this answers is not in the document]
fractalwrench: Cool - I think let's go with sending the whole object first as it's simpler, then we can always investigate the delta approach if we run into perf issues.

packages/react-native/native-js-interface.md
fractalwrench: Seems reasonable to make this synchronous as it's required for init.

packages/react-native/native-js-interface.md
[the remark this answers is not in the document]
fractalwrench: I'm a bit unclear on the scenarios where the native client isn't configured, and how the JS layer would behave in that scenario - I think it'd be good to talk that over. Assuming we can return values, it'd be fine to return null/an empty map.

packages/react-native/native-js-interface.md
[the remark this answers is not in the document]
fractalwrench: It feels reasonable to block the main thread when generating an error report - but maybe I'm coming at this from a mobile rather than JS perspective. For context there were some edge cases around termination in the previous implementation, specifically: https://github.com/bugsnag/bugsnag-react-native/pull/290

packages/react-native/native-js-interface.md
tomlongridge: Agreed - that makes more sense to me, even if it means future new methods if new state data is added.

tomlongridge: Would this also include `context`?

bengourley: Ok happy to go separate methods.

packages/react-native/native-js-interface.md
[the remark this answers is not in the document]
bengourley: 1. consider a better name. suggestion `notify()`
2. figure out if native threads are ever useful for JS exceptions, and if so, do we need them at the exact point the error happened (this will determine whether we need to make `getPayloadInfo()` synchronous.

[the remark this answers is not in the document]
kattrali: > It feels reasonable to block the main thread when generating an error report - but maybe I'm coming at this from a mobile rather than JS perspective.

We should use the new interface in a way that we can get around this requirement entirely - my thoughts for iOS are to bypass the existing delivery mechanism and use `-[BugsnagConfiguration session]` directly to send a request. That way, we can delay resolving the promise until the request completes (successfully or no), and the JS layer can then invoke the default fatal handler.

packages/react-native/native-js-interface.md
```diff
@@ -4,27 +4,143 @@ This is a working spec of this the interface that the native layer (`BugsnagReac
...
+
+The `report` contains this subset of the error reporting payload structure:
+
+| Key                        | Type                  |
+|----------------------------|-----------------------|
+| `exceptions`               | `Array[Exception]`    |
+| `severity`                 | `String`              |
+| `unhandled`                | `boolean`             |
+| `app`                      | `Map<String, String>` |
+| `device`                   | `Map<String, String>` |
```
kattrali: App and device have some non-string properties, like `cpuAbi` on Android (array of strings) or `freeDisk` and friends (number values).

packages/react-native/native-js-interface.md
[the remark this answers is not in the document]
fractalwrench: I'm concerned that bypassing the existing delivery mechanism will lead to edge-cases/different behaviour around how JS reports are cached, and that this will introduce additional complexity to our report (de)serialisation code.

Calling `bugsnag.notify()` with the JS payload coerced into an exception/error would give consistent behaviour for a variety of things (callbacks, payload validation, delivery caching, etc), and seems less risky. Maybe we could investigate exposing an internal `notify` method that takes a `Report` object as a parameter, rather than an `Exception`?

packages/react-native/native-js-interface.md
```diff
@@ -1,30 +1,172 @@
...
+
+A return value of `null` will indicate to the JS layer that the native client is not configured. The JS notifier can do something sensible with this information, like throw an error or log a warning.
+
+This method is required to be synchronous so it should be annotated as such. This is because during its synchronous configuration, the JS layer needs to obtain config from the native layer.
+
+###### Android
+
+```java
+@ReactMethod(isBlockingSynchronousMethod = true)
+public ReadableMap getConfig()
```
fractalwrench: It looks like this will need to be a WriteableMap (as discussed via chat)

PR #562, opened 2019-06-24
Setup build tools for Android React Native

fractalwrench:
## Goal

This changeset sets up Gradle to give quick feedback on whether local changes to the Android codebase compile and pass style checks.

## Changeset

- Setup an instrumentation test target with one example test
- Incremented dependencies to latest major versions
- Setup Android Lint to abort the build on unaddressed errors
- Setup Checkstyle for static analysis that ensures changes to Java files meet our style guidelines
- Updated existing Java files to pass style checks


fractalwrench:
Yep - the first time round it will bundle the JS and compile JVM source files to Dex, then run static analysis. Gradle task avoidance means that a lot of the compilation/static analysis can be skipped if the input of each task hasn't changed.

PR #563, opened 2019-06-24
Prevent error when running inline script plugin on empty stacktraces

bengourley:
In the case of an error without a stacktrace, the `beforeSend` implemented by `@bugsnag/plugin-inline-script-content` would throw an error (see #559).

Since the logic that runs the list of `beforeSend` callbacks tolerates errors, the effect of this was pretty minimal – just some undesirable error logs output.

The fix supplied by @andvla (thanks 🙏) was tweaked so that we won't attempt to look for `.lineNumber` on any falsey value for `stacktrace[0]`.

### DIFF
--- a/packages/plugin-inline-script-content/inline-script-content.js
+++ b/packages/plugin-inline-script-content/inline-script-content.js
@@ -74,7 +74,7 @@ module.exports = {
       }
 
       // only attempt to grab some surrounding code if we have a line number
-      if (frame === undefined || frame.lineNumber === undefined) return
+      if (!frame || frame.lineNumber === undefined) return
       frame.code = addSurroundingCode(frame.lineNumber)
     })
 
--- a/packages/plugin-inline-script-content/inline-script-content.js
+++ b/packages/plugin-inline-script-content/inline-script-content.js
@@ -74,7 +74,7 @@ module.exports = {
       }
 
       // only attempt to grab some surrounding code if we have a line number
-      if (!frame || frame.lineNumber === undefined) return
+      if (!frame || !frame.lineNumber) return
       frame.code = addSurroundingCode(frame.lineNumber)
     })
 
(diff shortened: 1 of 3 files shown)


packages/plugin-inline-script-content/inline-script-content.js
```diff
@@ -74,7 +74,7 @@ module.exports = {
       }
 
       // only attempt to grab some surrounding code if we have a line number
-      if (frame.lineNumber === undefined) return
+      if (!frame || frame.lineNumber === undefined) return
```
andvla: oh, I did `if (frame === undefined || frame.lineNumber === undefined) return` for consistency but in case you are checking for `!frame`, might change the second condition to do the same? so...:
```suggestion
      if (!frame || !frame.lineNumber) return
```

bengourley: I left it as `=== undefined` because `0` is a valid number which evaluates as falsey. However, `0` isn't really a valid value for line number so arguably this is unnecessary.

andvla:
@bengourley, could you please release that fix within the next version? Would like to start using it. Thanks 😉

bengourley:
Sure, we've pencilled in a release at some point this week.

Note that this isn't causing any issues or lost reports – the `beforeSend` callback runner tolerates the error so its just a log message, albeit undesirable!

andvla:
@bengourley After updating to 6.3.1 keep on getting this error
Screenshot 2019-06-26 at 14 55 57

And this is coming from `@bugsnag/browser/dist/bugsnag.js` line 2152, not sure if there's a way for me to fix it.

andvla:
I guess bugsnag/browser needs another release as well

bengourley:
@andvla this is now released in [v6.3.2](https://github.com/bugsnag/bugsnag-js/releases/tag/v6.3.2)

andvla:
@bengourley perfect, thanks!

byzyk:
Hi @bengourley! Was wondering why `6.3.2` is not available on NPM and if this is going to be published at some point?

Thanks for the fix btw 🎉

bengourley:
It should be available on npm: https://www.npmjs.com/package/@bugsnag/js/v/6.3.2

byzyk:
Aw sorry my bad was looking at react plugin 😃

PR #566, opened 2019-06-26
Expo ci fix turtle

Cawllec:
Fixes the CI failures with the cached version of Turtle (0.8.5) and fixes the version to avoid issues in the future.

### DIFF
--- a/dockerfiles/Dockerfile.expo-publisher
+++ b/dockerfiles/Dockerfile.expo-publisher
@@ -1,6 +1,6 @@
 FROM node:12-alpine
 
-RUN apk add --update bash git
+RUN apk add --update bash git python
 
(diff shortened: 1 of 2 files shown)


dockerfiles/Dockerfile.expo-publisher
```diff
@@ -1,6 +1,6 @@
 FROM node:12-alpine
 
-RUN apk add --update bash git
+RUN apk add --update bash git python
```
bengourley: we need python now?

Cawllec: Expo requires it in some modules for publishing.  It seems to find alternatives if it's not available, but they're more flakey than just having it included.

PR #572, opened 2019-07-01
Add deserialisation code for Breadcrumb in React Native Android

fractalwrench:
This changeset implements a deserialiser for a breadcrumb message received in the Android Native Module.

The `ReadableMapDeserializer` interface has been changed to use a `Map` parameter rather than a `ReadableMap` as this is easier to test and also allows for us to pass a greater range of `Map` implementations.

Note: as Breadcrumb's constructor is package-private, we will need to update bugsnag-android to make this public before releasing.


fractalwrench:
We can alter the implementation to accept arbitrary metadata objects if desired pretty easily - I think it was `Map<String, String>` in the spec.

PR #573, opened 2019-07-01
Write configuration into writablemap for RN Android layer

fractalwrench:
Serialises fields from the `Configuration` class into a `WritableMap` which will be accessible from JS.

This also adds a `WritableMapSerializer` interface that can be implemented by any other type wishing to be serialised.

### DIFF
--- a/packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
+++ b/packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
@@ -1,0 +1,27 @@
+package com.bugsnag.reactnative;
+
+import com.bugsnag.android.Configuration;
+
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.bridge.WritableNativeMap;
+
+public class ConfigSerializer implements WritableMapSerializer<Configuration> {
+
+    @Override
+    public WritableMap serialize(Configuration config) {
+        WritableMap map = new WritableNativeMap();
+        WritableMap endpoints = new WritableNativeMap();
+        endpoints.putString("notify", config.getEndpoint());
+        endpoints.putString("sessions", config.getSessionEndpoint());
+        map.putMap("endpoints", endpoints);
+        map.putString("apiKey", config.getApiKey());
+        map.putString("appVersion", config.getAppVersion());
+        map.putString("buildUuid", config.getBuildUUID());
+        map.putString("releaseStage", config.getReleaseStage());
+        map.putBoolean("sendThreads", config.getSendThreads());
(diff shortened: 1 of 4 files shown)


packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class ConfigSerializer implements WritableMapSerializer<Configuration> {
+
+    @Override
+    public WritableMap serialize(Configuration config) {
+        WritableMap map = new WritableNativeMap();
+        map.putString("apiKey", config.getApiKey());
+        map.putString("appVersion", config.getAppVersion());
+        map.putString("buildUuid", config.getBuildUUID());
+        map.putString("endpoint", config.getEndpoint());
```
bengourley: Is it possible to write the two endpoints into a nested `endpoints` map:

```java
WritableMap endpoints = new WritableNativeMap();
endpoints.put("notify", config.getEndpoint());
endpoints.put("sessions", config.getSessionEndpoint());
map.putMap("endpoints", endpoints);
```

And if so should we? 🤔 it matches the JS interface that way, but maybe that's not important.

bengourley:
It works nicely:

```
07-02 16:09:39.501  7616  7681 I ReactNativeJS: { detectNdkCrashes: true,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   detectAnrs: false,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   autoCaptureSessions: true,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   sendThreads: true,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   releaseStage: null,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   sessionEndpoint: 'https://sessions.bugsnag.com',
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   endpoint: 'https://notify.bugsnag.com',
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   buildUuid: null,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   appVersion: null,
07-02 16:09:39.501  7616  7681 I ReactNativeJS:   apiKey: 'YOUR_API_KEY' }
```

packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
fractalwrench: It's easy enough to update the format to match JS - and is probably worth the consistency.

PR #575, opened 2019-07-02
Update metadata in android layer

fractalwrench:
Updates the native metadata when the JS layer posts an update to the metadata. This overwrites the previously existing native metadata.


fractalwrench:
The JNI error was due to an unrelated JNI crash

PR #576, opened 2019-07-02
Begin deserializing error payload from JS in Android

fractalwrench:
Begins deserializing the error payload passed from JS to Android. This changeset handles the primitive fields in the error - separate PRs will be raised for exceptions etc.


fractalwrench:
Yep - arbitrary data can be serialised in deviceData etc.

As the `connectedCheck` issue appears to be affecting all the PRs I'm going to go ahead and merge this one then address it on the react-native branch. Strangely it seems to run the tests fine from Android Studio so I guess there's some classpath issue.

PR #584, opened 2019-07-12
Tolerate stacktrace generation error when in strict mode

### LINKED ISSUE #564: Error 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode
Getting this weird error randomly in my app.

```
'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
```

Screenshot 2019-06-24 20 51 45

bengourley:
When bundled inside a script that's running in strict mode, StackGenerator will throw an error. There's no better way to detect script mode than `try` to do what you need, and `catch` the error so that's what this PR does.

This adds an end-to-end test that confirms an error without a stack is correctly sent from strict mode.

Fixes #564.


bengourley:
Here's the easiest way to test this out…

Firstly, verify that the current version of Bugsnag fails attempting to generate a stacktrace for a stack-less error in strict mode:

1. clone the repo and checkout `master` or `next` (something that doesn't have this fix.
2. install dependencies and build the notifier `npm i && npm run boostrap && npm run build`
3. make a temporary directory `mkdir -p packages/browser/scratch` and insert the following HTML into `packages/browser/scratchstrict.html`

    ```html
    <!DOCTYPE html>
    <html>
      <head>
        <script type="module" src="../dist/bugsnag.min.js"></script>
        <script type="module">window.bugsnagClient = bugsnag('API_KEY')</script>
      </head>
      <body>
        <script type="module">
          console.log(bugsnagClient.BugsnagReport.getStacktrace(123))
        </script>
      </body>
    </html>
    ```
4. start a local webserver for the browser package `npx serve packages/browser`
5. go to `http://localhost:5000/scratch/strict.html` and verify that you see the message `TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them`

Then to verify that this branch fixes the problem:

1. checkout this branch `git checkout tolerate-strict-mode`
2. go to `http://localhost:5000/scratch/strict.html` and verify that you see `[]` logged out and not the error message you saw before

PR #591, opened 2019-07-18
Get React Native configuration options in JS from native layer

bengourley:
- renames `BugsnagReactNative` method `getConfig()` -> `configure()` per spec
- JS calls `configure()` to get native configuration map
- prevents adding `null` values to the configuration map


packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
```diff
@@ -15,9 +15,15 @@ public WritableMap serialize(Configuration config) {
         endpoints.putString("sessions", config.getSessionEndpoint());
         map.putMap("endpoints", endpoints);
         map.putString("apiKey", config.getApiKey());
-        map.putString("appVersion", config.getAppVersion());
+        String appVersion = config.getAppVersion();
+        if (appVersion != null) {
+            map.putString("appVersion", appVersion);
+        }
+        String releaseStage = config.getReleaseStage();
+        if (releaseStage != null) {
```
bengourley: The reason I made this update is because of the [JS validation logic for releaseStage](https://github.com/bugsnag/bugsnag-js/blob/778aa48f2a3f660429098b757004484838c4e093/packages/core/config.js#L58-L62) – if nothing is set then it falls back to `"production"` and explicitly setting it to `null` is not allowed, it has to be omitted _OR_ be a string.

fractalwrench: If `config.releaseStages` or `config.appVersion` is set to `null` in Android the user will have explicitly set the value as `null`. Therefore it seems like they should be set to `null` in JS too. Is this already the case?

bengourley: `null` isn't a valid value for `releaseStage` in JavaScript so this seemed like a sensible thing to do. The JS notifier crashed on startup because it didn't accept `null`. I guess not seeing as it could result in a difference of `releaseStage` in JS (`production`) vs. native (`null`) exceptions.

We should either allow only non-null strings on the native side, or allow `null` on the JS side. I don't know that we'd want to support errors without a `releaseStage` though so I'm in favour of making it strictly a non-null string.

fractalwrench: Going off our API docs I think `null` is a valid value for releaseStage, as the field isn't marked as required. Can the JS notifier be updated to accept `null` as valid?

bengourley: as discussed elsewhere, changes in the near future to `bugsnag-android` mean this value will _always_ be present in the Android's notifier's config and will be a string, so JS can rely it as such.

PR #594, opened 2019-07-22
feat(node): Support loading config from package.json

bengourley:
This supports configuring Bugsnag via `package.json` in Node.

Any primitive values, including maps/arrays of primitive values can be configured under the `bugsnag` key in `package.json`, e.g.:

```json
{
  "name": "MyApp",
  "version": "1.2",
  "bugsnag": {
    "apiKey": "API_KEY",
    "notifyReleaseStages": [ "staging", "production" ],
    "sendCode": false
  }
}
```

It looks first in the current working directory of the running application, then in the directory of the application's entrypoint. __Question:__ should we keep looking further up file paths if we don't find one?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build --scope '@bugsnag/node' --scope '@bugsnag/browser' --scope '@bugsnag/expo' && lerna run build --ignore '@bugsnag/node' --ignore '@bugsnag/browser' --ignore '@bugsnag/expo'",
     "test:lint": "standard",
-    "test:unit": "lerna run test --ignore '@bugsnag/browser' --ignore '@bugsnag/node'",
+    "test:unit": "lerna run test",
     "test:types": "lerna run test:types",
     "test:test-container-registry-login": "$(aws ecr get-login --profile=opensource --no-include-email)",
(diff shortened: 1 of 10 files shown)


snmaynard:
I think we should recurse upwards because node does this. Where does it go from - the application itself or from where the node_module is installed?

bengourley:
The value of `process.mainModule.paths` looks like this:

```
  [
    '/Users/bengourley/Development/bugsnag-js/packages/node/scratch/node_modules',
    '/Users/bengourley/Development/bugsnag-js/packages/node/node_modules',
    '/Users/bengourley/Development/bugsnag-js/packages/node_modules',
    '/Users/bengourley/Development/bugsnag-js/node_modules',
    '/Users/bengourley/Development/node_modules',
    '/Users/bengourley/node_modules',
    '/Users/node_modules',
    '/node_modules'
  ]
```

So we could `path.join(p, '..')` with each of those and just look there?

I don't know then if we should also check `process.cwd()`, maybe not?

bengourley:
I updated the implementation to look adjacent to all of `process.mainModule.paths`, and also upwards from the `process.cwd()` – this is because in the REPL, `process.mainModule` is `undefined` so without this, it wouldn't pick anything up.

PR #597, opened 2019-07-24
Fix representation of simple object errors (currently "[object Object]")

GlennKintscher:
### Description
If a library throws a simple object as an error, it is just displayed as "[object Object]", which is not very helpful for troubleshooting. This fix changes the behavior so that errors are converted to a JSON string if they are not of type Error.    

### Example
| Location | Current behaviour | Expected behaviour |
|-----------|---------------------|----------------------------------|
| Console | image | image
| Online | image | image

### DIFF
--- a/packages/core/lib/report-from-error.js
+++ b/packages/core/lib/report-from-error.js
@@ -12,6 +12,6 @@ module.exports = (maybeError, handledState) => {
     handledState,
     maybeError
   )
-  if (maybeError !== actualError) report.updateMetaData('error', 'non-error value', String(maybeError))
+  if (maybeError !== actualError) report.updateMetaData('error', 'non-error value', JSON.stringify(maybeError))
   return report
 }
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -30,7 +30,7 @@ module.exports = {
   },
   onUncaughtException: {
     defaultValue: () => (err, report, logger) => {
-      logger.error(`Uncaught exception${getContext(report)}, the process will now terminate…\n${(err && err.stack) ? err.stack : err}`)
+      logger.error(`Uncaught exception${getContext(report)}, the process will now terminate…\n${(err && err.stack) ? err.stack : JSON.stringify(err)}`)
       process.exit(1)
     },
     message: 'should be a function',
@@ -38,7 +38,7 @@ module.exports = {
   },
   onUnhandledRejection: {
     defaultValue: () => (err, report, logger) => {
-      logger.error(`Unhandled rejection${getContext(report)}…\n${(err && err.stack) ? err.stack : err}`)
+      logger.error(`Unhandled rejection${getContext(report)}…\n${(err && err.stack) ? err.stack : JSON.stringify(err)}`)
     },
     message: 'should be a function',
     validate: value => typeof value === 'function'


packages/core/lib/report-from-error.js
```diff
@@ -12,6 +12,6 @@ module.exports = (maybeError, handledState) => {
     handledState,
     maybeError
   )
-  if (maybeError !== actualError) report.updateMetaData('error', 'non-error value', String(maybeError))
+  if (maybeError !== actualError) report.updateMetaData('error', 'non-error value', JSON.stringify(maybeError))
```
GlennKintscher: Current behaviour: 
image

packages/core/lib/report-from-error.js
```diff
@@ -12,6 +12,6 @@ module.exports = (maybeError, handledState) => {
     handledState,
     maybeError
   )
-  if (maybeError !== actualError) report.updateMetaData('error', 'non-error value', String(maybeError))
```
GlennKintscher: Current behaviour: 
image

packages/core/lib/report-from-error.js
GlennKintscher: Expected behaviour (in this specific case):
image

packages/node/src/config.js
```diff
@@ -30,15 +30,15 @@ module.exports = {
...
-      logger.error(`Uncaught exception${getContext(report)}, the process will now terminate…\n${(err && err.stack) ? err.stack : err}`)
+      logger.error(`Uncaught exception${getContext(report)}, the process will now terminate…\n${(err && err.stack) ? err.stack : JSON.stringify(err)}`)
       process.exit(1)
     },
     message: 'should be a function',
     validate: value => typeof value === 'function'
   },
   onUnhandledRejection: {
     defaultValue: () => (err, report, logger) => {
-      logger.error(`Unhandled rejection${getContext(report)}…\n${(err && err.stack) ? err.stack : err}`)
```
GlennKintscher: Current behaviour:
image

packages/node/src/config.js
```diff
@@ -30,15 +30,15 @@ module.exports = {
...
+      logger.error(`Unhandled rejection${getContext(report)}…\n${(err && err.stack) ? err.stack : JSON.stringify(err)}`)
```
GlennKintscher: Expected behaviour (in this specific case):
image

mattdyoung:
Hi @GlennKintscher 

Thanks for the PR! This is a good idea which we'd like to take forward, but we have to be careful about trying to JSON stringify things as it can cause errors.

The readme for safe-json-stringify, which we maintain a fork of is quite informative https://github.com/debitoor/safe-json-stringify, so if converting to JSON we should ensure we use our safe-json-stringify module.

Alternatively, there is prior art in the repo for doing this with arbitrary args to for leaving console log arguments as breadcrumbs https://github.com/bugsnag/bugsnag-js/blob/21a1b088ed6d44f1682216ab2c1b5a07cd45fdf8/packages/plugin-console-breadcrumbs/console-breadcrumbs.js#L18-L26 so we could factor that out and use it in both places.

We'll review this and either suggest the necessary changes or more likely factor this in to our own PR.

PR #606, opened 2019-07-31
Testing: Better cache invalidation

Cawllec:
To avoid having to rebuild the `npx lerna bootstrap` step every time a change is made that shouldn't require it, the `package.json` and `package-lock.json` files are now pulled out and packages into an artefact `.tar` file before the build begins.
This process allows the `bootstrap` step to be cached so long as the artefact hash is unchanged (hence why the `.tar` file is created with a set time).

In addition, images with the tag `ci-latest` were not previously using caches when being built.  This gives them slightly different tags, which should allow the builds to be cached as much as possible.


Cawllec:
Note - Currently the pre-set time in the tar packaging process is set to the unix epoch.  This means  the tar will never invalidate a build unless its contents changes.  As an alternative, this could be set to a calculated time, such as the beginning of the week or month that would enable the cache to be rebuilt every once in a while.

PR #610, opened 2019-08-07
Expo SDK 34 support

bengourley:
- Update Expo notifier to use APIs that became available in SDK 33 and which were removed in SDK 34.
- Update `bugsnag-expo-cli` to prompt the user for a version, with the initial value set based on their installed version of Expo.

### DIFF
--- a/packages/expo-cli/commands/install.js
+++ b/packages/expo-cli/commands/install.js
@@ -78,6 +78,7 @@ const selectVersion = async (dir) => {
         : 'If you want the latest version of @bugsnag/expo hit enter, otherwise type the version you want',
       initial: isPre33 ? '6.3.0' : 'latest',
       validate: str => {
+        if (str === 'latest') return true
         if (semver.valid(str)) return true
         if (semver.validRange(str)) return true
         return 'Version must be: a valid semver version/range or "latest"'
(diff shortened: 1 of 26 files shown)


packages/expo-cli/commands/install.js
```diff
@@ -63,6 +65,30 @@ const checkManifest = async (dir) => {
...
+      type: 'text',
+      name: 'version',
+      message: isPre33
+        ? 'It looks like you’re using a version of Expo SDK <33. The latest version of Bugsnag works with SDK >= 33 so it’s recommended that you install the last version of Bugsnag that supported your Expo version: v6.3.0'
+        : 'If you want the latest version of @bugsnag/expo hit enter, otherwise type the version you want',
+      initial: isPre33 ? '6.3.0' : 'latest',
+      validate: str => {
+        if (semver.valid(str)) return true
+        if (semver.validRange(str)) return true
+        return 'Version must be: a valid semver version/range or "latest"'
```
tomlongridge: This doesn't seem to support `latest` as an input - fails with:
```
Version must be: a valid semver version/range or "latest"
```

bengourley: ah yes good catch, need to check `str === 'latest'`  :+1:

PR #626, opened 2019-10-08
fix(plugin-angular): Ensure exported JS from Node notifier matches the shape of the .d.ts export

### LINKED ISSUE #604: Cannot read property 'Client' of undefined
Hey, 
i currently have a Angular 7 Setup with ServerSide Rendering. 
In Browsermode evryting works fine and compiling to production also throws no errors, but
my Problem is when i start the server i get the following error: 

```
   __metadata("design:paramtypes", [_bugsnag_js__WEBPACK_IMPORTED_MODULE_1__["Bugsnag"].Client])
                                                                                         ^TypeError: Cannot read property 'Client' of undefined
    at Module../node_modules/@bugsnag/plugin-angular/dist/esm/index.js 
```

The Angular Setup is as following:

```
import bugsnag from '@bugsnag/js';
import {BugsnagErrorHandler} from '@bugsnag/plugin-angular';
const bugsnagClient = bugsnag(environment.APIKEYS_EXTERNAL.BUGSNAG);
export function errorHandlerFactory() {
  return new BugsnagErrorHandler(bugsnagClient)
}
@NgModule({
....
  providers: [ ...,{ provide: ErrorHandler, useFactory: errorHandlerFactory } ]
...
})
```

Maybe someone can help me
thank you :)

bengourley:
Angular uses this type information at runtime for DI. The browser notifier had it, but the node
notifier was missing it, so this wasn't working when trying to do SSR with Angular.

Updated the comment in the browser notifier to be more clear.

Fixes #604.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "6.4.1"
+  "version": "6.4.2-alpha.0"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "6.4.2-alpha.0"
+  "version": "6.4.2-alpha.1"
 }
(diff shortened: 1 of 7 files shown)


abigail-bugsnag:
LGTM! Tested on my example app which now works.

Issue #631, opened 2019-10-16
bugsnag.ts in npm package


GDreyV:
By some reason npm package v 6.4.2 has extra empty .ts file in @bugsnag\browser\dist\types\bugsnag.ts which causes error on build Angular app:
ERROR in ./node_modules/@bugsnag/browser/dist/types/bugsnag.ts
Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
Error: /root/src/frontend/node_modules/@bugsnag/browser/dist/types/bugsnag.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
The missing file seems to be part of a third party library. TS files in published libraries are often a sign of a badly packaged library. Please open an issue in the library repository to alert its author and ask them to package the library using the Angular Package Format (https://goo.gl/jB3GVv).

bengourley:
This reason this file was added is discussed in https://github.com/bugsnag/bugsnag-js/issues/523.

Maybe Angular's AOT mode is fixed now and we can remove it…

GDreyV:
I'm building it in pretty silly way, just `ng build --eval-source-map --prod`. With `ng build --eval-source-map --prod --aot` I'm getting exactly the same error. Angular version is `"@angular/core": "~7.2.0"`

mattdyoung:
Hi @GDreyV 

Thanks for the PR! Our team will review it and follow up with any comments there.

mattdyoung:
Fixed in [v6.4.3](https://github.com/bugsnag/bugsnag-js/releases/tag/v6.4.3)

Issue #523, opened 2019-04-15
Error When Rebuilding When Building With Ahead of Time Compilation


gregfletch:
When building an Angular application with the ahead of time compilation flag (`--aot`) set (i.e. `ng build --aot --watch` or `ng serve --aot`), the initial build succeeds, but subsequent builds fail with an error like:

```
ERROR in ./src/app/app.module.ngfactory.js
Module not found: Error: Can't resolve '@bugsnag/browser/dist/types/bugsnag' in '/Users/user/dev/ng-app/src/app'
```

Versions:

@bugsnag/core: 6.1.0
@bugsnag/js: 6.1.0
@bugsnag/plugin-angular: 6.1.0
Angular CLI: 7.3.8
Angular: 7.2.13

PACMAN49:
i m facing the same issue with
@bugsnag/core: 6.2.0
@bugsnag/js: 6.2.0
@bugsnag/plugin-angular: 6.2.0
Angular CLI: 7.3.8
Angular: 7.0.2

bengourley:
Thanks for the report. This seems like an issue with Angular's AOT mode. The second time it builds, for some reason, it looks for a `.ts|.tsx|.js|.mjs` file in `node_modules/@bugsnag/browser/dist/types` but there's only `.d.ts` files so is not satisfied. By adding an empty `.ts` file to this directory, the compiler gets tricked into thinking everything is ok and it all worked. This fix is added in #539.

PR #632, opened 2019-10-16
Fix angular build with correct resolving referenced modules

GDreyV:
Instead of creating stub file provide valid browser alias.
Initial issue https://github.com/bugsnag/bugsnag-js/issues/523
Updated issue https://github.com/bugsnag/bugsnag-js/issues/631

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -9,6 +9,9 @@
     "type": "git",
     "url": "git@github.com:bugsnag/bugsnag-js.git"
   },
+  "browser": {
+    "dist/types/bugsnag": "./dist/bugsnag.js"
+  },
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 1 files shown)


tomlongridge:
@GDreyV - are you able to provide a repro case for this? I've tried it with both Bugsnag Angular example (upgraded to 7.2.0) and the Angular basic example (https://angular.io/start) with Bugsnag added and neither result in the error you are seeing.

We need reproduction instructions/project to make this PR something we can release.

GDreyV:
Here's the test repository to reproduce the problem: https://github.com/GDreyV/bugsnag-angular-test
To be sure that we are on the same page I used following steps to reproduce the issue:
```
npm list -g --depth=0
+-- @angular/cli@7.3.9
+-- angular-cli@0.1.0
+-- express-generator@4.14.1
+-- gulp@3.9.1
+-- less@2.7.2
+-- node-sass@4.5.3
+-- npm@6.9.0
+-- typescript@3.3.4000
+-- typings@2.1.1
`-- windows-build-tools@1.2.1

ng new bugsnag-test
cd bugsnag-test
npm install --save @bugsnag/js @bugsnag/core @bugsnag/plugin-angular
```
update app.module.ts with bugsnag init
```
ng build --eval-source-map --prod --aot
ERROR in ./node_modules/@bugsnag/browser/dist/types/bugsnag.ts
Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
Error: d:\Work\Sympli\figma\bugsnag\case\bugsnag-test\node_modules\@bugsnag\browser\dist\types\bugsnag.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
The missing file seems to be part of a third party library. TS files in published libraries are often a sign of a badly packaged library. Please open an issue in the library repository to alert its author and ask them to package the library using the Angular Package Format (https://goo.gl/jB3GVv).
```

tomlongridge:
Reviewed and tested it - this looks good. Merging changes via #633.

Thanks @GDreyV !

PR #636, opened 2019-10-29
Tests: Update CI to make iOS build less flakey

Cawllec:
- Fixes Turtle-ci version against 0.11.1 - A known to be working version
- Separates out run and build docker steps in local build scripts to ensure fast failures when publishing and building the test fixture


bengourley:
The S8 test that fails has been fixed upstream (in `next` where this will land).

PR #642, opened 2019-11-12
Update expected value in Samsung S8 browser tests

bengourley:
The default browser on the S8 device we use on BrowserStack has changed and this value in the list of expected results was no longer correct.

It caused failures like [this](https://buildkite.com/bugsnag/at-bugsnag-js/builds/1000#56f40a0a-3263-447b-a2b2-4c09de4da314/94-384).

@Cawllec looked in to seeing if we could pin the browser to what was previously being used but nothing was forthcoming so we decided the best thing to do was just update the expected value.


bengourley:
Merged in spite of Expo test failures – they were not related to the changes on this branch.

PR #644, opened 2019-11-12
V7: Migrate all lint tooling to ESLint

bengourley:
TSLint is deprecated. ESLint was already being used under the hood by standard JS. This PR updates the entire linting setup so that common ESLint dependecies are used for both JS and TS lint, just with different parsers/presets.

The changeset in this PR can be summarised as follows:

### Add and remove tooling
- Remove `tslint` everywhere it was included
- Remove `standard`
- Add `eslint` as a direct dependency and the modules required to get an equivalent-ish [standardjs](https://standardjs.com/rules.html) setup
- Add TypeScript plugins and rulesets from the `@typescript/eslint-*` namespace
- Remove `tslint.json` (multiple locations)
- Add `.eslintrc.js` and `.eslintrs.ts.js` which set up different rules for each kind of source file
- Update the linting commands in `package.json`

### Run `eslint --fix` on existing code

Although we used the same "standard" preset, it hadn't been updated in a while so there were some violations sprinkled around – mostly relating to whitespace or unnecessary quoting of object keys.

### Manually fix remaining lint errors

A few JS violations had to be resolved by hand – mostly telling ESLint to ignore the next line for things that needed to be "just so".

Since the TypeScript rules changed quite a lot from TSLint to `@typescript/eslint-*`, a few tweaks were required, including removing the `I*` prefix from interface names.

----

This PR is split off from work that was done in the `spec-refactor` branch.


.eslintrc.ts.js
```diff
@@ -0,0 +1,18 @@
+// eslint config for .ts / .d.ts files
```
djskinner: Consider combining both `.eslintrc.js` and `.eslintrc.ts.js` into a single file using [`overrides`](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns).

package.json
```diff
@@ -14,20 +14,29 @@
...
     "lerna": "^2.11.0",
-    "standard": "^11.0.1",
+    "typescript": "^3.6.4",
     "uglify-js": "^3.4.9"
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build --scope '@bugsnag/node' --scope '@bugsnag/browser' --scope '@bugsnag/expo' && lerna run build --ignore '@bugsnag/node' --ignore '@bugsnag/browser' --ignore '@bugsnag/expo'",
-    "test:lint": "standard",
+    "test:lint": "eslint . && eslint --no-eslintrc -c .eslintrc.ts.js --ext .ts .",
```
djskinner: [`--report-unused-disable-directives`](https://eslint.org/docs/user-guide/command-line-interface) is a nice flag to consider using.

packages/plugin-inline-script-content/inline-script-content.js
```diff
@@ -102,7 +102,7 @@ module.exports = {
       'Notification', 'SVGElementInstance', 'Screen', 'TextTrack', 'TextTrackCue', 'TextTrackList',
       'WebSocket', 'WebSocketWorker', 'Worker', 'XMLHttpRequest', 'XMLHttpRequestEventTarget', 'XMLHttpRequestUpload'
     ], o => {
-      if (!win[o] || !win[o].prototype || !win[o].prototype.hasOwnProperty || !win[o].prototype.hasOwnProperty('addEventListener')) return
+      if (!win[o] || !win[o].prototype || !Object.prototype.hasOwnProperty.call(win[o].prototype, 'addEventListener')) return
```
djskinner: A quick search suggests `Object.prototype.hasOwnProperty.call` is supported and preferred whereas `hasOwnProperty` doesn't exist for "host" objects such as `window`.  Possibly a slight change in behaviour to be aware of.

packages/browser/types/bugsnag.d.ts
```diff
@@ -2,21 +2,21 @@ import * as BugsnagCore from "@bugsnag/core";
 
 // overwrite config interface, adding browser-specific options
 declare module "@bugsnag/core" {
-  interface IConfig {
```
djskinner: I guess these changes to the external type names constitute a breaking change.

bengourley: 👍 yup this will be shipped in a major release with a whole bunch more breaking changes. Stay tuned!

packages/plugin-inline-script-content/inline-script-content.js
bengourley: Yeah, I presume this was the reason the linter told me change it!

packages/browser/types/bugsnag.d.ts
snmaynard: @bengourley @djskinner should we have an UPGRADING.md or whatever as part of this change that highlights that this is required? Should this be in the CHANGELOG as part of this PR?

bengourley: This PR is already merged but I'm happy to include changelog entries for subsequent PRs. Upgrading guide-wise I think that should be done once everything has landed on the `v71 integration branch.

PR #647, opened 2019-11-14
V7: Rename autoCaptureSessions -> autoTrackSessions and simplify validation logic

bengourley:
Per the updated spec, the option to disable sessions is called `autoTrackSessions`. In addition to making that update, a suggestion made here¹ was incorporated to make the session endpoint always required (this only applies when endpoints is set).

Previously the validation of the `autoCaptureSessions` field was dependent on whether a session endpoint was provided, which made the logic unnecessarily complex.

¹https://github.com/bugsnag/bugsnag-js/pull/630/files#r343663933

### DIFF
--- a/examples/browser-cdn/app.js
+++ b/examples/browser-cdn/app.js
@@ -23,8 +23,8 @@ var bugsnagClient = bugsnag({
   appVersion: '1.2.3',
 
   // Bugsnag can track the number of “sessions” that happen in your application,
-  // and calculate a crash rate for each release. This defaults to false.
-  autoCaptureSessions: true,
+  // and calculate a crash rate for each release. This defaults to true.
+  autoTrackSessions: true,
 
   // defines the release stage for all events that occur in this app.
   releaseStage: 'development',
--- a/packages/plugin-browser-session/session.js
+++ b/packages/plugin-browser-session/session.js
@@ -18,11 +18,6 @@ const sessionDelegate = {
       return sessionClient
     }
 
-    if (!sessionClient.config.endpoints.sessions) {
-      sessionClient._logger.warn('Session not sent due to missing endpoints.sessions configuration')
-      return sessionClient
-    }
-
     sessionClient._delivery.sendSession({
       notifier: sessionClient.notifier,
       device: sessionClient.device,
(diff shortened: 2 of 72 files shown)


examples/browser-cdn/app.js
```diff
@@ -23,8 +23,8 @@ var bugsnagClient = bugsnag({
   appVersion: '1.2.3',
 
   // Bugsnag can track the number of “sessions” that happen in your application,
-  // and calculate a crash rate for each release. This defaults to false.
-  autoCaptureSessions: true,
+  // and calculate a crash rate for each release. This defaults to true.
```
bengourley: the default value has not been updated in this PR, this comment was out of date.

packages/plugin-browser-session/session.js
```diff
@@ -18,11 +18,6 @@ const sessionDelegate = {
       return sessionClient
     }
 
-    if (!sessionClient.config.endpoints.sessions) {
```
bengourley: The config schema now validates that `endpoints.sessions` must now be defined so this branch of code is obsolete.

packages/plugin-server-session/session.js
```diff
@@ -37,11 +37,6 @@ const sendSessionSummary = client => sessionCounts => {
     return
   }
 
-  if (!client.config.endpoints.sessions) {
```
bengourley: As above, commenting here too for an equivalent change… the config schema now validates that `endpoints.sessions` must now be defined so this branch of code is obsolete.

test/browser/features/fixtures/auto_notify/script/a.html
```diff
@@ -8,7 +8,7 @@
       var API_KEY = decodeURIComponent(window.location.search.match(/API_KEY=([^&]+)/)[1])
       var bugsnagClient = bugsnag({
         apiKey: API_KEY,
-        endpoints: { notify: ENDPOINT },
+        endpoints: { notify: ENDPOINT, sessions: '/noop' },
```
bengourley: Loads of the end-to-end tests that relied on being allowed to not set a custom session endpoint now need some value to satisfy the config validation, but all of these tests only make assertions about errors.

I've set it to this value here (and many other places), which hopefully which suggests it's not used at all to appease the config validation.

PR #650, opened 2019-11-18
V7: Breadcrumb refactor

bengourley:
- Remove individual breadcrumb flags in favour of `enabledBreadcrumbTypes` option
  The `*BreadcrumbsEnabled` plugin-specfic options have been removed and breadcrumbs are now controlled by generic list in `enabledBreadcrumbTypes`.
- Rename `breadcrumb.{ name -> message, metaData -> metadata }`
  These fields have been renamed. Note that `metaData` -> `metadata` will happen elsewhere in the codebase too but I limited the scope to just breadcrumbs for this PR
- Update `leaveBreadcrumb()` type signature to be more explicit
  The `metadata` object is only allowed to be one level deep in breadcrumbs so the type signature now enforces that

### DIFF
--- a/packages/core/breadcrumb.js
+++ b/packages/core/breadcrumb.js
@@ -1,19 +1,19 @@
 const { isoDate } = require('./lib/es-utils')
 
 class BugsnagBreadcrumb {
-  constructor (name = '[anonymous]', metaData = {}, type = 'manual', timestamp = isoDate()) {
+  constructor (message, metadata, type, timestamp = isoDate()) {
     this.type = type
-    this.name = name
-    this.metaData = metaData
+    this.message = message
+    this.metadata = metadata
     this.timestamp = timestamp
   }
 
   toJSON () {
     return {
       type: this.type,
-      name: this.name,
+      name: this.message,
       timestamp: this.timestamp,
-      metaData: this.metaData
+      metaData: this.metadata
     }
   }
 }
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -124,19 +124,19 @@ class BugsnagClient {
     return this._sessionDelegate.startSession(this)
   }
 
-  leaveBreadcrumb (name, metaData, type, timestamp) {
+  leaveBreadcrumb (message, metadata, type) {
     if (!this._configured) throw new Error('client not configured')
 
     // coerce bad values so that the defaults get set
-    name = name || undefined
-    type = typeof type === 'string' ? type : undefined
-    timestamp = typeof timestamp === 'string' ? timestamp : undefined
-    metaData = typeof metaData === 'object' && metaData !== null ? metaData : undefined
+    message = typeof message === 'string' ? message : ''
+    type = typeof type === 'string' ? type : 'manual'
+    metadata = typeof metadata === 'object' && metadata !== null ? metadata : {}
 
-    // if no name and no metaData, usefulness of this crumb is questionable at best so discard
-    if (typeof name !== 'string' && !metaData) return
+    // if no message, discard
+    if (!message) return
 
-    const crumb = new BugsnagBreadcrumb(name, metaData, type, timestamp)
+    const crumb = new BugsnagBreadcrumb(message, metadata, type)
+
     // check the breadcrumb is the list of enabled types
     if (!this.config.enabledBreadcrumbTypes || !includes(this.config.enabledBreadcrumbTypes, crumb.type)) return
 
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -135,10 +135,10 @@ class BugsnagClient {
     // if no message, discard
     if (!message) return
 
-    const crumb = new BugsnagBreadcrumb(message, metadata, type)
-
     // check the breadcrumb is the list of enabled types
-    if (!this.config.enabledBreadcrumbTypes || !includes(this.config.enabledBreadcrumbTypes, crumb.type)) return
+    if (!this.config.enabledBreadcrumbTypes || !includes(this.config.enabledBreadcrumbTypes, type)) return
+
+    const crumb = new BugsnagBreadcrumb(message, metadata, type)
 
     // push the valid crumb onto the queue and maintain the length
     this.breadcrumbs.push(crumb)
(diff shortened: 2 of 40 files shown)


packages/core/breadcrumb.js
```diff
@@ -1,19 +1,19 @@
 const { isoDate } = require('./lib/es-utils')
 
 class BugsnagBreadcrumb {
-  constructor (name = '[anonymous]', metaData = {}, type = 'manual', timestamp = isoDate()) {
+  constructor (message, metadata, type, timestamp = isoDate()) {
```
bengourley: These default values have been removed in favour of the validation/default values being done in the `client.leaveBreadcrumb()` method rather than being split between there and here.

packages/core/client.js
```diff
@@ -124,19 +124,21 @@ class BugsnagClient {
...
-    timestamp = typeof timestamp === 'string' ? timestamp : undefined
-    metaData = typeof metaData === 'object' && metaData !== null ? metaData : undefined
+    message = typeof message === 'string' ? message : ''
+    type = typeof type === 'string' ? type : 'manual'
+    metadata = typeof metadata === 'object' && metadata !== null ? metadata : {}
 
-    // if no name and no metaData, usefulness of this crumb is questionable at best so discard
-    if (typeof name !== 'string' && !metaData) return
+    // if no message, discard
+    if (!message) return
```
bengourley: The spec doesn't allow a breadcrumb without a message, so we are more strict about that now.

packages/core/client.js
```diff
@@ -124,19 +124,21 @@ class BugsnagClient {
     return this._sessionDelegate.startSession(this)
   }
 
-  leaveBreadcrumb (name, metaData, type, timestamp) {
+  leaveBreadcrumb (message, metadata, type) {
```
bengourley: `timestamp` is not in the public method signature according to the new spec

packages/core/client.js
```diff
@@ -124,19 +124,21 @@ class BugsnagClient {
...
-    // if no name and no metaData, usefulness of this crumb is questionable at best so discard
-    if (typeof name !== 'string' && !metaData) return
+    // if no message, discard
+    if (!message) return
 
-    const crumb = new BugsnagBreadcrumb(name, metaData, type, timestamp)
+    const crumb = new BugsnagBreadcrumb(message, metadata, type)
+
+    // check the breadcrumb is the list of enabled types
+    if (!this.config.enabledBreadcrumbTypes || !includes(this.config.enabledBreadcrumbTypes, crumb.type)) return
```
bengourley: `enabledBreadcrumbTypes` is guaranteed by config to be either `null` or an array of valid types.

packages/core/client.js
```diff
@@ -205,13 +207,11 @@ class BugsnagClient {
...
 
       // only leave a crumb for the error if actually got sent
-      if (this.config.autoBreadcrumbs) {
-        this.leaveBreadcrumb(event.errorClass, {
-          errorClass: event.errorClass,
-          errorMessage: event.errorMessage,
-          severity: event.severity
-        }, 'error')
-      }
+      BugsnagClient.prototype.leaveBreadcrumb.call(this, event.errorClass, {
```
bengourley: Node's notifier monkey-patches the `leaveBreadcrumb()` method to log a warning that breadcrumbs aren't supported.

Calling the method via the prototype ensures "internal" calls to `leaveBreadcrumb()` don't trigger the log. I can explain more if this doesn't make sense!

packages/core/types/breadcrumb.d.ts
```diff
@@ -1,9 +1,8 @@
 declare class Breadcrumb {
-  public name: string;
-  public metaData: object;
+  public message: string;
+  public metadata: object;
   public type: string;
   public timestamp: string;
-  constructor(name: string, metaData?: object, type?: string, timestamp?: string);
```
bengourley: Breadcrumb construction is not public, it can only be done via `leaveBreadcrumb()`.

packages/plugin-console-breadcrumbs/console-breadcrumbs.js
```diff
@@ -1,14 +1,12 @@
...
 /*
  * Leaves breadcrumbs when console log methods are called
  */
 exports.init = (client) => {
   const isDev = /^dev(elopment)?$/.test(client.config.releaseStage)
 
-  const explicitlyDisabled = client.config.consoleBreadcrumbsEnabled === false
-  const implicitlyDisabled = (client.config.autoBreadcrumbs === false || isDev) && client.config.consoleBreadcrumbsEnabled !== true
-  if (explicitlyDisabled || implicitlyDisabled) return
+  if (!client.config.enabledBreadcrumbTypes || !includes(client.config.enabledBreadcrumbTypes, 'log') || isDev) return
```
bengourley: This change has the known impact of meaning console log breadcrumbs can't be enabled in development.

packages/core/config.js
```diff
@@ -68,10 +70,13 @@ module.exports.schema = {
...
-    defaultValue: () => true,
-    message: 'should be true|false',
-    validate: (value) => typeof value === 'boolean'
+  enabledBreadcrumbTypes: {
+    defaultValue: () => BREADCRUMB_TYPES,
+    message: `should be null or a list of available breadcrumb types (${BREADCRUMB_TYPES.join(',')})`,
+    validate: value => value === null || (isArray(value) && reduce(BREADCRUMB_TYPES, (accum, maybeType) => {
+      if (accum === false) return accum
+      return includes(BREADCRUMB_TYPES, maybeType)
+    }, true))
```
djskinner: I think this validation logic is wrong. Consider the following failing test case:

```
  describe('schema', () => {
    ...

    describe('enabledBreadcrumbTypes', () => {
      describe('validation', () => {
        it('fails when a supplied value is not a valid breadcrumb type', () => {
	  const enabledBreadcrumbTypesValidator = config.schema.enabledBreadcrumbTypes.validate;
          expect(enabledBreadcrumbTypesValidator(['UNKNOWN_BREADCRUMB_TYPE'])).toBe(false)
        })
      })
    })
  })
```

Gives `Expected true to be false.`

bengourley: Thanks for the failing test case 🙌 

I added it to the codebase and fixed the bug (I was iterating over `BREADCRUMB_TYPES` rather than `values`).

packages/core/client.js
bengourley: Ah yeah good point. We previously did it after because the default `type="manual"` was only populated after the instantiation, but now it's done inside this method we can do it earlier.

PR #651, opened 2019-11-20
v7: hoist devDependencies

djskinner:
Move common devDependencies to the root of the lerna repo. Moved dependencies have been updated. Note that devDependencies providing "binary" executables have not been hoisted. 

Reference: https://github.com/lerna/lerna/blob/master/README.md#common-devdependencies

### DIFF
--- a/packages/plugin-expo-app/package-lock.json
+++ b/packages/plugin-expo-app/package-lock.json
@@ -101,28 +101,6 @@
 				"to-fast-properties": "^2.0.0"
 			}
 		},
-		"@bugsnag/core": {
-			"version": "7.0.0-pre-alpha-ben.6",
-			"resolved": "https://registry.npmjs.org/@bugsnag/core/-/core-7.0.0-pre-alpha-ben.6.tgz",
-			"integrity": "sha512-NL33soHPbL+8xm/jIMPlIOIzUErEJgelDC3HOqSMNwPpGidE5768H28MUFGTF77TatRkihcZS5MNf4Ymt888Pg==",
-			"requires": {
-				"@bugsnag/cuid": "^3.0.0",
-				"@bugsnag/safe-json-stringify": "4.0.0",
-				"error-stack-parser": "^2.0.2",
-				"iserror": "0.0.2",
-				"stack-generator": "^2.0.3"
-			}
-		},
-		"@bugsnag/cuid": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/cuid/-/cuid-3.0.0.tgz",
-			"integrity": "sha512-LOt8aaBI+KvOQGneBtpuCz3YqzyEAehd1f3nC5yr9TIYW1+IzYKa2xWS4EiMz5pPOnRPHkyyS5t/wmSmN51Gjg=="
-		},
-		"@bugsnag/safe-json-stringify": {
-			"version": "4.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/safe-json-stringify/-/safe-json-stringify-4.0.0.tgz",
-			"integrity": "sha512-eb0yG+PNTdpRXIhfEq4IFi2wBfpE1DjrAlkyehinu9RxE1PKteqJioLmfQjvkG3UviGu/k5pcamYMxNuCAMIoQ=="
-		},
 		"ansi-styles": {
 			"version": "3.2.1",
 			"resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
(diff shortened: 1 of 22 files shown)


packages/plugin-expo-app/package-lock.json
```diff
@@ -101,28 +101,6 @@
...
-		"@bugsnag/cuid": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/cuid/-/cuid-3.0.0.tgz",
-			"integrity": "sha512-LOt8aaBI+KvOQGneBtpuCz3YqzyEAehd1f3nC5yr9TIYW1+IzYKa2xWS4EiMz5pPOnRPHkyyS5t/wmSmN51Gjg=="
-		},
-		"@bugsnag/safe-json-stringify": {
-			"version": "4.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/safe-json-stringify/-/safe-json-stringify-4.0.0.tgz",
-			"integrity": "sha512-eb0yG+PNTdpRXIhfEq4IFi2wBfpE1DjrAlkyehinu9RxE1PKteqJioLmfQjvkG3UviGu/k5pcamYMxNuCAMIoQ=="
-		},
```
djskinner: These lines were removed by lerna bootstrap. I believe this change is correct as the changed packages now reflect the other packages that do not contain these entries in their package-lock.json which makes things more consistent.

packages/plugin-expo-device/package-lock.json
```diff
@@ -101,28 +101,6 @@
...
-		"@bugsnag/cuid": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/cuid/-/cuid-3.0.0.tgz",
-			"integrity": "sha512-LOt8aaBI+KvOQGneBtpuCz3YqzyEAehd1f3nC5yr9TIYW1+IzYKa2xWS4EiMz5pPOnRPHkyyS5t/wmSmN51Gjg=="
-		},
-		"@bugsnag/safe-json-stringify": {
-			"version": "4.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/safe-json-stringify/-/safe-json-stringify-4.0.0.tgz",
-			"integrity": "sha512-eb0yG+PNTdpRXIhfEq4IFi2wBfpE1DjrAlkyehinu9RxE1PKteqJioLmfQjvkG3UviGu/k5pcamYMxNuCAMIoQ=="
-		},
```
djskinner: Also here. See previous comment

packages/plugin-react-native-global-error-handler/package-lock.json
```diff
@@ -101,28 +101,6 @@
...
-		"@bugsnag/cuid": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/cuid/-/cuid-3.0.0.tgz",
-			"integrity": "sha512-LOt8aaBI+KvOQGneBtpuCz3YqzyEAehd1f3nC5yr9TIYW1+IzYKa2xWS4EiMz5pPOnRPHkyyS5t/wmSmN51Gjg=="
-		},
-		"@bugsnag/safe-json-stringify": {
-			"version": "4.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/safe-json-stringify/-/safe-json-stringify-4.0.0.tgz",
-			"integrity": "sha512-eb0yG+PNTdpRXIhfEq4IFi2wBfpE1DjrAlkyehinu9RxE1PKteqJioLmfQjvkG3UviGu/k5pcamYMxNuCAMIoQ=="
-		},
```
djskinner: Also here. See previous comment

packages/plugin-react-native-unhandled-rejection/package-lock.json
```diff
@@ -101,28 +101,6 @@
...
-		"@bugsnag/cuid": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/cuid/-/cuid-3.0.0.tgz",
-			"integrity": "sha512-LOt8aaBI+KvOQGneBtpuCz3YqzyEAehd1f3nC5yr9TIYW1+IzYKa2xWS4EiMz5pPOnRPHkyyS5t/wmSmN51Gjg=="
-		},
-		"@bugsnag/safe-json-stringify": {
-			"version": "4.0.0",
-			"resolved": "https://registry.npmjs.org/@bugsnag/safe-json-stringify/-/safe-json-stringify-4.0.0.tgz",
-			"integrity": "sha512-eb0yG+PNTdpRXIhfEq4IFi2wBfpE1DjrAlkyehinu9RxE1PKteqJioLmfQjvkG3UviGu/k5pcamYMxNuCAMIoQ=="
-		},
```
djskinner: Also here. See previous comment

packages/plugin-expo-app/package-lock.json
bengourley: Yeah it looks like these got a little messed up and stopped linking locally to `@bugsnag/core` and instead pulling that prerelease version down from npm.

I agree these changes seem correct.

djskinner:
> I'm somewhat unsure about hoisting non-common dependencies – i.e `ncp`, `timekeeper` and `rimraf`. If they are only used by one package, is it of benefit to move the dependency up to the top level?

Yeah I hesitated on this one initially but thought it would be worth a discussion. Happy to change the PR to only hoist dependencies that are currently used in > 1 package.

The principal reason I managed to sell myself on the idea was considering what happens when a developer comes to use `timekeeper` in a second package. In all likelihood they would not be aware of the preference to hoist dev dependencies when used in > 1 package. So they add `timekeeper` to package2, likely getting a different version. Hopefully it gets picked up at PR review and the hoisting can be done then (with the associated package-lock churn). If not caught immediately we'll be seeing this "hoist devDependencies" more often once individual cases are discovered. All that assumes they were aware `timekeeper` was being used in the first place (which is less obvious if not stated in the root package.json) and didn't opt to use a different package offering the same functionality because they weren't aware.

So on balance I felt it was simpler, cleaner and clearer to have all devDependencies hoisted where possible rather than having to deal with the "oh, this is now used in more than one package" scenario each time.

Not that it's a convincing argument in and of itself but it's interesting to note that [lerna does it](https://github.com/lerna/lerna/search?q=touch&unscoped_q=touch).

PR #653, opened 2019-11-25
Add React-native buildkite tests

Cawllec:
This change adds a test fixture that will allow quick setup of tests when the React-native notifier is moved into this monorepo.


test/react-native/scripts/build-ios.sh
```diff
@@ -0,0 +1,14 @@
+#!
```
bengourley: I'd always do `set -e` so that any errors cause this script to exit with an error.

I'd also be more explicit in the shebang so that this script behaves the same wherever it is run `#!/bin/bash` or `#!/bin/env bash`

.dockerignore
```diff
@@ -1,11 +1,6 @@
 node_modules
-.git
-examples
 **/node_modules
+!packages/plugin-node-surrounding-code/test/fixtures/node_modules/mock-module/index.js
```
bengourley, line 3: I would slacken the specificity of this

```suggestion
!packages/plugin-node-surrounding-code/test/fixtures/node_modules/
```

or even better can you put another `.dockerignore` in that package?

dockerfiles/Dockerfile.publisher
```diff
@@ -0,0 +1,27 @@
...
+
+COPY package*.json ./
+RUN npm install
+
+COPY . .
+
+RUN npm run bootstrap
+RUN npm run build
+
+RUN git checkout .
```
bengourley, line 24: does this happen due to the `npm run bootstrap` command making changes to lockfiles?

If so it would be good to avoid that in the first place. You can add npm args to lerna with `--` so you could add a new command to `package.json`:

```
  "bootstrap:ci": "lerna bootstrap -- --no-save"
```

I think this should work, see [npm's docs](https://docs.npmjs.com/files/package-locks#using-locked-packages).

dockerfiles/Dockerfile.publisher
```diff
@@ -0,0 +1,27 @@
...
+
+ARG PUBLISH_URL
+RUN ./node_modules/.bin/lerna publish --canary --yes --force-publish --registry $PUBLISH_URL --dist-tag $(git rev-parse --short HEAD) prerelease
```
bengourley, line 27: Do we need to heed this warning?

> If you have publish canary releases from multiple active development branches in CI, it is recommended to customize the --preid and --dist-tag <tag> on a per-branch basis to avoid clashing versions.
[source](https://github.com/lerna/lerna/tree/master/commands/publish#--canary)

It looks like the commit hash is used as the npm dist tag, not as the actual version of the package?

dockerfiles/Dockerfile.react-native-android-builder
```diff
@@ -0,0 +1,50 @@
...
+RUN echo "_auth=$REG_BASIC_CREDENTIAL" >> ~/.npmrc
+RUN echo "email=$REG_NPM_EMAIL" >> ~/.npmrc
+RUN echo "always-auth=true" >> ~/.npmrc
+
+ARG REGISTRY_URL
+
+RUN npm install --registry $REGISTRY_URL
+RUN npm install @bugsnag/react-native@$(git rev-parse --short HEAD) --registry $REGISTRY_URL
+
+RUN ./node_modules/.bin/react-native link
```
bengourley, line 45: This only needs to happen on RN pre 0.60. I think it will just log a warning in 0.60+ but they may remove it completely at some point.

test/react-native/features/fixtures/app/App.js
```diff
@@ -0,0 +1,106 @@
+import React, {Component} from 'react';
```
bengourley, line 1: this file mixes single/double quotes and I guess is just generally not linted.

Would be good to extend the eslint configuration to cover these files and then fix the lint errors.

test/react-native/features/fixtures/app/scenarios/HandledCaughtErrorScenario.js
```diff
@@ -0,0 +1,12 @@
+import Scenario from "./Scenario";
+import Bugsnag from '@bugsnag/react-native';
```
bengourley, line 2: again, mixed quote style

test/react-native/scripts/build-ios.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
+
+if [ -z "$REACT_NATIVE_VERSION" ]; then
+    echo "The environment variable 'REACT_NATIVE_VERSION' must be set"
+    exit 1
+fi
+
+cp -r test/react-native/features/fixtures/app test/react-native/features/fixtures/$REACT_NATIVE_VERSION/app
```
bengourley, line 10: I guess we will have to be careful that everything in `test/react-native/features/fixtures/app` is compatible with all versions of RN that we test with.

dockerfiles/Dockerfile.react-native-android-builder
Cawllec: I agree, considering we only have a single fixture that requires this I'll look at adding a conditional to only run on that version (0.55).

test/react-native/scripts/build-ios.sh
Cawllec: Yes.  At the moment the JS is fine to run across all versions of RN, but should it occur that this is not the case this may have to be refactored into separate scenario sets.

PR #654, opened 2019-11-26
V7: Callback refactor

bengourley:
- removes `ignore()` from event – now the only way to ignore is to `return` false in ~a `beforeSend`~ an `onError` callback.
- rename `beforeSend` -> `onError`
- refactors the `run-before-send` and `async-some` modules to make more sense (please let me know if this does seem more straightforward now… it took a lot for me to get my head around it and I wrote it!)


packages/core/types/event.d.ts
```diff
@@ -42,7 +42,6 @@ declare class Event {
     originalError?: any,
   );
 
-  public isIgnored(): boolean;
   public ignore(): void;
```
djskinner: Does this need to be removed?

packages/core/lib/async-every.js
```diff
@@ -0,0 +1,26 @@
...
+//   - or the end of the array is reached
+// the callback (cb) will be passed (null, false) if any of the items in arr
+// caused fn to call back with false, otherwise it will be passed (null, true)
+module.exports = (arr, fn, cb) => {
+  let index = 0
+
+  const next = () => {
+    if (index >= arr.length) return cb(null, true)
+    fn(arr[index], (err, result) => {
+      if (err) return cb(err, true)
```
djskinner: `cb(err, true)` - would this be better being `cb(err, undefined)` or simply `cb(err)`

packages/core/lib/test/async-every.test.js
```diff
@@ -0,0 +1,71 @@
...
+        expect(inFlight).toBe(1)
+        inFlight--
+        cb(null)
+      }, 10)
+    }, (err, result) => {
+      expect(err).toBe(null)
+      expect(result).toBe(true)
+      done()
+    })
+  })
```
djskinner: Looks like a duplicate of the above

packages/core/lib/callback-runner.js
```diff
@@ -0,0 +1,46 @@
+const some = require('./async-every')
+
+module.exports = (callbacks, event, onCallbackError, cb) => {
+  // This function is how we support different kinds of callback:
+  //  - synchronous - return value
+  //  - node-style async with callback - cb(err, value)
+  //  - promise/thenable - resolve(value)
+  // It normalises each of these into the lowest common denominator – a node-style callback
+  const runMaybeAsyncCallback = (fn, cb) => {
+    if (typeof fn !== 'function') return cb(null, true)
```
djskinner: I wonder if this also ought to be async i.e. called within `setTimeout`? Also applies to the other sync `cb` calls. As it stands `runMaybeAsyncCallback` is maybe going to call `cb` sync or async. Would it be cleaner if it was always async?

Not sure it matters in our case but it can have an effect if consumers are expecting something to be called on the next event loop but it gets called synchronously.

packages/core/lib/callback-runner.js
```diff
@@ -0,0 +1,46 @@
...
+        const ret = fn(event)
+        // check if it returned a "thenable" (promise)
+        if (ret && typeof ret.then === 'function') {
+          return ret.then(
+            // resolve
+            val => setTimeout(() => cb(null, val), 0),
+            // reject
+            err => {
+              setTimeout(() => {
+                onCallbackError(err)
```
djskinner: `onCallbackError` is not covered by unit tests

packages/core/lib/callback-runner.js
```diff
@@ -0,0 +1,46 @@
...
+    if (typeof fn !== 'function') return cb(null, true)
+    try {
+      // if function appears sync…
+      if (fn.length !== 2) {
+        const ret = fn(event)
+        // check if it returned a "thenable" (promise)
+        if (ret && typeof ret.then === 'function') {
+          return ret.then(
+            // resolve
+            val => setTimeout(() => cb(null, val), 0),
```
djskinner: both `setTimeout(..., 0)` and `setTimeout(...)` are used. Maybe just stick with the succinct one.

packages/core/lib/async-every.js
bengourley: Removed the second parameter as it was not used in the case of an error.

packages/core/lib/callback-runner.js
bengourley: Added a test for this.

packages/core/lib/callback-runner.js
bengourley: Removed the `0` version.

packages/core/lib/callback-runner.js
bengourley: I totally see your point. Made them all async now.

bengourley: Reneged on this (pushed a commit and then removed it) – there are whole bunch of unit tests that fail because they run synchronously after calling `notify()` – expecting the synchronous callbacks to have run. This isn't a problem in and of itself, but it highlighted that a lot of the internal callbacks are synchronous and they are used to get the device/app state at the moment of an error – if they run asynchronously, there's the chance for things to change after the error has happened.

One of the notifier spec requirements is that "callbacks run in the context of the error" and if we enforce that even sync callbacks should be run asynchronously I'd say it violates that.

PR #656, opened 2019-11-29
V7: Simplify client configuration and store config privately

bengourley:
The existing configuration routine was laborious and can be simplified to be done in one fell swoop

This is just an implementation detail really – this logic is done internally in each exported notififer – but it removes some methods from the public client interface (making it compliant with the spec).

```diff
  var NOTIFIER_DETAILS = { name: 'Bugsnag x', url: 'https://bugsnag.com', version: '0.0' }
- const client = new Client(NOTIFIER_DETAILS)
- client.setOptions(userOpts)
- client.configure(customSchema)
+ const client = new Client(userOpts, schema, NOTIFIER_DETAILS)
```

One benefit of the configuration being done via the constructor is that we can remove some of the "client not configured" errors – it's now impossible to have an instance of a client that's not configured.

Since loads of unit tests create `Client`s directly, this changeset is pretty big, but it does end up removing a lot of boilerplate lines from those tests.

The other change in this PR is to store config at `client._config` (`_` to signify that it is private) and remove its `public` definition from the types to match up with the spec.

### DIFF
--- a/packages/browser/src/notifier.js
+++ b/packages/browser/src/notifier.js
@@ -33,28 +33,13 @@ module.exports = (opts) => {
   // handle very simple use case where user supplies just the api key as a string
   if (typeof opts === 'string') opts = { apiKey: opts }
 
-  // support renamed/deprecated options
-
-  let warningMessage = ''
-
-  if (opts.endpoints && opts.endpoints.notify && !opts.endpoints.sessions) {
-    warningMessage += 'notify endpoint is set but sessions endpoint is not. No sessions will be sent.'
-  }
-
-  const bugsnag = new Client({ name, version, url })
-
-  bugsnag.setOptions(opts)
+  // configure a client with user supplied options
+  const bugsnag = new Client(opts, schema, { name, version, url })
 
   // set delivery based on browser capability (IE 8+9 have an XDomainRequest object)
   bugsnag.delivery(window.XDomainRequest ? dXDomainRequest : dXMLHttpRequest)
 
-  // configure with user supplied options
-  // errors can be thrown here that prevent the lib from being in a useable state
-  bugsnag.configure(schema)
-
-  if (warningMessage) bugsnag._logger.warn(warningMessage)
-
-  // always-on browser-specific plugins
+  // add browser-specific plugins
   bugsnag.use(pluginDevice)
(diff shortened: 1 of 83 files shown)


packages/browser/src/notifier.js
```diff
@@ -33,28 +33,13 @@ module.exports = (opts) => {
   // handle very simple use case where user supplies just the api key as a string
   if (typeof opts === 'string') opts = { apiKey: opts }
 
-  // support renamed/deprecated options
-
-  let warningMessage = ''
-
-  if (opts.endpoints && opts.endpoints.notify && !opts.endpoints.sessions) {
-    warningMessage += 'notify endpoint is set but sessions endpoint is not. No sessions will be sent.'
```
bengourley: it's now impossible to get into this state, as of #647

packages/plugin-window-onerror/test/onerror.test.js
```diff
@@ -53,7 +46,7 @@ describe('plugin: window onerror', () => {
     //   const client = new Client(VALID_NOTIFIER)
     //   const payloads = []
     //   client.setOptions({ apiKey: 'API_KEY_YEAH' })
-    //   client.configure()
+    //   client._configure()
```
djskinner: It's only a comment but I think this should be unchanged. Maybe the all commented code needs updating or just removing? Same for other places in this file.

packages/plugin-window-unhandled-rejection/test/unhandled-rejection.test.js
```diff
@@ -64,7 +59,7 @@ describe('plugin: unhandled rejection', () => {
     // it('works with DOMExceptions', done => {
     //   const client = new Client(VALID_NOTIFIER)
     // setOptions({ apiKey: 'API_KEY_YEAH' })
-    //   client.configure()
+    //   client._configure()
```
djskinner: Commented code as per other comment

packages/plugin-window-onerror/test/onerror.test.js
bengourley: We have a backlog ticket to convert all these commented out unit tests into maze runner tests.

I'll undo the change.

PR #658, opened 2019-11-29
V7: Metadata refactor

bengourley:
- Rename `metaData` -> `metadata`
- Move the storage to a private property `_metadata` and remove from the public interface
- Add `add/get/clearMetadata()` methods via a shared delegate to `Event` and `Client`


packages/core/lib/metadata-delegate.js
```diff
@@ -0,0 +1,53 @@
+module.exports = {
```
djskinner: Unit test would be useful, particularly as documentation to show how it relies on the binding of function context when called. It's not a common pattern to see `this` out of class syntax these days so it's not immediately obvious what `this` is referring to based on this code alone, indeed it depends entirely on how this module is imported/called. So a unit test showing it being used as a sort of class mixin would be useful.

packages/core/event.js
```diff
@@ -21,7 +22,7 @@ class BugsnagEvent {
     this.errorClass = stringOrFallback(errorClass, '[no error class]')
     this.errorMessage = stringOrFallback(errorMessage, '[no error message]')
     this.groupingHash = undefined
-    this.metaData = {}
+    this._metadata = {}
```
djskinner: What are the benefits of this mixin pattern compared with pure functions?

For example:

```
constructor ()
  ...
  this._metadata = {}
}

addMetadata (section, ...args) {
    this._metadata = addMetadata(this._metadata, section, ...args);
    return this;
}

```

It's not clear from looking at this class why `this._metadata = {}` is being assigned as it is never mutated in this class but in fact it's crucial for the `'./lib/metadata-delegate'` mixin to work properly, for omitting it would cause a TypeError.

In its current form `metadataDelegate` has a fairly complex API in that requires `this._metadata = {}` defining in the consuming class and calls to its functions being bound so the same context.

packages/core/lib/metadata-delegate.js
```diff
@@ -0,0 +1,53 @@
...
+      return this
+    }
+
+    // clear a single value from a section
+    if (this._metadata[section]) {
+      delete this._metadata[section][key]
+      return this
+    }
+
+    return this
```
djskinner: Is it part of the spec that these methods return `this` to form a fluent API? I recall also seeing it in previous PRs. Is it necessary/used? I've not seen any test coverage around it as far as I can remember and the type declarations do not expose this fact (at least the new metadata methods in this PR are returning `void`, not `this`). Just wondering if we could get rid and shave a few bytes?

packages/core/types/common.d.ts
```diff
@@ -94,16 +94,6 @@ export interface Session {
...
 
-export interface NotifyOpts {
-  context?: string;
-  device?: object;
-  request?: object;
-  user?: object;
-  metaData?: object;
-  severity?: "info" | "warning" | "error";
-  onError?: OnError;
-}
```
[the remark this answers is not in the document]
bengourley: hmm, it should have happened in #655 – I must have messed up somehow. Since #655 is already merged `NotifyOpts` is no longer used but I can split it out into a separate PR if you prefer?

packages/core/event.js
bengourley: I think it would be a lie to say it's "pure" but I have pushed an update using your approach, as I think it's much better. We can see if the size diff is ±93 bytes!

packages/core/lib/metadata-delegate.js
bengourley: Yeah it's not in the spec – I've removed them.

packages/core/lib/metadata-delegate.js
bengourley: Even though this is tested via the client/event tests, do you think we should add more – given we've now switched away from using `this`?

packages/core/event.js
bengourley: The diff is now 67 bytes (which includes the removal of all the `return this`) so it's smaller and better 🙌

packages/core/types/common.d.ts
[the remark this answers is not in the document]
djskinner: Can leave it in this PR for me, just wanted to be sure it was right.

packages/core/lib/metadata-delegate.js
djskinner: Ah yeah, I didn't click on it was being covered as part of integration tests.

packages/core/lib/metadata-delegate.js
djskinner: Nice, I think there are potentially some more to go back and clean too. I will take a look this week.

PR #659, opened 2019-12-02
V7: Remove non-public methods from client interface

bengourley:
Replaces `delivery()` with `_setDelivery()` and replaces `logger()`/`sessionDelegate()` with direct property access.

These methods were only used internally, but they were on the the public interface.


packages/core/client.js
```diff
@@ -111,21 +111,11 @@ class BugsnagClient {
     return this._plugins[`~${name}~`]
   }
 
-  delivery (d) {
+  _setDelivery (d) {
     this._delivery = d(this)
     return this
```
djskinner: I don't think this return `this` is needed

PR #662, opened 2019-12-04
fix(plugin-network-breadcrumbs): fetch accepts Request instance as an argument

djskinner:
Fixes the fetch monkey-patch to accept a Request instance as a first argument.

### DIFF
--- a/packages/plugin-network-breadcrumbs/network-breadcrumbs.js
+++ b/packages/plugin-network-breadcrumbs/network-breadcrumbs.js
@@ -119,8 +119,9 @@ const monkeyPatchFetch = () => {
   if (!('fetch' in win) || win.fetch.polyfill) return
 
   const oldFetch = win.fetch
-  win.fetch = function fetch (...args) {
-    const [urlOrRequest, options] = args
+  win.fetch = function fetch () {
+    const urlOrRequest = arguments[0]
+    const options = arguments[1]
 
     let method
     let url = null
@@ -145,7 +146,7 @@ const monkeyPatchFetch = () => {
 
     return new Promise((resolve, reject) => {
       // pass through to native fetch
-      oldFetch(...args)
+      oldFetch(...arguments)
         .then(response => {
           handleFetchSuccess(response, method, url)
           resolve(response)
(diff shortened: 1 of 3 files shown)


packages/plugin-network-breadcrumbs/network-breadcrumbs.js
```diff
@@ -120,11 +120,21 @@ const monkeyPatchFetch = () => {
...
-    if (options && options.method) {
-      method = options.method
+    let url
+
+    if (typeof urlOrRequest === 'string') {
+      url = urlOrRequest
+      if (options && options.method) {
+        method = options.method
+      }
+    } else {
```
bengourley: this block is entered when `typeof urlOrRequest !== 'string'` – it is then assumed to be an object, but it could also be `null` or `undefined` at this point.

If you call `fetch(null)` it causes the following error to be reported to Bugsnag:

```
TypeError: urlOrRequest is null
    fetch network-breadcrumbs.js:134
    <anonymous> (index):20
```

packages/plugin-network-breadcrumbs/network-breadcrumbs.js
```diff
@@ -120,11 +120,21 @@ const monkeyPatchFetch = () => {
...
+      url = urlOrRequest.url
+      method = urlOrRequest.method
```
bengourley: According to [MDN's `fetch()` docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) the second parameter `options` can be an object containing options such as `method`.

I tried it in Firefox and it seems that `options.method` takes precedence over `urlOrRequest.method`, so that should be accounted for here.

packages/plugin-network-breadcrumbs/test/network-breadcrumbs.test.js
```diff
@@ -179,6 +186,54 @@ describe('plugin: network breadcrumbs', () => {
...
+        name: 'fetch() succeeded',
+        metaData: {
+          status: 200,
+          request: 'PUT /'
+        }
+      }))
+      done()
+    })
+  })
+
```
bengourley: Please can you add a test cases for the following:

- `fetch(null)`
- `fetch(new Request('/foo'), { method: 'PUT' })`

PR #664, opened 2019-12-05
V7: notify() and Event construction refactor

bengourley:
This is a pretty substantial PR. I've outlined the main changes under the headings below. I also spent a lot of time manually testing on different browsers.

### Refactor `notify()` to conform with spec

- `notify()`no longer accepts pre-constructed events, and is now only for manually reporting errors
- `_notify()` has been created to report automatically detected errors, where plugins send events they have constructed themselves

### Rehome and reuse `normaliseError()`

- The logic in the `normaliseError()` function (which deals with extracting an error from whatever was provided) was previously only used for manual `notify()` calls. This has been moved to `event.js` where it is now used on all events – this meant that code which duplicated this functionality could be removed, such as:
  - automatic error detection in browsers
  - `error-to-event.js` lib

### Clean up and fix bugs in stacktrace generation logic

- The logic around ensuring we have a stacktrace, and making sure the correct number of internal Bugsnag frames are removed from the stacktrace has been cleaned up, and in the process a few bugs were fixed
- `Event` previously had two exported methods `getStacktrace()` and `ensureEvent()` – these have had their obsolete parts removed and combined into one exported `Event.create()` method (which is not fully public, but is required by plugin implementors)

### Use a consistent pattern for reporting non-errors

Some values coerce well into errors – for example, if somebody calls `notify('error message')` we can just construct an error with the message `error message` and get a stacktrace from that. In other situations, anything less than an `Error` is not of very much use (unhandled rejections, for example).

There were various ways in which this library would report whatever it received to the dashboard, and also try to communicate "we couldn't do much with this, but we added what we received to `metadata`" and this has now been consolidated into a single pattern:

An error with the name `InvalidError` shall be constructed, where the message is: `${component} received a non-error. See "${component}" tab for more detail.`
Whatever was recevied shall then be added to the `${component}` metadata tab under the key `non-error parameter`.


bengourley:
A concrete example where the consolidated non-error reporting logic has meant an improvement in the payload is the following – this is in IE8 where `notify({ nestedNotifyObject: true })` has been called.

```diff
5,6c5,6
<       "errorClass": "Error",
<       "message": "Bugsnag usage error. notify() expected error/opts parameters, got unsupported object",
---
>       "errorClass": "Bugsnag::InputError",
>       "message": "notify() received a non-error. See \"notify()\" tab for more detail.",
37a37,41
>     "notify()": {
>       "non-error parameter": {
>         "nestedNotifyObject": true
>       }
>     },

**Comment by bengourley** [comment]
A concrete example where a bug has been fixed by the cleanup of the stacktrace generation logic is the following – this is in IE11 where a `throw 'global throw string'` was called. The method was incorrectly marked as `anonymous` and an extra stackframe was added (there should only be one).

```diff
     {
-      "errorClass": "window.onerror",
+      "errorClass": "Error",
       "message": "global throw string",
       "stacktrace": [
         {
           "file": "http://localhost:5000/scratch/notify-refactor/test-cases/global-throw-string.js",
-          "method": "anonymous",
           "lineNumber": 1,
           "columnNumber": 1
        },
-       {
-         "method": "anonymous"
-       }
      ],
      "type": "browserjs"
    }
```

packages/core/event.js
```diff
@@ -126,33 +127,130 @@ const stringOrFallback = (str, fallback) => typeof str === 'string' && str ? str
...
+  }
+  return event
+}
+
+const makeSerialisable = (err) => {
+  if (err === null) {
+    return 'null'
+  } else if (err === undefined) {
+    return 'undefined'
+  } else if (isError(err)) {
```
djskinner: Struggling to understand this part. Under what conditions do we have both `error.name === 'InvalidError'` and `isError(err) === true`?

packages/core/event.js
```diff
@@ -126,33 +127,130 @@ const stringOrFallback = (str, fallback) => typeof str === 'string' && str ? str
...
+    return {
+      [Object.prototype.toString.call(err)]: {
+        name: err.name,
+        message: err.message,
+        code: err.code,
+        stack: err.stack
+      }
+    }
+  } else {
+    return err
```
djskinner: Does this need serialising, e.g. `String(err)` or does `addMetadata` take care of that?

packages/core/event.js
bengourley: I need to take another look at this. This logic used to live in the rejection handler and it was for when the rejection reason did not have a `stack`, but it was an error – such as a `DOMException`.

packages/core/event.js
bengourley: the metadata serialisation takes care of this – the cases where it doesn't serialise are captured above: `null`/`undefined`/`DOMException`.

packages/core/event.js
```diff
@@ -126,33 +127,130 @@ const stringOrFallback = (str, fallback) => typeof str === 'string' && str ? str
...
+  let internalFrames = 0
+
+  const createAndLogInputError = (reason, value) => {
+    if (logger) logger.warn(`${component} received a non-error: "${reason}"`)
+    const err = new Error(`${component} received a non-error. See "${component}" tab for more detail.`)
+    err.name = 'InvalidError'
+    return err
+  }
+
+  if (!tolerateNonErrors) {
```
bengourley: In some cases:

- the promise rejection handler (both in the browser and node)
- the node `uncaughtException` handler

we are really limited in what we can do to get a stacktrace. So we use this option to ensure that the resulting error communicates as such.

packages/core/event.js
```diff
@@ -126,33 +127,130 @@ const stringOrFallback = (str, fallback) => typeof str === 'string' && str ? str
...
+    if (isError(maybeError)) {
+      error = maybeError
+    } else {
+      error = createAndLogInputError(typeof maybeError, maybeError)
+      internalFrames += 2
```
bengourley: The +2 is because when know the stacktrace will include a frame from `normaliseError` and `createAndLogInputError` at this point.

packages/core/types/client.d.ts
```diff
@@ -29,6 +29,11 @@ declare class Client {
     onError?: common.OnError,
     cb?: (err: any, event: Event) => void,
   ): void;
+  public _notify(
```
bengourley: This isn't intended to be part of the public interface, but is required for plugins which want to report unhandled errors.

packages/core/types/event.d.ts
```diff
@@ -1,16 +1,14 @@
...
-    error: any,
-    errorFramesToSkip?: number,
-    generatedFramesToSkip?: number,
-  ): Stackframe[];
-
-  public static ensureEvent(
-    eventOrError: any,
-    errorFramesToSkip?: number,
-    generatedFramesToSkip?: number,
+  public static create(
```
bengourley: As per `_notify()` this isnt part of the public interface, but it is required for plugins to create unhandled errors.

packages/core/types/event.d.ts
```diff
@@ -32,14 +30,6 @@ declare class Event {
   };
   public originalError: any;
 
-  constructor(
```
bengourley: the constructor should not be public

test/node/features/restify.feature
```diff
@@ -53,10 +53,9 @@ Scenario: throwing non-Error error
   And the event "unhandled" is true
   And the event "severity" equals "error"
   And the event "severityReason.type" equals "unhandledErrorMiddleware"
-  And the exception "errorClass" equals "Error"
-  And the exception "message" matches "^Handled a non-error\."
+  And the exception "errorClass" equals "InvalidError"
+  And the exception "message" matches "^restify middleware received a non-error\."
   And the exception "type" equals "nodejs"
-  And the "file" of stack frame 0 equals "node_modules/@bugsnag/plugin-restify/dist/bugsnag-restify.js"
```
bengourley: The fact that this (and other similar tests) had this stackframe made it seem like the errors came from Bugsnag. The updated stacktrace mechanism means that this is no longer the top frame, but I couldn't find a way to assert that this was _not_ equal to this string.

packages/core/event.js
djskinner: A code comment with this info on `tolerateNonErrors` might be useful.

packages/plugin-node-uncaught-exception/uncaught-exception.js
```diff
@@ -1,15 +1,14 @@
-const createEventFromErr = require('@bugsnag/core/lib/event-from-error')
-
 let _handler
 module.exports = {
   init: client => {
     if (!client._config.autoDetectErrors) return
     _handler = err => {
-      client.notify(createEventFromErr(err, {
+      const event = client.BugsnagEvent.create(err, true, {
```
djskinner: Following on from https://github.com/bugsnag/bugsnag-js/pull/664/files#r354934957, why do we tolerate non errors here but not in `plugin-node-unhandled-rejection` and `plugin-window-unhandled-rejection`?

packages/core/event.js
bengourley: I moved some of this logic back into the promise rejection handler plugin, as it was only ever used by that.

bengourley: To answer your original question:

> Under what conditions do we have both `error.name === 'InvalidError'` and `isError(err) === true`?

I think it can never have been the case.

packages/plugin-node-uncaught-exception/uncaught-exception.js
bengourley: I've updated this – it should be the same as the ones you mentioned.

PR #665, opened 2019-12-06
V7: Callback refactor

bengourley:
- Add public `onBreadcrumb` and `onSession` breadcrumbs
- Add private `onSessionPayload` breadcrumbs (will be required by a subsequent PR for attaching data to a session)
- Add client methods for adding/removing all kinds of callback
- Add config options for supplying public callbacks

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -7,6 +7,7 @@ const inferReleaseStage = require('./lib/infer-release-stage')
 const isError = require('./lib/iserror')
 const runCallbacks = require('./lib/callback-runner')
 const metadataDelegate = require('./lib/metadata-delegate')
+const runSyncCallbacks = require('./lib/sync-callback-runner')
 
 const noop = () => {}
 
@@ -135,18 +136,9 @@ class BugsnagClient {
 
   startSession () {
     const session = new BugsnagSession()
-    // run synchronous onSession callbacks
-    let ignore = false
-    const cbs = this._cbs.s.slice(0)
-    while (!ignore) {
-      if (!cbs.length) break
-      try {
-        ignore = cbs.pop()(session) === false
-      } catch (e) {
-        this._logger.error('Error occurred in onSession callback, continuing anyway…')
-        this._logger.error(e)
-      }
-    }
+
+    // run onSession callbacks
+    const ignore = runSyncCallbacks(this._cbs.s.slice(0), session, 'onSession', this._logger)
 
     if (ignore) {
       this._logger.debug('Session not started due to onSession callback')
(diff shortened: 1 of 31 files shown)


packages/core/test/client.test.js
```diff
@@ -474,6 +499,86 @@ describe('@bugsnag/core/client', () => {
...
+
+      expect(c._cbs.e.length).toBe(1)
+      expect(c._cbs.s.length).toBe(1)
+      expect(c._cbs.b.length).toBe(1)
+
+      c.startSession()
+      expect(sSpy).toHaveBeenCalledTimes(1)
+      c.notify(new Error(), () => {}, () => {
+        expect(bSpy).toHaveBeenCalledTimes(1)
+        expect(bSpy).toHaveBeenCalledTimes(1)
```
djskinner: duplicate assertion

packages/core/test/client.test.js
```diff
@@ -474,6 +499,86 @@ describe('@bugsnag/core/client', () => {
...
+
+        c.removeOnError(eSpy)
+        c.removeOnSession(sSpy)
+        c.removeOnBreadcrumb(bSpy)
+
+        c.startSession()
+        expect(sSpy).toHaveBeenCalledTimes(1)
+        c.notify(new Error(), () => {}, () => {
+          expect(bSpy).toHaveBeenCalledTimes(1)
+          expect(bSpy).toHaveBeenCalledTimes(1)
```
djskinner: duplicate assertion

packages/core/client.js
```diff
@@ -114,11 +134,53 @@ class BugsnagClient {
...
+    const cbs = this._cbs.s.slice(0)
+    while (!ignore) {
+      if (!cbs.length) break
+      try {
+        ignore = cbs.pop()(session) === false
+      } catch (e) {
+        this._logger.error('Error occurred in onSession callback, continuing anyway…')
+        this._logger.error(e)
+      }
+    }
```
djskinner: This could be refactored although it doesn't help to reduce the bundle diff:

```
  runSyncCallbacks (callbacks, callbackArg, callbackType) {
    let ignore = false
    const cbs = callbacks.slice(0)
    while (!ignore) {
      if (!cbs.length) break
      try {
        ignore = cbs.pop()(callbackArg) === false
      } catch (e) {
        this._logger.error(`Error occurred in ${callbackType} callback, continuing anyway…`)
        this._logger.error(e)
      }
    }
    return ignore
  }
```

packages/core/test/client.test.js
bengourley: Good spot – this was actually meant to make an assertion about `eSpy`.

packages/core/client.js
bengourley: I'll take this – we can reuse it for the internal session payload callbacks too.

PR #667, opened 2019-12-09
CI Lerna improvements

Cawllec:
Attempts to combat ongoing Lerna bootstrapping failures by:
- Updating Lerna to the latest version (this seems to have little effect, but will enable new/faster lerna features)
- Setting a concurrency limit of 2 on the bootstrapping process

So far there have been no bootstrapping failures under from this setup, although the number of tests run have been very low.

In addition it was determined that adding more logging verbosity does not help with the debugging process, as the failing part of the bootstrapping is logged in full regardless.

### DIFF
--- a/.buildkite/expo-pipeline.yml
+++ b/.buildkite/expo-pipeline.yml
@@ -14,6 +14,7 @@ steps:
           push:
             - expo-publisher:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:expo-publisher-${BRANCH_NAME}
             - expo-publisher:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:expo-publisher-latest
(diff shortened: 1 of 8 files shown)


Cawllec:
The bootstrapping takes approximately 1minute 30seconds.  I don't believe this is a significant increase from the previous runs, or that it will overall effect the time taken.

Cawllec:
Ah, it broke again unfortunately.

Cawllec:
With concurrency set to 1 the last build took approx 4minutes, which seems like a significant slowdown.

bengourley:
When logging was set to verbose, did lerna say anything interesting before it failed?

Cawllec:
Nope, just listed the previous packages.  The actually error messages were identical.

Issue #668, opened 2019-12-10
Expo SDK 36 Error due to react-native-netinfo


vikrantnegi:
Hi,

I just upgraded my expo app to sdk 36 and I'm getting an error with `react-native-netinfo` import. I believe this can be related to this expo/expo#4764 

![Screenshot_2019-12-10-10-20-04-197_host exp exponent](https://user-images.githubusercontent.com/3294287/70496787-edcc6200-1b36-11ea-8e56-f4232b0de2ed.jpg)

rafamrs:
I also got this, and this is definitely related with https://github.com/expo/expo/issues/4764

mattdyoung:
@vikrantnegi  @rafamrs 

Thanks for the report. We're working on testing the fix and aim to have a release out by Monday.

rafamrs:
Thats great to hear @mattdyoung. Thanks for addressing the issue so quickly

vikrantnegi:
Thank you @mattdyoung I will be waiting for the fix to test it out on my app.

phillipsam:
@vikrantnegi @rafamrs 
released in [v6.5.0](https://github.com/bugsnag/bugsnag-js/releases/tag/v6.5.0)

PR #670, opened 2019-12-10
Expo v36 support

bengourley:
As of Expo 36, the `react-native` module no longer has `NetInfo` installed by default. This means `@bugsnag/expo` in its current state does not work with SDK 36 (#668).

This PR adds the `@react-native-community/netinfo` package as a dependency to `@bugsnag/expo`, and updates its usage (the api of the module has changed since it moved to the community org).

In the same way we supported SDK 34 and had to drop support for older versions (#610), this PR also updates `bugsnag-expo-cli` to install the relevant version of Bugsnag for the Expo SDK version detected.

### DIFF
--- a/packages/delivery-expo/network-status.js
+++ b/packages/delivery-expo/network-status.js
@@ -25,10 +25,10 @@ module.exports = class NetworkStatus {
   _watch () {
     // get the initial status
     NetInfo.fetch().then(state => {
-      this._update(state.isInternetReachable)
+      this._update(state.isConnected)
       // then listen for subsequent changes
       NetInfo.addEventListener(state => {
-        this._update(state.isInternetReachable)
+        this._update(state.isConnected)
       })
     })
   }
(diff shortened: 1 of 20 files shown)


packages/delivery-expo/network-status.js
```diff
@@ -24,11 +24,11 @@ module.exports = class NetworkStatus {
...
     // get the initial status
-    NetInfo.isConnected.fetch().then(isConnected => {
-      this._update(isConnected)
+    NetInfo.fetch().then(state => {
+      this._update(state.isInternetReachable)
       // then listen for subsequent changes
-      NetInfo.isConnected.addEventListener('connectionChange', isConnected => {
-        this._update(isConnected)
+      NetInfo.addEventListener(state => {
+        this._update(state.isInternetReachable)
```
tomlongridge: Is a change to use `isInternetReachable` (as opposed to `isConnected`) the correct state to check? For example, a device on a private network using an On Premise installation presumably wouldn't have internet connectivity but we'd still want to trigger a change of status.

bengourley: Hmm good point – I'll revert to using `isConnected`.

PR #673, opened 2019-12-16
V7: Remove client device prop

bengourley:
`client.device` should not exist – `device` now exists on `Event` only.

The session payload callbacks are now hooked up to get `device` info into the session payload.

### DIFF
--- a/packages/plugin-browser-session/session.js
+++ b/packages/plugin-browser-session/session.js
@@ -1,6 +1,5 @@
 const { includes } = require('@bugsnag/core/lib/es-utils')
 const inferReleaseStage = require('@bugsnag/core/lib/infer-release-stage')
-const runSyncCallbacks = require('@bugsnag/core/lib/sync-callback-runner')
 
 module.exports = {
   init: client => { client._sessionDelegate = sessionDelegate }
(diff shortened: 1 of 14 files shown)


packages/plugin-browser-session/session.js
```diff
@@ -30,8 +31,15 @@ const sessionDelegate = {
           user: sessionClient._user
         }
       ]
-    })
+    }
+
+    const ignore = runSyncCallbacks(sessionClient._cbs.sp.slice(0), payload, 'onSessionPayload', sessionClient._logger)
```
djskinner: Don't think `sessionClient._cbs.sp.slice(0)` is technically needed as `runSyncCallbacks` also copies the array with slice.

djskinner: `slice(0)` could also be replaced with `slice()` although neither change affects the bundle size AFAICT

bengourley: Good point, though I can't remember why the array is copied at all 🤔

bengourley: the sync callback runner copies the array and pops of functions until there are none left – that's why it needs to be copied.

PR #675, opened 2019-12-17
V7: Rename trackError -> _track and remove from public types

bengourley:
Sessions are now part of the public interface via `onSession` callbacks, so the public `trackError()` function has been suitably renamed (it takes events not errors) and made private.

### DIFF
--- a/packages/core/test/session.test.js
+++ b/packages/core/test/session.test.js
@@ -11,7 +11,7 @@ describe('@bugsnag/core/session', () => {
       expect(s.events).toEqual({ handled: 0, unhandled: 0 })
     })
   })
-  describe('track()', () => {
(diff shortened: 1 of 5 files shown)


bengourley:
I can't fathom why this change _increased_ the bundle size 🤔

packages/core/test/session.test.js
```diff
@@ -11,14 +11,14 @@ describe('@bugsnag/core/session', () => {
       expect(s.events).toEqual({ handled: 0, unhandled: 0 })
     })
   })
-  describe('trackError()', () => {
+  describe('track()', () => {
```
djskinner: `_track()`? No big deal really.

PR #677, opened 2019-12-17
V7: Remove client `app` property

bengourley:
`client.app` has been removed. Only `event.app` is available now.

This switches to using `onSession` and `onError` callbacks to add `app` data to payloads.

Since the `version`, `releaseStage` values can no longer be updated at runtime, logic that checked various places has been simplified to refer to the immutable value from config.

### DIFF
--- a/packages/plugin-expo-app/app.js
+++ b/packages/plugin-expo-app/app.js
@@ -25,6 +25,12 @@ module.exports = {
       }
     }
 
+    client.addOnSession(session => {
+      if (Constants.manifest.revisionId) {
+        session.app.codeBundleId = Constants.manifest.revisionId
+      }
+    })
+
     client.addOnError(event => {
       const now = new Date()
       const inForeground = AppState.currentState === 'active'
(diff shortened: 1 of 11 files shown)


packages/plugin-expo-app/app.js
```diff
@@ -25,6 +25,12 @@ module.exports = {
       }
     }
 
+    client.addOnSession(session => {
+      if (Constants.manifest.revisionId) {
```
djskinner: Whats the reason for some of this being done here and others in the expo notifier setup?

packages/plugin-expo-app/test/app.test.js
```diff
@@ -3,38 +3,7 @@
...
 
-describe('plugin: expo device', () => {
-  it('should should use version if defined (all platforms)', done => {
-    const VERSION = '1.0.0'
-    const plugin = proxyquire('../', {
-      'expo-constants': {
-        default: {
-          platform: {
-          },
-          manifest: { version: VERSION }
```
djskinner: I don't think there's any test coverage ensuring we capture this now

packages/plugin-expo-app/app.js
bengourley: `appVersion` got moved to the Expo notifier setup because it must be set in configuration and therefore supplied before the notifier starts.

`codeBundleId` is not a configuration option and since there is nowhere to put it on the client, it needs to be added to session/event payloads in callbacks.

packages/plugin-expo-app/test/app.test.js
bengourley: It's not covered by unit tests anymore but it is covered by this maze-runner step:

https://github.com/bugsnag/bugsnag-js/blob/9fcf7efbccd2cad74079ae89256c090c6c6ea640/test/expo/features/app.feature#L12
which asserts that the value is passed through from the fixture app's `app.json`:
https://github.com/bugsnag/bugsnag-js/blob/9fcf7efbccd2cad74079ae89256c090c6c6ea640/test/expo/features/fixtures/test-app/app.json#L11

PR #679, opened 2019-12-18
V7: Remove Bugsnag* prefix from internal classes

bengourley:
The only good reason for these prefixes was naming clashes, but since they are all module-scoped, it's safe to remove them. This means the class names exactly match the spec and the bundle size may be a little smaller.


packages/core/event.js
```diff
@@ -6,10 +6,10 @@ const jsRuntime = require('./lib/js-runtime')
 const metadataDelegate = require('./lib/metadata-delegate')
 const isError = require('./lib/iserror')
 
-class BugsnagEvent {
+class Event {
   constructor (errorClass, errorMessage, stacktrace = [], handledState = defaultHandledState(), originalError) {
     // duck-typing ftw >_<
-    this.__isBugsnagEvent = true
+    this.__isEvent = true
```
djskinner: Where is this used?

bengourley: Good shout! It's no longer used since #664.

PR #681, opened 2019-12-19
V7: Move context and breadcrumbs to private properties

bengourley:
Per the spec, context and breadcrumbs should be private.

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -19,21 +19,18 @@ class Client {
     this._config = {}
     this._schema = schema
 
-    // // i/o
+    // i/o
     this._delivery = { sendSession: noop, sendEvent: noop }
     this._logger = { debug: noop, info: noop, warn: noop, error: noop }
 
     // plugins
     this._plugins = {}
 
+    // state
+    this._breadcrumbs = []
     this._session = null
-
     this._metadata = {}
-
-    this.breadcrumbs = []
-
-    // setable props
-    this.context = undefined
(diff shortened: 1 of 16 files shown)


packages/core/config.js
```diff
@@ -88,9 +88,14 @@ module.exports.schema = {
...
   },
+  context: {
+    defaultValue: () => undefined,
+    message: 'should be a string',
+    validate: value => value === undefined || typeof value === 'string'
+  },
   user: {
     defaultValue: () => null,
-    message: '(object) user should be an object',
+    message: 'user should be an object',
```
djskinner: Could remove 'user' here? I guess the context would be sufficient to know it was user? Also makes it consistent with the other messages and may save a byte if we're lucky!

packages/core/client.js
```diff
@@ -82,6 +79,14 @@ class Client {
     return metadataDelegate.clear(this._metadata, section, key)
   }
 
+  getContext () {
+    return this._context
+  }
+
+  setContext (c) {
+    this._context = c
+  }
```
djskinner: Do these need adding to the public interface typedefs?

bengourley: yeah – I was in the middle of a big types refactor when I realised this change was needed, so they are in #682 to save a conflict!

packages/core/config.js
bengourley: 6 bytes after minification!

PR #682, opened 2019-12-20
V7: Improve type definitions

bengourley:
The PR improves the types in various ways

- does away with the hacky way config was extended for each platform
- makes `request`, `app` and `device` more specific, and reuses their definitions
- adds more rigourous tests, which are written in typescript – ensuring the types actually match the JS object at runtime


bengourley:
I was on the fence about the nesting anyway – I saw it as a good way to structure the miscellanaeous types that were not typically needed by the user but were exported just case. But I have just flattened and exported them all at the top level, which removes the nesting and the dubious name.

PR #684, opened 2019-12-20
Use jest for core package

djskinner:
- Adds basic jest setup
- Converts `packages/core` to use jest
- Adds basic linting plugins for test files

_Approach_

Use jest projects to group tests which share similar configurations. This is necessary as some packages require different environment e.g. browser/node/react-native. One downside is that new packages will likely need adding to the configuration. This could be likely be detected by a drop in code coverage reports.

_Code coverage_

Jest will generate code coverage for test files with `jest --coverage` with no further configuration. There may be additional config required to ensure files that are not required by tests are included in the coverage report but this should be relatively straightforward (slightly complicated for a mono-repo setup).

Once we have a single coverage report for the entire monorepo it will be relatively straightforward to configure danger to post a coverage report using `danger-plugin-istanbul-coverage`.

_Migration_

I have most of the packages now converted but there's not reason we can't roll this out in a series of smaller PRs starting with this one if the approach is agreeable.

_Speed_

There is initial setup for jest so it's taking 5-10 seconds on my local machine. However once the setup is done it's pretty quick. All the tests I have converted are running in less than 20 seconds which seems a lot quicker than the existing setup.

To Do:
- [ ] Fix conflicts, update against target branch


.eslintrc.js
```diff
@@ -31,6 +31,18 @@ module.exports = {
       rules: {
         '@typescript-eslint/no-explicit-any': 'off'
       }
+    },
+    // Linting for tests
+    {
+      files: [
+        'packages/core/**/*.test.[tj]s?(x)'
```
djskinner: Linting coverage will be expanded as we roll out jest. I separated the test fixes from the lint fixes in the commit history. If you look at the individually you can see the actual test changes required to make the tests work are minimal (in this case). Fixing lint errors and warnings creates a bigger diff but it mostly around using the safer return Promise approach for async tests, which is auto fixable.

babel.config.js
```diff
@@ -1,5 +1,7 @@
 module.exports = api => {
-  if (api) api.cache(false)
+  if (api && !api.env('test')) {
```
djskinner: Jest must set `api.cache(false)` itself and Babel doesn't like you calling it twice

jest.config.js
```diff
@@ -0,0 +1,12 @@
+const testsForPackage = (packageName) => `<rootDir>/packages/${packageName}/**/*.test.[jt]s`
+
+module.exports = {
+  projects: [
+    {
+      displayName: 'core packages',
+      testMatch: [
+        testsForPackage('core')
```
djskinner: Beyond the different different environment requirements for some tests (browser/node/react-native) the grouping is fairly arbitrary. We could add more groups for structure. This will shape up some more as we add more jest coverage.

package.json
```diff
@@ -42,7 +48,7 @@
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build --scope '@bugsnag/node' --scope '@bugsnag/browser' --scope '@bugsnag/expo' && lerna run build --ignore '@bugsnag/node' --ignore '@bugsnag/browser' --ignore '@bugsnag/expo'",
     "test:lint": "eslint . --report-unused-disable-directives",
-    "test:unit": "lerna run test --ignore '@bugsnag/browser' --ignore '@bugsnag/node'",
+    "test:unit": "jest && lerna run test --ignore '@bugsnag/browser' --ignore '@bugsnag/node'",
```
djskinner: Could do `jest --coverage` here. I haven't as intend to leave coverage till last.

packages/core/lib/test/async-every.test.js
```diff
@@ -3,53 +3,61 @@ const { describe, it, expect } = global
 const every = require('../async-every')
 
 describe('async-every', () => {
-  it('handles iterator errors', done => {
-    every([1], (item, cb) => cb(new Error('derp')), (err, result) => {
-      expect(err)
-      expect(err.message).toBe('derp')
-      done()
+  it('handles iterator errors', () => {
+    return new Promise(resolve => {
```
djskinner: the lint rules prefer returning a promise over the done callback approach - see https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-test-callback.md

packages/core/test/event.test.js
```diff
@@ -1,12 +1,13 @@
-const { describe, it, expect } = global
```
djskinner: Only this file required actual adjustments to the test setup. The rest are lint changes mostly converting async tests to the promise approach.

packages/core/lib/test/async-every.test.js
bengourley: Can we use `async`/`await` instead?

packages/core/test/client.test.js
```diff
@@ -27,45 +25,53 @@ describe('@bugsnag/core/client', () => {
...
+            warn: log,
+            error: log
+          }
+        })
+        client._logger.debug('hey')
       })
-      client._logger.debug('hey')
     })
+
+    // eslint-disable-next-line jest/expect-expect
```
bengourley: I think it would be better to add an assertion to this test rather than disable the lint error

package.json
```diff
@@ -14,8 +14,12 @@
     "@babel/plugin-transform-shorthand-properties": "^7.0.0",
     "@babel/plugin-transform-spread": "^7.0.0",
     "@babel/plugin-transform-template-literals": "^7.0.0",
+    "@babel/preset-env": "^7.7.5",
+    "@babel/preset-react": "^7.7.4",
+    "@babel/preset-typescript": "^7.7.4",
     "@typescript-eslint/eslint-plugin": "^2.5.0",
     "@typescript-eslint/parser": "^2.5.0",
+    "babel-jest": "^24.9.0",
```
bengourley: Is it necessary to add all these babel modules? If we're running the Jest tests on a recent version of Node, all of the syntax should be valid, right?

jest.config.js
```diff
@@ -0,0 +1,12 @@
+const testsForPackage = (packageName) => `<rootDir>/packages/${packageName}/**/*.test.[jt]s`
+
+module.exports = {
+  projects: [
+    {
+      displayName: 'core packages',
```
bengourley: I would call this `core package` or just `core`.

babel.config.js
bengourley: It would be really good if we could remove the complexity from this file. I had to look up `api.cache(false)` to see what the point of it was!

I made another comment about whether we need the tests to be compiled with babel at all – if we didn't, then we wouldn't need any logic in this file, and we could just turn it into a static `module.exports = { ... }` declaration, removing the `api.cache()` thing entirely.

bengourley:
Oh, also I think you'll need to add the jest config file to the docker images for ci (for example, [here](https://github.com/bugsnag/bugsnag-js/blob/next/dockerfiles/Dockerfile.ci#L11)), otherwise it looks like jest tries to run all the unconverted tests too: https://buildkite.com/bugsnag/at-bugsnag-js/builds/1234#300ac8d7-74e3-4d89-acf0-e9ef42aa5fbb

packages/core/lib/test/async-every.test.js
djskinner: That doesn't look to be supported by this eslint rule. I'm going to disable this rule for now and we can come back to it later.

package.json
djskinner: I've removed babel-jest etc for now. They will need adding back later when we introduce TypeScript.

packages/browser/types/test/types.test.js
```diff
@@ -1,16 +1,12 @@
...
-
-afterAll(() => {
-  cleanupFiles.forEach(f => spawnSync('rm', ['-f', `${__dirname}/fixtures/${f}`]))
-})
 
 const assertTsProgramCompiles = p => {
-  cleanupFiles.push(`${p}.js`)
   const { stdout, stderr, err } = spawnSync('./node_modules/.bin/tsc', [
     '--strict',
+    '--noEmit',
```
djskinner: The [`--noEmit` flag](https://www.typescriptlang.org/docs/handbook/compiler-options.html) can be used to avoid having to emit files and clean up manually

packages/browser/types/test/types.test.js
```diff
@@ -1,16 +1,12 @@
...
+    '--skipLibCheck',
```
djskinner: When adding jest to the monorepo a number of @types packages were updated which include new TS functionality such as `unknown`. One way to get around it is to not type check node_modules using `--skipLibCheck`. Alternatively updating to TypeScript > 3 fixes the issues. Let me know which approach you prefer @bengourley. Of the two I think bumping to the latest TypeScript is best but I can see an argument for doing both as there's no need to type check node_modules for these particular tests.

packages/core/package.json
```diff
@@ -18,8 +18,7 @@
     "*.js"
   ],
   "scripts": {
-    "test": "nyc --reporter=lcov -- jasmine '!(node_modules)/**/*.test.js'",
-    "test:types": "tsc -p types/test/tsconfig.json && jasmine types/test/test.js"
+    "test:types": "tsc -p types/test/tsconfig.json && jest types/test/test.js"
```
djskinner: Starting jest for a single test file like this is not ideal due to boot up time. I've opted to keep the approach as similar to the existing as possible however, so `npm run test:types` still remains a separate function. This is in the knowledge that it will only be a temporary thing until we add TypeScript.

packages/core/types/test/test.js
```diff
@@ -1,96 +0,0 @@
-"use strict";
```
djskinner: This is an auto-generated (by tsc for testing), `.gitignore`d file so I removed it from source control.

packages/core/types/test/tsconfig.json
```diff
@@ -55,5 +55,6 @@
     /* Experimental Options */
     // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+    "skipLibCheck": true
```
djskinner: Same here. We could bump TypeScript instead.

packages/browser/types/test/types.test.js
bengourley: I think we should update to TypeScript v3

.eslintrc.js
```diff
@@ -1,11 +1,18 @@
-// eslint config for .js files
```
djskinner: We need to work out what linting overrides we're going to want for TypeScript files particularly as it related to formatting, such as semi-colons. The current `.d.ts` definition files use semi-colons but there is no enforced linting preference. Presumably we'll want to insist on one way or the other as we do for the JS file.

babel.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = api => {
-  if (api) api.cache(false)
```
djskinner: The complexity of this file is back now that babel-jest and TypeScript have been added back. Not really sure there's much we can do about this though. See https://jestjs.io/docs/en/getting-started#using-typescript

packages/core/lib/async-every.d.ts
```diff
@@ -0,0 +1,7 @@
+type NodeCallbackType<T = any> = (error?: Error | null, result?: T) => void;
```
djskinner: [TypeScripts looks for sibling type declaration files for relative imports](https://www.typescriptlang.org/docs/handbook/module-resolution.html#module-resolution-strategies). Whilst TypeScript will infer types from JS files when `allowJs` is true it's not often sufficient. Without these files many test cases need extra type decoration. In the case of lib files as here I think this is a good intermediate step towards full TypeScript conversion.

packages/core/lib/test/async-every.test.ts
```diff
@@ -1,12 +1,10 @@
...
+import every from '../async-every'
 
 describe('async-every', () => {
   it('handles iterator errors', done => {
-    every([1], (item, cb) => cb(new Error('derp')), (err, result) => {
-      expect(err)
-      expect(err.message).toBe('derp')
+    every([1], (item, cb) => cb(new Error('derp')), (err) => {
+      expect(err).toBeTruthy()
+      expect(err!.message).toBe('derp')
```
djskinner: This non-null assertion operator is currently triggering a lint warning:

```
> eslint --ext .ts,.js--report-unused-disable-directives .


/Users/dan/bugsnag-js/packages/core/lib/test/async-every.test.ts
  7:14  warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion

✖ 1 problem (0 errors, 1 warning)
```

I think we should avoid letting warnings creep in with `--max-warnings=0` and decided whether this rule is globally disabled in test files or ignored on a per-usage basis.

packages/core/test/client.test.js
```diff
@@ -1,5 +1,4 @@
-const { describe, it, expect, fail, spyOn } = global
```
djskinner: Converting this file to TS is a little more tricky.  When using a relative import of a JS file such as `'../client'`, TS looks for type declarations in `../client.d.ts`. Since it doesn't exist it infers it's own from the JS file itself. This is largely insufficient and these tests would have a lot of forced casting.

On the flip side we also want to test out the manually maintained definitions. This can be achieved by importing from the root with `import { Client } from '..'`. The problem is that the public definitions do not contain all the internal types that of often manipulated in this tests e.g. (`client._sessionDelegate = sessionDelegate`) so again some casting would be needed.

Needs some more though on how we're going to handle this one.

packages/core/test/event.test.ts
```diff
@@ -26,39 +27,34 @@ describe('@bugsnag/core/event', () => {
 
   describe('addMetadata()', () => {
     it('updates a whole new section', () => {
-      const Event = require('../event')
       const r = new Event('Err', 'bad', [])
       r.addMetadata('specific detail', { extra: 'stuff' })
-      expect(r._metadata['specific detail']).toEqual({ extra: 'stuff' })
+      expect((r._metadata as any)['specific detail']).toEqual({ extra: 'stuff' })
```
djskinner: Here's an example where the types inferred by TS from the JS file are insufficient. In this case it infers _metadata to have type `{}` so casting has been used to hide the type error. A better way might be to have a sibling `event.d.ts` file that properly defines the types so that we don't lose the benefits of static type checking. The problem is how to reconcile that with the other `event.d.ts` that represents only the API we wish consumers to use.

packages/core/test/session.test.ts
```diff
@@ -1,6 +1,4 @@
-const { describe, it, expect } = global
-
-const Session = require('../session')
+import Session from '../session'
```
djskinner: This is an example where the JS infer has worked in our favour. `toJSON()` and `_track` are not defined in the manual type definitions but have been well-inferred from the JS and so we have no type issues.

packages/core/test/types.test.ts
```diff
@@ -1,5 +1,4 @@
-import Bugsnag, { Client, Config } from '../..'
-import "jasmine"
+import Bugsnag, { Client, Config } from '..'
```
djskinner: These are moved for now. We may wish to keep them to test against the public manually maintained type definitions whilst separate internal types are used for testing internals.

packages/core/types/common.d.ts
```diff
@@ -129,5 +129,5 @@ interface Request {
 export interface User {
   id?: string;
   email?: string;
-  name?: string
+  name?: string;
```
djskinner: We probably need to work out whether we're going with semi-colons or not for ts files and enforce it with a rule. Currently, as in this case, there is no linting rule, in contrast with the JS files.

tsconfig.json
```diff
@@ -55,5 +61,8 @@
     /* Experimental Options */
     // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
-  }
+  },
+  "include": [
```
djskinner: Ideally this would be temporary, possibly convert to an exclude to rule out the things we don't want to type test rather than the current whitelist approach.

.eslintrc.js
bengourley: Can we use standard on ts files too, for consistency? I found this https://github.com/standard/eslint-config-standard-with-typescript

babel.config.js
bengourley: Yeah – necessary evil now. Never mind!

packages/core/lib/test/async-every.test.ts
bengourley: > I think we should avoid letting warnings creep in with --max-warnings=0

I'm happy with that rule :+1:

What do you think about this lint rule? It seems somewhat overzealous for tests. The trouble is the compiler doesn't know `.toBeTruthy()` will bail when `err` is `null` so it needs convincing. Is there a better construct to use on the `err` in this test?

The refactor seems pretty straightforward otherwise:

```ts
expect(err).toBeTruthy()
if (err) expect(err!.message).toBe('derp')
```

packages/core/test/client.test.js
bengourley: I would be in favour of importing the public types (so they get fully tested) and then writing ad-hoc types for the private internals in the test file. Does that sound ok to you?

.eslintrc.js
djskinner: I took a quick look. I think it can be done if we disable the rules requiring parser services (for now at least). I'm going to split some of this out into separate PRs as I feel like there's too much going on here now.

djskinner: First one: https://github.com/bugsnag/bugsnag-js/pull/711

djskinner: Second one: https://github.com/bugsnag/bugsnag-js/pull/712

packages/core/test/client.test.ts
```diff
@@ -1,13 +1,14 @@
...
-const Session = require('../session')
+/* eslint-disable @typescript-eslint/no-empty-function */
+import Client from '../Client'
+import Event from '../Event'
+import { Session } from '..'
 
 describe('@bugsnag/core/client', () => {
   describe('constructor', () => {
     it('can handle bad input', () => {
+      // @ts-ignore
```
djskinner: We need to disable type checking for tests that deliberately test for bad input

packages/core/test/client.test.ts
```diff
@@ -240,13 +252,14 @@ describe('@bugsnag/core/client', () => {
...
     it('leaves a breadcrumb of the error', () => {
-      const payloads = []
+      const payloads: any[] = []
       const client = new Client({ apiKey: 'API_KEY_YEAH' })
       client._setDelivery(client => ({ sendEvent: (payload) => payloads.push(payload) }))
       client.notify(new Error('foobar'))
       expect(client._breadcrumbs.length).toBe(1)
       expect(client._breadcrumbs[0].type).toBe('error')
       expect(client._breadcrumbs[0].message).toBe('Error')
+      // @ts-ignore
```
djskinner: This one might actually point to a shortcoming in the Breadcrumb typings, or the data used in the test needs correcting. I wasn't sure, what do you think @bengourley ?

packages/core/types/client.d.ts
```diff
@@ -13,7 +13,7 @@ import Event from './event'
 import Session from './session'
 
 declare class Client {
-  private constructor();
+  protected constructor();
```
djskinner: I'm not sure we need this declaration at all - we don't have it on the Event public types.

packages/core/test/client.test.ts
bengourley: Yeah looks to be the case. Previously breadcrumb metadata was only allowed to be 1 level deep – now it can be of arbitrary depth. Let's leave ignore in this PR but fix this in another PR.

packages/core/types/client.d.ts
bengourley: It was to deter people from instantiating the class directly but yeah… I'm not sure if we need it.

PR #685, opened 2019-12-23
V7: static interface

bengourley:
This PR adds a static interface to the notifier, meaning that developers using Bugsnag don't have to hold on to a reference to `Client` in their applications (though they can if they want to).

### DIFF
--- a/packages/browser/src/notifier.js
+++ b/packages/browser/src/notifier.js
@@ -7,6 +7,8 @@ const Event = require('@bugsnag/core/event')
 const Session = require('@bugsnag/core/session')
 const Breadcrumb = require('@bugsnag/core/breadcrumb')
 
+const { reduce, keys } = require('@bugsnag/core/lib/es-utils')
+
 // extend the base config schema with some browser-specific options
 const schema = { ...require('@bugsnag/core/config').schema, ...require('./config') }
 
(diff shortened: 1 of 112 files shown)


packages/browser/src/notifier.js
```diff
@@ -29,49 +31,69 @@ const pluginStripQueryString = require('@bugsnag/plugin-strip-query-string')
...
+      Bugsnag._client._logger.warn('Bugsnag.init() was called more than once. Ignoring.')
+      return Bugsnag._client
+    }
+    Bugsnag._client = Bugsnag.createClient(opts)
+    Bugsnag._client._depth += 1
+    return Bugsnag._client
+  }
 }
 
+reduce(keys(Client.prototype), (accum, m) => {
```
djskinner: This might be worthy of a brief comment to aid understanding. Since `Bugsnag` is being mutated and the final accumulated value is never used, it is behaving more like `forEach` than `reduce`.

An alternative to consider:

```
map(keys(Client.prototype), (m) => {
  if (/^_/.test(m)) return
  Bugsnag[m] = function () {
    if (!Bugsnag._client) return console.log(`Bugsnag.${m}() was called before Bugsnag.init()`)
    return Bugsnag._client[m].apply(Bugsnag._client, arguments)
  }
})
```

packages/core/types/bugsnag.d.ts
```diff
@@ -0,0 +1,66 @@
...
+  Config,
+  NotifiableError,
+  OnErrorCallback,
+  OnSessionCallback,
+  OnBreadcrumbCallback,
+  User,
+  Plugin
+} from "./common";
+
+export default interface BugsnagStatic {
```
djskinner: Consider using `export default interface BugsnagStatic extends Client` to save having to restate the `Client` type signature.

PR #687, opened 2019-12-28
fix params

marktran:
`getRequestAndMetaDataFromReq` is incorrectly nested inside `createReportFromErr`.

Koa plugin: https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-koa/src/koa.js#L73


mattdyoung:
Hi @marktran 

Thanks for the PR! We'll look at getting this reviewed and merged in due course.

bengourley:
I've reopened this in #695 so that our authenticated checks will run.

PR #692, opened 2020-01-03
V7: Ensure session interface matches spec

bengourley:
The session class and usage required a bit of tweaking now that the spec has settled down.

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -139,10 +139,14 @@ class Client {
     const session = new Session()
 
     session.app = {
-      releaseStage: this._config.releaseStage,
-      version: this._config.appVersion,
-      type: this._config.appType
+      ...{
+        releaseStage: this._config.releaseStage,
+        version: this._config.appVersion,
+        type: this._config.appType
+      }
     }
+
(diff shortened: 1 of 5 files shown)


packages/core/client.js
```diff
@@ -139,11 +139,15 @@ class Client {
     const session = new Session()
 
     session.app = {
-      releaseStage: this._config.releaseStage,
-      version: this._config.appVersion,
-      type: this._config.appType
+      ...{
```
djskinner: What does this creation and spreading of an object literal achieve?

packages/core/client.js
```diff
@@ -139,11 +139,15 @@ class Client {
...
+        releaseStage: this._config.releaseStage,
+        version: this._config.appVersion,
+        type: this._config.appType
+      }
     }
 
+    session.user = { ...this._user }
```
djskinner: Should this be `session._user`?

bengourley: yes!

packages/core/client.js
bengourley: ah this a mistake, I'll rework it

PR #696, opened 2020-01-06
feat(node): Use util.inspect() before printing non-errors

bengourley:
`throw { ... }` or `reject({ ... })` is bad practice, but sometimes it happens. Previously (as
highlighted in #597), Bugsnag would just print `[object Object]`, which is unhelpful. This
uses util.inspect() to get a better textual representation of anything that doesn't have
a .stack property, before logging it to the console.

Partially fixes #597 (work on the `v7` branch addresses the part that affects the error report).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


bengourley:
Rebased against `next` and added changelog.

PR #709, opened 2020-01-23
feat(delivery-react-native): Build specified payload to send over bridge

bengourley:
This update takes the prototyped `@bugsnag/delivery-react-native` component and makes it ready to use:

- Added a readme
- Depend on local version of `@bugsnag/core` (from the local monorepo rather than fetching an old version from npm)
- Update the payload that gets transferred to the native layer
- Update the unit test based on the new implementation

The unit tests will not pass for this because of some other plugins that are beyond the scope of this PR.

To test that the values are crossing the bridge as expected, you can drill down to `BugsnagReactNative.java` in the node modules of the RN project you are testing on and add the following:

```java
    import android.util.Log;

    // ... snip    

    @ReactMethod
    private void dispatch(ReadableMap payload, Promise promise) {
        Log.i("Bugsnag", payload.getArray("errors").getMap(0).getString("errorClass"));
        Log.i("Bugsnag", payload.getArray("errors").getMap(0).getString("errorMessage"));
        Log.i("Bugsnag", payload.getString("severity"));
        //...
    }
```

(full path `node_modules/@bugsnag/react-native/android/src/main/java/com/bugsnag/reactnative/BugsnagReactNative.java`)

Then you can filter logcat just to the `Bugsnag` logs like so `adb logcat Bugsnag "*:S"`.

### DIFF
--- a/packages/delivery-react-native/delivery.js
+++ b/packages/delivery-react-native/delivery.js
@@ -16,5 +16,7 @@ module.exports = (client, NativeClient) => ({
       groupingHash: event.groupingHash
     }).then(() => cb()).catch(cb)
   },
-  sendSession: () => {}
+  sendSession: () => {
+    client._logger.warn('@bugsnag/delivery-react-native sendSession() should never be called', new Error().stack)
+  }
 })
--- a/packages/delivery-react-native/delivery.js
+++ b/packages/delivery-react-native/delivery.js
@@ -1,11 +1,20 @@
 module.exports = (client, NativeClient) => ({
   sendEvent: (payload, cb = () => {}) => {
     const event = payload.events[0]
-    // this is because JS beforeSend operates on the event – report is not in scope
-    // report.threads = event.get('threads')
-    NativeClient.dispatch(event.toJSON()).then(() => cb()).catch(cb)
+    NativeClient.dispatch({
+      errors: event.errors,
+      severity: event.severity,
+      severityReason: event._handledState.severityReason,
+      unhandled: event.unhandled,
+      app: event.app,
+      device: event.device,
+      threads: event.threads,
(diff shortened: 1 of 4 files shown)


packages/delivery-react-native/delivery.js
```diff
@@ -1,11 +1,20 @@
...
+      context: event.context,
+      user: event._user,
+      metadata: event._metadata,
+      groupingHash: event.groupingHash
+    }).then(() => cb()).catch(cb)
   },
-  sendSession: () => {
-    // TODO: log/warn here? If this gets used the something has gone wrong…
-  }
+  sendSession: () => {}
```
fractalwrench: Going off the previous comments something it appears that this isn't expected to be called - it'd be good to add some sort of warning message here if this does occur.

packages/delivery-react-native/test/delivery.test.js
```diff
@@ -1,25 +1,41 @@
...
-      expect(report.get('errorMessage')).toBe('123')
-      expect(report.get('threads')).toEqual(threads)
+      expect(sent.length).toBe(1)
+      expect(sent[0].errors[0].errorMessage).toBe('oh no')
+      expect(sent[0].severity).toBe('warning')
+      expect(sent[0].severityReason.type).toBe('handledException')
+      expect(sent[0].unhandled).toBe(false)
+      expect(sent[0].app).toEqual({ releaseStage: 'production', version: null, type: null })
+      expect(sent[0].device).toEqual({})
+      // TODO enable once event.threads exists
```
fractalwrench: assuming that this will be addressed as part of a later PR

bengourley: Yup, we need too add `threads` to JS's `event` interface but I'll do that in a different PR.

PR #710, opened 2020-01-23
feat(plugin-react-native-client-sync): Maintain client state between JS<->Native

bengourley:
This plugin is responsible for communicating state changes from JS -> Native and from Native -> JS. The following data is considered state that needs to be synchronised:

- breadcrumbs
- user
- context
- metadata

From JS -> Native, the changes can be propagated directly via methods on the native `BugsnagReactNative` class that is surfaced to JS.

From Native -> JS, the changes must be delivered via events and the implementation for this differs slightly based on the native platform.

The unit test CI job will not pass for this PR because of some other RN plugins that are beyond the scope of this PR. The unit tests for the client sync plugin do pass.

### DIFF
--- a/packages/plugin-react-native-client-sync/README.md
+++ b/packages/plugin-react-native-client-sync/README.md
@@ -1,0 +1,6 @@
+# @bugsnag/plugin-react-native-client-sync
+
+This plugin maintains client state between the JS client and the native client in React Native. It is included in the React Native notifier.
+
+## License
+MIT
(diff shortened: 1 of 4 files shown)


bengourley:
To test out the Native -> JS sync I used dropped in the following snippets into the Android `BugsnagReactNative` class:

```java
import com.facebook.react.modules.core.DeviceEventManagerModule;
import com.facebook.react.bridge.Arguments;

DeviceEventManagerModule.RCTDeviceEventEmitter emitter = reactContext.getJSModule(
  DeviceEventManagerModule.RCTDeviceEventEmitter.class
);

WritableMap params = Arguments.createMap();
params.putString("type", "UserUpdate");
WritableMap user = Arguments.createMap();
user.putString("name", "Ben");
user.putString("email", "ben.gourley@bugsnag.com");
user.putString("id", "123456");
params.putMap("data", user);
emitter.emit("bugsnag::sync", params);
```

PR #713, opened 2020-01-25
chore (build): bump to docker-compose-buildkite-plugin v.3.1.0

twometresteve:
Update to docker-compose-buildkite-plugin v.3.1.0 in preparation for Buildkite agent v4.3.5

Node 4 and 6 test initially failed due to a change in how the plugin handles the step `command.  Fixed by changing to a list of parameters and moving beneath the plugin config.


bengourley:
Looks good but @twometresteve, can you switch the target to be `next` (that should have been the default)?

PR #715, opened 2020-01-28
test: simplify types tests with --noEmit

djskinner:
A small change to simplify test setup and teardown of the types tests


packages/core/types/test/test.js
```diff
@@ -1,96 +0,0 @@
-"use strict";
```
djskinner: This is a .gitignored file that can be removed as it is generated when needed during testing.

PR #716, opened 2020-01-28
feat(plugin-react-native-event-sync): Sync the appropriate fields from the native notifier

bengourley:
The purpose of this plugin is to fetch info from the native notifier that a developer might want to inspect or change in a JS `onError` callback.

The changes to support this are as follows:

- Add `threads` to the event interface (previously not implemented in JS for obvious reasons)
- Retrieve payload info from `NativeClient.getPayloadInfo()` in an async `onError` callback which gets added to the front of the queue (with `addOnError(cb, true)`)
- Merge `app` and `device` in case anything exists already (it shouldn't)
- Take `threads` and `breadcrumbs` as-is (nothing should ever exist there in JS yet)

This should be the last of the RN plugin PRs before I can get the unit tests all passing.


bengourley:
You can test what data the JS notifier is able to retrieve from the native notifier by adding an `onError` callback:

```js
Bugsnag.start({
  onError: event => console.log(event)
})
```

Then inspecting the logcat output.

Note that the native notifier doesn't actually respond with any data for `getPayloadInfo()` yet, so you will need to make it do that. For example, I did this to simulate device:

```java
import com.facebook.react.bridge.WritableMap;
import com.facebook.react.bridge.WritableNativeMap;
import com.facebook.react.bridge.Arguments;

WritableMap params = Arguments.createMap();
WritableMap device = Arguments.createMap();
device.putString("id", "1066");
params.putMap("device", device);
```

PR #718, opened 2020-01-29
feat(react-native): Remove unnecessary plugins

bengourley:
The following plugins (which were built for Expo) implement functionality that already exists in the underlying native notifiers:

- `@bugsnag/plugin-react-native-connectivity-breadcrumbs`
- `@bugsnag/plugin-react-native-orientation-breadcrumbs`
- `@bugsnag/plugin-react-native-app-state-breadcrumbs`

They are therefore not needed as part of the React Native notifier, as they would result in duplicate breadcrumbs.

### DIFF
--- a/packages/react-native/android/build.gradle
+++ b/packages/react-native/android/build.gradle
@@ -35,7 +35,7 @@ android {
 }
 
 dependencies {
-    api 'com.bugsnag:bugsnag-android:4.22.1001'
+    api 'com.bugsnag:bugsnag-android:4.22.1002'
     implementation 'com.facebook.react:react-native:+'
 
     androidTestImplementation "junit:junit:4.12"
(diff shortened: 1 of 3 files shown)


bengourley:
@fractalwrench and @robinmacharg I've tagged you for review but all I really need is confirmation that the following breadcrumbs are collected automatically by `bugsnag-android` and `bugsnag-cocoa`:

- device orientation changes (i.e it is rotated)
- network status changes (e.g. connecting/disconnecting from wifi)
- when the app enters the foreground/background

fractalwrench:
These breadcrumbs are not collected consistently at the moment on Android, but there is a ticket to address this. If it's needed for the React Native work we could prioritise this change.

bengourley:
What do you mean by `not collected consistently` ?

fractalwrench:
I've checked the backlog and it looks like none of these breadcrumbs are not implemented on Android currently. The original intention was to implement them later as they'd been an additive change, but we can add them as part of this work if needed.

robinmacharg:
The short answer is Yes-ish.  Nearly.

> * device orientation changes (i.e it is rotated)
> * when the app enters the foreground/background

Yes to both.  I've eyebal;led the code and seen events hit the dashboard.

> * network status changes (e.g. connecting/disconnecting from wifi)

[This is a PR waiting to merge.](https://github.com/bugsnag/bugsnag-cocoa/pull/448)

djskinner:
And assuming the unit test failures are expected.

PR #720, opened 2020-01-30
feat(react-native): Implement Configuration.load() interface, and allow selective JS config

bengourley:
- Implements a function with loads configuration from the native layer (via `BugsnagReactNative.configure()`
- Exposes that method at `Configuration.load()`
- Only allows a strict subset of options to be supplied in JS
- Propagates setting of state (user, context, metadata) down to native layer, if supplied in JS

NB. required lots of fiddling with the Jest/Babel/Types setup to get tests working.


babel.config.js
```diff
@@ -2,39 +2,34 @@ module.exports = api => {
...
-        '@babel/preset-env',
-        {
-          targets: {
-            node: 'current'
-          }
-        }
-      ],
-      '@babel/preset-typescript'
+      '@babel/preset-typescript',
+      'module:metro-react-native-babel-preset'
```
bengourley: This config is a bit weird – this means we now basically rely on this preset (`module:metro-react-native-babel-preset`) to compile all of our non-react native tests too. I couldn't figure out a way to do:

```
if react native tests, load these plugins/presets
else if browser/node tests, load these plugins/presets
else load these plugin presets // <- this is what actually gets used for the production bundle
``

PR #727, opened 2020-02-04
feat(browser): Integrate maxEvents functionality with new interface

bengourley:
- Ensure counts can be reset via the static interface as well as the client instance
- Rename `refresh`->`resetEventCount`
- Add method to type definitions

### DIFF
--- a/packages/core/types/client.d.ts
+++ b/packages/core/types/client.d.ts
@@ -68,6 +68,9 @@ declare class Client {
   public use(plugin: Plugin, ...args: any[]): Client;
   public getPlugin(name: string): any;
 
+  // implemented on the browser notifier only
+  public resetEventCount?(): void;
+
   // access to internal classes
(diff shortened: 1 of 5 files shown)


packages/core/types/client.d.ts
```diff
@@ -68,6 +68,9 @@ declare class Client {
   public use(plugin: Plugin, ...args: any[]): Client;
   public getPlugin(name: string): any;
 
+  // implemented on the browser notifier only
+  public resetEventCount?(): void;
```
bengourley: This seemed like the most reasonable way to implement this. If I did something like `class BrowserClient extends  Client` to add the method only for the browser, it upsets the "universal" interface. Specifically it upset the Angular plugin which does `import { Client } from '@bugsnag/js'` – it doesn't know if `Client` should be a `Client` or `BrowserClient` since it can be run in Node or Browser.

PR #728, opened 2020-02-04
refactor: Keep timestamps as Date objects until serialisation

bengourley:
As per the updated spec, ensures dates are instances of `Date` until they are serialised.

JSON serialises dates to `ISO8601` anyway, so anywhere dates are included in the body of the payload, the iso conversion now happens implicitly.

Where dates are used in headers, the conversion has to happen manually, but I found a terser way of doing this in IE8 (`JSON.parse(JSON.stringify(new Date()))` which meant we could ditch the `isoString()` function in `es-utils`. Hopefully this shaves a byte or two 🤞.

### DIFF
--- a/packages/delivery-x-domain-request/delivery.js
+++ b/packages/delivery-x-domain-request/delivery.js
@@ -1,5 +1,4 @@
 const payload = require('@bugsnag/core/lib/json-payload')
-const { isoDate } = require('@bugsnag/core/lib/es-utils')
 
 module.exports = (client, win = window) => ({
   sendEvent: (event, cb = () => {}) => {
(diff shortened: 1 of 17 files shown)


packages/delivery-x-domain-request/delivery.js
```diff
@@ -36,8 +35,11 @@ module.exports = (client, win = window) => ({
   }
 })
 
-const getApiUrl = (config, endpoint, version, win) =>
-  `${matchPageProtocol(config.endpoints[endpoint], win.location.protocol)}?apiKey=${encodeURIComponent(config.apiKey)}&payloadVersion=${version}&sentAt=${encodeURIComponent(isoDate())}`
+const getApiUrl = (config, endpoint, version, win) => {
+  const isoDate = JSON.parse(JSON.stringify(new Date()))
```
tomlongridge: Is it worth citing/explaining this for future reference?

PR #729, opened 2020-02-04
Enable better BrowserStack test naming

Cawllec:
Enable various Buildkite environment variables to be passed through to maze-runner which will help with linking particular Buildkite jobs to BrowserStack runs.
In this case, only the app-automate service will have the new names, which is currently only utilised by the Expo maze tests.


bengourley:
Can you point this at `next` please? That should have been the default branch.

bengourley:
Going to close this off as it's quite stale now.

PR #734, opened 2020-02-07
test: use jest on plugin-react-native-app-state-breadcrumbs

djskinner:


babel.config.js
```diff
@@ -34,7 +35,14 @@ module.exports = api => {
       ],
       '@babel/preset-typescript'
     )
+    plugins.unshift(
```
djskinner: This plugin seems to want to go to the top. It's likely needed for all the react-native plugins. We could create an override for this too although the glob patterns may be more complicated.

babel.config.js
```diff
@@ -34,7 +35,14 @@ module.exports = api => {
...
+      ['@babel/plugin-proposal-class-properties', { loose: true }]
+    )
+    overrides.push({
+      test: './node_modules/react-native/**/*',
+      presets: ['module:metro-react-native-babel-preset']
```
djskinner: Use overrides to only apply presets where they are needed

packages/plugin-react-native-app-state-breadcrumbs/test/app-state.test.ts
```diff
@@ -1,25 +1,29 @@
-const { describe, it, expect } = global
+import Client from '@bugsnag/core/client'
+import plugin from '../'
 
-const proxyquire = require('proxyquire').noCallThru().noPreserveCache()
+// @ts-ignore
```
djskinner: Still working this issue out - see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311

packages/plugin-react-native-app-state-breadcrumbs/test/app-state.test.ts
```diff
@@ -36,53 +40,24 @@ describe('plugin: react native app state breadcrumbs', () => {
...
-    const AppState = {
-      addEventListener: (type, fn) => {
-        _cb = fn
-      }
-    }
-    const plugin = proxyquire('../', {
-      'react-native': { AppState }
-    })
-
+    // @ts-ignore
```
djskinner: Apparently `null` is not a valid type for `enabledBreadcrumbTypes` but setting to `undefined` causes the test to fail. Needs further investigation.

tsconfig.json
```diff
@@ -44,10 +44,7 @@
     //   "node_modules/@types",
     //   "packages/core/types"
     // ],
-    "types": [                                /* Type declaration files to be included in compilation. */
```
djskinner: Better to just install the `@types` packages and let TS pick them up automatically.

tsconfig.json
```diff
@@ -64,5 +61,6 @@
   },
   "include": [
     "packages/core",
+    "packages/plugin-react-native-app-state-breadcrumbs",
```
djskinner: This is one we are going to have to remember to keep up-to-date. Will look at better ways of handling this.

packages/plugin-react-native-app-state-breadcrumbs/test/app-state.test.ts
bengourley: the types are wrong – according to the spec `null` is an allowed value. Let's not hold up this PR because of it. I'll make a note to fix it elsewhere.

PR #738, opened 2020-02-10
Fix expo publish: Add top level node modules to watched folders

Cawllec:
This fixes the issue with metro bundler being unable to hash the required to-level module `react-native`.

This fix may potentially be problematic, as it may introduce directories and modules into the expo test fixture that we may not want included.


Cawllec:
I've changed the way we're trying to exclude the top-level node modules from the build.  Instead of allowing the metro bundler to find them and hope it ignores them, I'm just deleting them before the publish as they're actually unnecessary after the initial `bootstrap/build` phase is complete.

PR #739, opened 2020-02-11
Setup react-native android layer to use snapshot, add unit test coverage

fractalwrench:
## Goal

Makes the RN Android layer use the latest available snapshot rather than having to rely on mavenLocal.

Additionally, this adds some JVM unit test coverage for methods which have already been implemented on `BugsnagReactNative`.

## Changeset
- Altered the example app and react-native package to use the maven snapshot repository + 5.0.0-rc01 dependency, rather than assembling a local artefact
- Added kotlin-stdlib gradle dependency. This is already added to the final artefact by bugsnag-android, but adding it via gradle allows us to write kotlin code
- Setup mockito for [mocking Kotlin classes](https://stackoverflow.com/questions/14292863/how-to-mock-a-final-class-with-mockito), which are final by default
- Added a unit test to verify that `BugsnagReactNative` forwards on the session tracking calls to the appropriate method on `Client`
- Converted `BugsnagReactNative` to use Kotlin
- Implemented `BugsnagReactNative#setUser()`

## Tests
Ran an example app and confirmed that a JS exception can be delivered to the dashboard.


packages/react-native/CONTRIBUTING.md
```diff
@@ -36,19 +36,7 @@ registry=http://localhost:4873
...
-git fetch
-git checkout configuration-immutability-refactor
-
-# increment the version in gradle.properties to something arbitrary and high
-
-./gradlew assembleRelease publishToMavenLocal
-```
-
-Then update the version of `bugsnag-android` in `packages/react-native/android/build.gradle` to the arbitrary version you selected. Any time new Android changes are pulled down you will need to repeat this process and update the version number.
+Currently this branch relies on a snapshotted version of bugsnag-android, which should _just work_.
```
bengourley: image

PR #740, opened 2020-02-12
Serialize android configuration in react-native bridge

fractalwrench:
## Goal
Serializes the Android configuration into a `WritableMap` that can be crossed across the RN bridge, for use by the JS notifier when it initialises.

This PR depends on #739 which has not yet been merged.

## Changeset

- Altered the `WritableMapSerializer` method to take a `Map` parameter rather than a `WritableNativeMap`. The `WritableNativeMap` relies on the Android framework and is therefore much harder to unit test.
- Added an [extension function](https://kotlinlang.org/docs/reference/extensions.html) that converts a `Map` into a `WritableNativeMap`, which reduces the surface area of code that interacts with React Native Java APIs.
- Updated `ConfigSerializer` to serialize all configuration options that might be needed by the JS layer.

## Tests

- Added unit test for the `ConfigSerializer` class
- Verified manually in an example app that bugsnag loads and is capable of sending a handled JS error. Note: you currently **must** alter the `releaseStage` serialized in `ConfigSerializer` to `"production"` as the JS layer incorrectly crashes when this is a null value.

### DIFF
--- a/packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
+++ b/packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
@@ -24,7 +24,9 @@ public class ConfigSerializer implements WritableMapSerializer<ImmutableConfig> 
         map.put("enabledReleaseStages", config.getEnabledReleaseStages());
         map.put("releaseStage", config.getReleaseStage());
         map.put("buildUuid", config.getBuildUuid());
-        map.put("appVersion", config.getAppVersion());
+        if (config.getAppVersion() != null) {
+            map.put("appVersion", config.getAppVersion());
+        }
         map.put("versionCode", config.getVersionCode());
         map.put("codeBundleId", config.getCodeBundleId());
         map.put("type", config.getAppType());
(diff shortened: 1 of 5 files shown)


fractalwrench:
Rebased onto target branch in order to pass CI.

packages/react-native/android/src/main/java/com/bugsnag/reactnative/ConfigSerializer.java
```diff
@@ -1,39 +1,66 @@
...
+        map.put("apiKey", config.getApiKey());
+        map.put("autoDetectErrors", config.getAutoDetectErrors());
+        map.put("autoTrackSessions", config.getAutoTrackSessions());
+        map.put("sendThreads", config.getSendThreads().toString());
+        map.put("discardClasses", config.getDiscardClasses());
+        map.put("projectPackages", config.getProjectPackages());
+        map.put("enabledReleaseStages", config.getEnabledReleaseStages());
+        map.put("releaseStage", config.getReleaseStage());
+        map.put("buildUuid", config.getBuildUuid());
+        map.put("appVersion", config.getAppVersion());
```
bengourley:
```suggestion
        if (config.getAppVersion() != null) {
            map.put("appVersion", config.getAppVersion());
        }
```

PR #751, opened 2020-02-17
Populate React Native payload info for Android

fractalwrench:
## Goal

Populates the `payloadInfo()` method for Android which can be invoked over the React Native bridge. This gives JavaScript errors access to the same metadata values which are available for native events.

## Changeset

- Converted `InternalHooks` to a class which can be instantiated and that accesses hidden fields on `Client`, which facilitates mocking this object for testing.
- Added serializers for `App/Device/Breadcrumb/Stackframe`, which convert a POJO to a map that can be serialized over the React Native bridge.
- Updated `MapExtensions` to support converting long values over the React bridge
- Implement `getPayloadInfo()` so that it returns a promise value when invoked.

## Tests
Added new unit tests for serializers, and verified that the app does not crash when triggering a handled JS error.


packages/react-native/android/src/main/java/com/bugsnag/reactnative/MapExtensions.kt
```diff
@@ -13,6 +13,7 @@ internal fun Map<String, Any?>.toWritableMap(): WritableMap {
             null -> nativeMap.putNull(key)
             is Boolean -> nativeMap.putBoolean(key, obj)
             is Int -> nativeMap.putInt(key, obj)
+            is Long -> nativeMap.putDouble(key, obj.toDouble())
```
tomlongridge: Is `toDouble()` the same as `as Double` used below?
```suggestion
            is Long -> nativeMap.putDouble(key, obj as Double)
```

packages/react-native/android/src/main/java/com/bugsnag/reactnative/AppSerializer.kt
```diff
@@ -0,0 +1,19 @@
+package com.bugsnag.reactnative
+
+import com.bugsnag.android.AppWithState
+
+internal class AppSerializer : WritableMapSerializer<AppWithState> {
+    override fun serialize(map: MutableMap<String, Any?>, app: AppWithState) {
+        map["duration"] = app.duration
```
tomlongridge: Alphabetical ordering of these makes it easier to check against the spec!

packages/react-native/android/src/test/java/com/bugsnag/reactnative/BreadcrumbSerializerTest.java
```diff
@@ -0,0 +1,43 @@
...
+    private Map<String, Object> metadata;
+
+    /**
+     * Generates a Breadcrumb for verifying the serializer
+     */
+    @Before
+    public void setup() {
+        Date timestamp = new Date(0);
+        metadata = new HashMap<>();
+        metadata.put("foo", "bar");
```
tomlongridge: Should we look to test some more complex metadata?

packages/react-native/android/src/main/java/com/bugsnag/reactnative/MapExtensions.kt
fractalwrench: Updated to use `.toDouble()` for both, rather than relying on casts.

PR #757, opened 2020-03-03
Update getPayloadInfo() to take boolean param

fractalwrench:
`getPayloadInfo()` needs to take a boolean parameter to indicate the handledState, so that the native layer can determine whether to collect a thread trace or not.

This change updates the JS layer to pass in the `event.unhandled` value in a map, and updates `BugsnagReactNative` to pass this value into the plugin.

Verified by testing an example app manually with and without the change and ensuring an error was only sent when both changes were in place.

### DIFF
--- a/examples/reactnative/android/app/src/main/java/com/reactnative/MainApplication.java
+++ b/examples/reactnative/android/app/src/main/java/com/reactnative/MainApplication.java
@@ -3,7 +3,7 @@ package com.reactnative;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
-import com.bugsnag.reactnative.BugsnagPackage;
+import com.bugsnag.android.BugsnagPackage;
 import com.facebook.react.ReactNativeHost;
 import com.facebook.react.ReactPackage;
 import com.facebook.react.shell.MainReactPackage;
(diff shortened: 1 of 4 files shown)


examples/reactnative/android/app/src/main/AndroidManifest.xml
```diff
@@ -23,7 +23,7 @@
       </activity>
       <activity android:name="com.facebook.react.devsupport.DevSettingsActivity" />
       <meta-data android:name="com.bugsnag.android.API_KEY"
-                 android:value="8f4b9d17b582d6e4c74d43d3e295b4c5"/>
+                 android:value="5d1ec8bd39a74caa1267142706a7fb20"/>
```
tomlongridge: Was this deliberate and/or does it matter?

fractalwrench: This is not deliberate and doesn't particularly matter - I'll revert to the last value regardless.

examples/reactnative/android/app/src/main/java/com/reactnative/MainApplication.java
```diff
@@ -46,7 +46,6 @@ public ReactNativeHost getReactNativeHost() {
   @Override
   public void onCreate() {
     super.onCreate();
-    BugsnagReactNativePlugin.register();
```
bengourley: I'm surprised to see this removed and not see the plugin added to configuration. How does the plugin now load itself?

PR #758, opened 2020-03-04
Forward MessageEvent to React Bridge

fractalwrench:
Forwards `MessageEvent` messages received on native layer state changes to the React Bridge so that the JS layer can update itself.

This implementation uses the `RCTDeviceEventEmitter` module to send messages, which the JS layer can subscribe to. Each message is converted from a `MessageEvent` to a `WritableMap` which the React Bridge is capable of serializing. The subscription itself is taken out in JS and will be addressed in a separate PR when the JS implementation reaches this stage.

Tested by verifiying that the message is logged in an example app when setting the user.

### DIFF
--- a/packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
+++ b/packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
@@ -13,6 +13,14 @@ import com.facebook.react.modules.core.DeviceEventManagerModule.RCTDeviceEventEm
 class BugsnagReactNative(private val reactContext: ReactApplicationContext) :
     ReactContextBaseJavaModule(reactContext) {
 
+    internal companion object {
+        private const val UPDATE_CONTEXT = "ContextUpdate"
+        private const val UPDATE_USER = "UserUpdate"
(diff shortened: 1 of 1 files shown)


packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
```diff
@@ -25,15 +27,35 @@ class BugsnagReactNative(reactContext: ReactApplicationContext) :
...
+    private fun emitEvent(event: MessageEvent) {
+        logger.d("Received MessageEvent: ${event.type}")
+
+        val map = Arguments.createMap()
+        map.putString("type", event.type)
+
+        when (event.type) {
+            "ContextUpdate" -> map.putString("data", event.data as String?)
+            "UserUpdate" -> map.putMap("data", Arguments.makeNativeMap(event.data as Map<String, Any?>?))
+            "MetadataUpdate" -> map.putMap("data", Arguments.makeNativeMap(event.data as Map<String, Any?>?))
```
tomlongridge: Is it worth considering enumerating these types?

fractalwrench: Extracted these types to constants

PR #759, opened 2020-03-05
V7: Update config behaviour and plugin interface

bengourley:
The spec now stipulates that the only invalid configuration that should throw an error is `apiKey`. All others should log a warning and proceed with the default value.

The plugin interface has also been updated. `init()` is now `load()`, and the `client.use()` method has been disabled in favour of passing plugins in to `config.plugins[]`.

Todo:
- [x] Update changelog
- [x] Update upgrade guide


packages/plugin-network-breadcrumbs/network-breadcrumbs.js
```diff
@@ -7,167 +7,165 @@ const REQUEST_METHOD_KEY = 'BS~~M'
...
+        if (this.status >= 400) {
+          // contacted server but got an error response
+          client.leaveBreadcrumb('XMLHttpRequest failed', metadata, BREADCRUMB_TYPE)
+        } else {
+          client.leaveBreadcrumb('XMLHttpRequest succeeded', metadata, BREADCRUMB_TYPE)
+        }
+      }
 
-function handleXHRError () {
-  if (includes(getIgnoredUrls, this[REQUEST_URL_KEY])) {
```
djskinner: Was this a bug then?

packages/plugin-node-uncaught-exception/test/uncaught-exception.test.js
```diff
@@ -7,30 +7,31 @@ const plugin = require('../')
 describe('plugin: node uncaught exception handler', () => {
   it('should listen to the process#uncaughtException event', () => {
     const before = process.listeners('uncaughtException').length
-    const c = new Client({ apiKey: 'api_key' })
-    c.use(plugin)
+    const c = new Client({ apiKey: 'api_key', plugins: [plugin] })
     const after = process.listeners('uncaughtException').length
     expect(after - before).toBe(1)
+    expect(c).toBe(c)
```
djskinner: What is the purpose of these assertions?

packages/plugin-react/src/index.js
```diff
@@ -1,7 +1,12 @@
-module.exports = {
-  name: 'react',
-  init: (client, React = window.React) => {
+module.exports = class BugsnagReactPlugin {
+  constructor (React = window.React) {
     if (!React) throw new Error('cannot find React')
```
djskinner: Should we be throwing here if we take the idea that the only configuration that should blow up is an invalid API KEY to its full extent?

packages/plugin-vue/src/index.js
```diff
@@ -1,7 +1,12 @@
-module.exports = {
-  name: 'vue',
-  init: (client, Vue = window.Vue) => {
+module.exports = class BugsnagVuePlugin {
+  constructor (Vue = window.Vue) {
     if (!Vue) throw new Error('cannot find Vue')
```
djskinner: Same question as with plugin-react

packages/plugin-network-breadcrumbs/network-breadcrumbs.js
bengourley: oh, I didn't even notice. Looks like it was!

packages/plugin-react/src/index.js
bengourley: If we don't blow up here, when a user attempts to do `const ErrorBoundary = getPlugin('react')`, that will return `undefined`, so it will unexpectedly blow up when they try to render. The former seems more desirable.

packages/plugin-node-uncaught-exception/test/uncaught-exception.test.js
bengourley: Appeasing the linter :(

instantiating a client and then not referring to it leads to an unused variable error. If I remove the assignment to `client`, it errors with the "don't use new for side effects" rule.

I didn't want to ignore the line where it was instantiated, so this seemed like the least bad option. I'm sure there's a better way.

djskinner: Fair enough. We can revisit this as part of the TS conversion.

PR #764, opened 2020-03-12
feat(react-native): Send notifier version, RN version and js engine to native

bengourley:
Depends on https://github.com/bugsnag/bugsnag-android/pull/852.


bengourley:
To test this out you need to publish the current bugsnag-android v5-react-native branch to mavenLocal and depend on that.

PR #768, opened 2020-03-16
Update context in React Native Cocoa layer

fractalwrench:
Updates the context in the Cocoa layer when the JS layer invokes the native method.


fractalwrench:
Cocoa would generally use `nil` if an inappropriate value was passed in, so this is less of a concern than it would be on Android.

PR #772, opened 2020-03-18
Update metadata in Cocoa react-native layer

fractalwrench:
Updates the metadata when the JS layer invokes the native method. If the update is nil, the section will be cleared, otherwise its values will be added to existing metadata.


fractalwrench:
Rebased and updated PR to use `v6` Cocoa syntax.

PR #775, opened 2020-03-19
feat: Warn when API key is not 32 hex chars

bengourley:
The spec says to reject non-string, null or empty api keys. Anything else should be accepted, but if it is not exactly 32 hex chars, a warning should be printed.

### DIFF
--- a/packages/core/test/client.test.ts
+++ b/packages/core/test/client.test.ts
@@ -36,6 +36,37 @@ describe('@bugsnag/core/client', () => {
         unhandledRejections: true
       })
     })
+
+    it('warns with a valid but incorrect-looking api key', () => {
+      const logger = {
+        debug: jest.fn(),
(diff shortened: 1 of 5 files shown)


packages/core/test/client.test.ts
```diff
@@ -36,6 +36,37 @@ describe('@bugsnag/core/client', () => {
         unhandledRejections: true
       })
     })
+
+    it('warns with a valid but incorrect-looking api key', () => {
```
djskinner: Do _valid but incorrect-looking_ api keys exist?

bengourley: Valid in the sense that the notifier will accept it, but not one our app would ever generate. This allows us to put a placeholder API key in examples/snippets etc. that will still run properly, but not send errors to any dashboard.

PR #776, opened 2020-03-19
fix: plugin typing

djskinner:
- Fixes typing issue with the React and Vue plugins with interface merging. See https://github.com/microsoft/TypeScript/issues/340 for details and fix.
- Adds plugin-react and plugin-vue to type checking so that declarations are checked.
- Fixes `Cannot use namespace 'React' as a type.ts(2709)` type error which allowed anything to be passed to the React plugin.

### DIFF
--- a/bin/bundle-types
+++ b/bin/bundle-types
@@ -3,6 +3,11 @@
 # exit on all errors
 set -e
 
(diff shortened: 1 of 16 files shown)


djskinner:
@bengourley I will test from the branch today.

PR #777, opened 2020-03-19
Implement leaveBreadcrumb in Cocoa React Native layer

fractalwrench:
Implements the `leaveBreadcrumb` method in the Cocoa layer for React Native. This adds a breadcrumb to the native layer when the JS layer invokes the native method.


fractalwrench:
Putting on hold for now as `addBreadcrumbWithBlock` is being removed in a separate PR, so this will need adjusting at the same time

fractalwrench:
Rebased against latest v6 Cocoa changes, and updated PR to use new syntax. I've confirmed that a JS breadcrumb is reported to the dashboard for a native event.

Screenshot 2020-04-27 at 16 49 08

PR #783, opened 2020-03-24
feat(react-native): Support remote debugging in Chrome

bengourley:
This PR ensures that Bugsnag works when a React Native application is run in remote debugging mode, where the JS is executed in a Chrome browser.

Synchronous native methods are not allowed when debugging remotely, which we use for initialisation (to ensure no errors are missed). The approach, with caveats is as follows:

- `Configuration.load()` and `Bugsnag.createClient()` are removed from the public interface, reducing the impact of asynchronous behaviour
- In normal situations, the way Bugsnag initialises remains exactly the same
- In the remote debugger, `Bugsnag.start()` now returns a proxy which stubs a client, until the async configure method has completed. Until such point, it will no-op and log any synchronous method calls. After successful configuration it will forward method calls on to the initialised client.

N.B. this is only on Android for the time being.

### DIFF
--- a/packages/react-native/src/notifier.js
+++ b/packages/react-native/src/notifier.js
@@ -1,6 +1,14 @@
 const NativeModules = require('react-native').NativeModules
 const NativeClient = NativeModules.BugsnagReactNative
 
+const REMOTE_DEBUGGING_WARNING = `Bugsnag cannot initialise synchronously when running in the remote debugger.
+
+Error reporting is still supported, but synchronous calls afted Bugsnag.start() will no-op. This means Bugsnag.leaveBreadcrumb(), Bugsnag.setUser() and all other methods will only begin to work after a short delay.
+
+This only affects the remote debugger. Execution of JS in the normal way (on the device) is not affected.`
+
+const isDebuggingRemotely = !global.nativeCallSyncHook
+
 const name = 'Bugsnag React Native'
 const { version } = require('../package.json')
 const url = 'https://github.com/bugsnag/bugsnag-js'
@@ -16,7 +24,7 @@ const delivery = require('@bugsnag/delivery-react-native')
 
 const BugsnagPluginReact = require('@bugsnag/plugin-react')
 
-const { schema, load } = require('./config')
+const { schema, load, loadAsync } = require('./config')
 
 const internalPlugins = [
   require('@bugsnag/plugin-react-native-session')(NativeClient),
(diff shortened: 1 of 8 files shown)


packages/react-native/src/notifier.js
```diff
@@ -1,6 +1,14 @@
 const NativeModules = require('react-native').NativeModules
 const NativeClient = NativeModules.BugsnagReactNative
 
+const REMOTE_DEBUGGING_WARNING = `Bugsnag cannot initialise synchronously when running in the remote debugger.
```
djskinner: Message looks good to me. Just a reminder about initialise -> initialize as you mentioned.

packages/react-native/src/notifier.js
```diff
@@ -1,6 +1,14 @@
...
+
+Error reporting is still supported, but synchronous calls afted Bugsnag.start() will no-op. This means Bugsnag.leaveBreadcrumb(), Bugsnag.setUser() and all other methods will only begin to work after a short delay.
+
+This only affects the remote debugger. Execution of JS in the normal way (on the device) is not affected.`
+
+const isDebuggingRemotely = !global.nativeCallSyncHook
```
djskinner: Documentation for this is hard to find. I found https://github.com/facebook/react-native/commit/417e191a1cfd6a049d1d4b0a511f87aa7f176082 which partially helps my understanding.

packages/react-native/src/notifier.js
```diff
@@ -29,61 +37,86 @@ const internalPlugins = [
...
+      Bugsnag._client = createClient(opts)
+      Bugsnag._client._depth += 1
+      return Bugsnag._client
+    } else {
+      console.warn(REMOTE_DEBUGGING_WARNING)
+      let initialised = false
+      const stubClient = {}
+      CLIENT_METHODS.reduce((accum, m) => {
+        stubClient[m] = new Proxy(() => {
+          console.warn(`Bugsnag.${m}() will no-op because Bugsnag has not yet initialised`)
```
djskinner: Consider: `This call to Bugsnag.${m}() is a no-op because remote debugging is enabled and Bugsnag has not yet initialized`

packages/react-native/types/bugsnag.d.ts
```diff
@@ -1,13 +1,12 @@
...
+import { Client, Config } from '@bugsnag/core'
 
 // these properties are allowed to be configured in the JS layer
 type Configurable = 'onError' | 'onBreadcrumb' | 'logger' | 'metadata' | 'user' | 'context' | 'plugins'
 
 type ReactNativeConfig = Pick<Config, Configurable>
 
-interface ReactNativeBugsnagStatic extends BugsnagStatic {
-  start(apiKeyOrOpts?: string | ReactNativeConfig | LoadedConfig): Client
-  createClient(apiKeyOrOpts?: string | ReactNativeConfig | LoadedConfig): Client
```
djskinner: Are multiple clients no support in RN?

packages/react-native/src/notifier.js
bengourley: Yeah that's the only place I could see it. I'll add a comment pointing to there.

packages/react-native/types/bugsnag.d.ts
bengourley: No – it doesn't make sense as it is backed by a singleton native notifier so everything is would be routed by that anyway.

PR #787, opened 2020-03-30
V7: allow top level metadata

bengourley:
To avoid breaking custom filters unnecessarily, this PR enables metadata to be supplied without a section key.

### DIFF
--- a/packages/core/types/client.d.ts
+++ b/packages/core/types/client.d.ts
@@ -34,8 +34,8 @@ declare class Client {
   ): void;
 
   // metadata
-  public addMetadata(section: string, values: { [key: string]: any }): void;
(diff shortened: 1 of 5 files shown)


packages/core/types/client.d.ts
```diff
@@ -34,8 +34,8 @@ declare class Client {
   ): void;
 
   // metadata
-  public addMetadata(section: string, values: { [key: string]: any }): void;
-  public addMetadata(section: string, key: string, value: any): void;
+  public addMetadata(section: string | null, values: { [key: string]: any }): void;
+  public addMetadata(section: string | null, key: string, value: any): void;
```
djskinner: These also need changing on `Event`.

PR #789, opened 2020-03-30
Maze runner retry on failure and fail fast

snmaynard:
Companion to bugsnag/maze-runner#91


twometresteve:
I'm happy with this if you are, @bengourley , and we'll be removed the blanket fail fast and retry from MaeRunner in due course.  I'll leave you to approve and merge this according to your taste.

PR #797, opened 2020-04-03
feat(plugin-react): allow FallbackComponent to reset ErrorBoundary

### LINKED ISSUE #638: Question: how to integrate Bugsnag in a custom React ErrorBoundary?
I use custom [error boundaries](https://reactjs.org/docs/error-boundaries.html) in my application. They are used in different parts of the page, have a beautiful look and a «retry» button. The proposed Bugsnag approach to wrap the whole application in the Bugsnag's error boundary won't work in my case because the errors are already caught by my error boundaries.

```jsx
function MyApp() {
  return (
    <div>
      <MyErrorBoundary>
        Column 1
      </MyErrorBoundary>
      <MyErrorBoundary>
        Column 2
      </MyErrorBoundary>
    </div>
  );
}

const ErrorBoundary = bugsnagClient.getPlugin('react')

ReactDOM.render(
  <ErrorBoundary> {/* Will never catch */}
    <MyApp />
  </ErrorBoundary>,
  document.getElementById('app')
)
```

So the question is: what is the correct way to add a Bugsnag error reporting to my custom error boundary component?

```jsx
class MyErrorBoundary extends React.PureComponent {
  state = {hasError: false};

  static getDerivedStateFromError(error) {
    return {hasError: true};
  }

  componentDidCatch(error, errorInfo) {
    // What to type here?
    // Maybe `bugsnagClient.notify(error)`? But how to send `errorInfo` too?
  }

  render() {
    if (this.state.hasError) {
      // You can render any custom fallback UI
      return 'My custom pretty error display with a retry button';
    }

    return this.props.children; 
  }

  retry() {
    this.setState({hasError: false});
  }
}
```

djskinner:
Closes #638


packages/plugin-react/src/index.js
```diff
@@ -15,6 +15,11 @@ module.exports = class BugsnagReactPlugin {
           error: null,
           info: null
         }
+        this.handleClearError = this.handleClearError.bind(this)
+      }
+
+      handleClearError () {
+        this.setState({ error: null, info: null })
```
bengourley, line 22: did you think better of making this async to prevent synchronous loops in the end?

djskinner: I added a test case and tried it out on a real app and it doesn't result in a stack overflow. The error repeats a handful of times then stops so I assume React has its own detection mechanism and prevents an infinite rendering loop.

PR #799, opened 2020-04-06
V7: revert top level metadata

bengourley:
Removes the feature that was added in #787, and provides guidance on how to maintain consistency with the display and custom filters after the upgrade.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -235,6 +235,13 @@ It remains possible to supply initial metadata in configuration:
   })
 ```
 
+Previously, it was possible to add "top-level" metadata, i.e. data that you did not provide a specific `section` name for. In the dashboard, this would display under a tab with the heading "Custom". Since a section name is now required, for continuity both visually and for any custom filters you may have set up based on such data, you should set the section name to `'custom'`:
+
+```diff
+- bugsnagClient.metaData = { processId: 'aa874cbd', role: 'image-resizer' }
++ Bugsnag.addMetadata('custom', { processId: 'aa874cbd', role: 'image-resizer' })
+```
+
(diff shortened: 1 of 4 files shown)


UPGRADING.md
```diff
@@ -287,21 +283,7 @@ Here are some examples:
...
--       instanceId: component.instanceId
--     }
--   }
-- })
-+ Bugsnag.notify(err, event => {
-+   event.addMetadata('component, {
-+     instanceId: component.instanceId
-+   }
-+ })
++ Bugsnag.notify(err, event => { severity: 'info' })
```
bullmo: This is a regression

UPGRADING.md
```diff
@@ -287,21 +283,7 @@ Here are some examples:
 ```diff
   // changing severity
 - bugsnagClient.notify(err, { severity: 'info' })
-+ Bugsnag.notify(err, event => { event.severity = 'info' })
-
-  // adding metadata
```
bullmo, line 292: Personally I think that this example is useful (in addition to the example about when there is no section)

packages/core/types/client.d.ts
```diff
@@ -34,8 +34,8 @@ declare class Client {
   ): void;
 
   // metadata
-  public addMetadata(section: string | null, values: { [key: string]: any }): void;
-  public addMetadata(section: string | null, key: string, value: any): void;
+  public addMetadata(section: string, values: { [key: string]: any }): void;
+  public addMetadata(section: string, key: string, value: any): void;
```
djskinner, line 38: I believe the equivalent types in `event.d.ts` will need updating too.

PR #800, opened 2020-04-07
RN: Integration tests

bengourley:
This PR adds some initial integration tests to determine whether this is a useful level at which to author tests.


bengourley:
> I think there will always be some unavoidable boilerplate when doing mocks in jest, especially when also using TypeScript. The alternative to mocks is to mock inline as we have a tendency to do in the dashboard but its swings and roundabouts.

I tried to do it inline and would prefer to do that. I don't like that the filesystem-based mocks don't live next to the tests, nor that they are loaded for every test and you have to clean state on them after each test.

If you think it should be possible I'll try again.

> Is the rename packages/react-native/react-native.config.js → ...t-native/_react-native.cosdlfkjnfig.sdfjs meant to be in this PR?

Definitely not!

djskinner:
>  I don't like that the filesystem-based mocks don't live next to the tests, nor that they are loaded for every test and you have to clean state on them after each test.

Yes. Another possible limitation of filesystem-based mocks is that is becomes more difficult to specify custom mock behaviour. You don't always want your mocks to behave in exactly the same way for all test suites. With the inline approach you can more explicitly set the behaviour you need for each test suite and if and when necessary share mocking code as you please with fewer restrictions.

With `jest.mock` and `jest.requireActual` you should be able to get most of the capabilities supplied by `jest.genMockFromModule` and filesystem-based mocks.

PR #808, opened 2020-04-20
fix(expo-cli): Ensure appropriate code is inserted based on Bugsnag version

bengourley:
V7-based code was inserted regardless of which version of Bugsnag was actually installed.

This fix detects the version range in package.json to determine whether to insert pre- or post-V7 compatible code.

### DIFF
--- a/packages/expo-cli/commands/install.js
+++ b/packages/expo-cli/commands/install.js
@@ -6,10 +6,11 @@ const install = require('../lib/install')
 const { onCancel } = require('../lib/utils')
 const { blue } = require('kleur')
 const semver = require('semver')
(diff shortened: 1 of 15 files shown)


packages/expo-cli/commands/install.js
```diff
@@ -6,10 +6,11 @@ const install = require('../lib/install')
 const { onCancel } = require('../lib/utils')
 const { blue } = require('kleur')
 const semver = require('semver')
+const detectInstalled = require('.../lib/detect-installed')
```
djskinner, line 9:
```suggestion
const detectInstalled = require('../lib/detect-installed')
```

PR #810, opened 2020-04-21
test(plugin-vue): convert to jest

djskinner:
Unfortunately this doesn't help to add coverage for the manually maintained types. To achieve this in the future we could potentially add co-located (index.d.ts) types and derive a public subset from these.


packages/plugin-vue/test/index.test.ts
```diff
@@ -1,6 +1,6 @@
-const { describe, it, expect } = global
-const BugsnagVuePlugin = require('../src')
-const Client = require('@bugsnag/core/client')
+import BugsnagVuePlugin from '../src'
```
bengourley, line 1: If you import from `../` and ensure `npm run build` has run, would this cover the exported type defs too?

djskinner: Yes I tried that and it works. It's an option. I held off doing it in this PR because it would be different to how most other tests are currently running and introduces a dependency on the test command.

bengourley: `npm run build` is already a dependency of `npm run test:unit`, if that's what you are concerned about?

djskinner: Of course, for the remaining type tests.

If we do this here I wonder about the inconsistency between packages. Some would be running tests against the source, others against the dist. Likely we'd want to take one approach or the other and be consistent throughout.

Whether tests should run against source or transpiled output is a largely philosophical debate (I've had with myself many times in the past) but one practical downside is that many of the unit tests test internals that are not represented by the public types. We may lose some of the benefits of writing tests with TypeScript. We might be able to get around that by using module augmentation rather than the co-located type pattern currently in use. However, module augmentation is fairly limited so we would need to spike it out.

I suppose another factor that would determine our approach here is what the end goal is. Presumably having the source itself written in TypeScript and the public declarations automatically derived is the long term goal. It if is, then augmenting the manual public types to create the private types seems like the wrong way round. Rather the public types should be a derived subset of the private types (e.g. by using `Pick` and `Exclude`), whether that be automatically from TypeScript source or, as an interim step, manually maintained co-located `.d.ts` files that are verified by the fact the tests are written in TypeScript and tested against those (full) types.

bengourley: Ok – in light of that, I'll merge this as is. We now have tickets on the internal backlog to add tests for the public types of plugins.

PR #815, opened 2020-04-23
Package React Native Android artefacts within node_modules

fractalwrench:
## Goal

Packages the React Native Android AARs within the node_modules. This guarantees that the correct version of bugsnag-android is used and avoids Gradle caching issues which we have observed.

This mirrors the approach taken by Facebook's `react-native` package, which uses a local maven repository for Android artefacts.

This should require no change to a user's project with the default setup. There is a chance that a user might have manually added their own Gradle dependency on bugsnag-android, which would result in a classloader error being thrown on startup. This risk can be mitigated by ensuring we document that users should remove any manually added Gradle dependency when upgrading.

## Changeset

- Added a maven repository at `packages/react-native/android/com/bugsnag/android`
- Copied bugsnag-android artefacts into maven repository
- Updated the build.gradle file to add the local maven repository as a source. In development this location defaults to `packages/react-native`
- Created a shell script to automate the generation and copying of AAR artefacts
- The shell script uses the version name in bugsnag-android with a `-react-native` suffix added; this prevents any potential for nameclash with artefacts released on jcenter/mavenCentral

## Tests
Verified that the new packaging works in the example app, and when running `./gradlew test -Pbugsnagdev=true` in the `packages/react-native` directory


packages/react-native/update-android.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
```
bengourley, line 2: I always recommend `set -e` on shell scripts so that it exits on any error

fractalwrench: Added `set -e` to the top of the script.

PR #817, opened 2020-04-24
fix(expo): Ensure types allow start() can be called with no arguments

bengourley:
The implementation of `start()` on Expo allows no arguments. This updates the types to reflect that.


packages/expo/types/test/fixtures/all-options.ts
```diff
@@ -30,3 +30,5 @@ Bugsnag.start({
   logger: undefined,
   redactedKeys: ["foo",/bar/]
 })
+
+Bugsnag.start()
```
twometresteve, line 34: Did you mean to include this?

PR #826, opened 2020-04-29
Serialize cocoa config for React Native JS layer

fractalwrench:
## Goal

Serializes the Configuration object into JSON for use by the JS layer on React Native.

## Changeset
- Updated vendored bugsnag-cocoa code to latest on `v6`
- Created `BugsnagConfigSerializer` which converts `BugsnagConfiguration` into an `NSDictionary`
- Serialized all the configuration options available in Cocoa v6
- Returned the serialized config in the `configure()` method

This has been tested by manually verifying that the example app runs and does not log any warnings about invalid configuration values.

### DIFF
--- a/packages/react-native/ios/BugsnagReactNative/BugsnagConfigSerializer.m
+++ b/packages/react-native/ios/BugsnagReactNative/BugsnagConfigSerializer.m
@@ -19,7 +19,7 @@
     BSGDictInsertIfNotNil(dict, config.enabledReleaseStages, @"enabledReleaseStages");
     BSGDictInsertIfNotNil(dict, config.releaseStage, @"releaseStage");
     BSGDictInsertIfNotNil(dict, config.appVersion, @"appVersion");
-    BSGDictInsertIfNotNil(dict, config.appType, @"type");
+    BSGDictInsertIfNotNil(dict, config.appType, @"appType");
     BSGDictInsertIfNotNil(dict, @(config.persistUser), @"persistUser");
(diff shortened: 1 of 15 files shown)


packages/react-native/ios/BugsnagReactNative/BugsnagConfigSerializer.m
```diff
@@ -0,0 +1,90 @@
...
+
+- (NSDictionary *)serialize:(BugsnagConfiguration *)config {
+    NSMutableDictionary *dict = [NSMutableDictionary new];
+    BSGDictInsertIfNotNil(dict, config.apiKey, @"apiKey");
+    BSGDictInsertIfNotNil(dict, @(config.autoDetectErrors), @"autoDetectErrors");
+    BSGDictInsertIfNotNil(dict, @(config.autoTrackSessions), @"autoTrackSessions");
+    BSGDictInsertIfNotNil(dict, config.enabledReleaseStages, @"enabledReleaseStages");
+    BSGDictInsertIfNotNil(dict, config.releaseStage, @"releaseStage");
+    BSGDictInsertIfNotNil(dict, config.appVersion, @"appVersion");
+    BSGDictInsertIfNotNil(dict, config.appType, @"type");
```
bengourley, line 22:
```suggestion
    BSGDictInsertIfNotNil(dict, config.appType, @"appType");
```

fractalwrench:
Using Objective-C's introspection is an interesting thought - it would require a bit of planning/investigating for excluding properties however, as I'm not aware of any way of distinguishing between public vs private properties.

The `appType` is serialized as `type` for the Android plugin too as that's what the JSON key normally is - I'll raise a PR to address it there too.

PR #832, opened 2020-05-04
feat: allow enhancing runtimeVersions info

fractalwrench:
## Goal

Allows setting extra runtime version information recorded for events and sessions. This changeset largely mirrors the Android implementation and relies on https://github.com/bugsnag/bugsnag-cocoa/pull/578

## Changeset

- `BugsnagReactNative.m` now reads the `reactNativeVersion` and `engine` values when `configure()` is called, and passes these onto the native layer
- Updated vendored code to use https://github.com/bugsnag/bugsnag-cocoa/pull/578
- Whenever a new session/error is captured, the extra runtime version info is added in the native layer

## Tests
Verified with a debugger that the object passed to an onSession/onError block contains the appropriate runtime version information.

Note that there is a known issue with the Cocoa implementation where the Event is not fully populated/persisted, so runtimeVersions information will not be sent in the actual payload. This will be addressed by a separate internal ticket.

### DIFF
--- a/packages/react-native/ios/BugsnagReactNative/BugsnagReactNative.m
+++ b/packages/react-native/ios/BugsnagReactNative/BugsnagReactNative.m
@@ -11,6 +11,8 @@
 - (NSArray *)collectBreadcrumbs;
 - (NSArray *)collectThreads;
 @property id notifier;
+@property id sessionTracker;
+@property BugsnagMetadata *metadata;
 @end
 
 @interface Bugsnag ()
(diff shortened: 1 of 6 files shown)


fractalwrench:
@bengourley I've now updated the native Cocoa implementation so that the JS engine + RN version are passed up in the `getPayloadInfo()` method (confirmed by logging the information out).

There is still  a known issue with the Cocoa implementation where the Event is not fully populated/persisted, so runtimeVersions information will not be sent in the actual payload, and won't be populated in an `onSendError` block. I've verified the value will be set once this address by using a debugger to inspect the `notifyInternal` method call.

PR #839, opened 2020-05-06
plugins: Support obtaining React/Vue reference after Bugsnag.start()

bengourley:
This PR updates the browser plugins to support passing a reference to the framework in use at a point in time _after_ Bugsnag has been started. The details are outlined below:

## `@bugsnag/plugin-react`

Sometimes it will be the case that a `React` reference will not be available until after Bugsnag has been initialised.

This PR introduces an alternative API to lazily pass the reference, but maintains backward compatibility.

The two modes of operation are now as follows:

### Existing
```js
Bugsnag.start({ plugins: [new BugsnagPluginReact(React)] })
const ErrorBoundary = Bugsnag.getPlugin('react')
```

### New (lazy)
```js
Bugsnag.start({ plugins: [new BugsnagPluginReact()] })
const ErrorBoundary = Bugsnag.getPlugin('react')!.createErrorBoundary(React)
```

### New (non-lazy)
```js
Bugsnag.start({ plugins: [new BugsnagPluginReact(React)] })
const ErrorBoundary = Bugsnag.getPlugin('react')!.createErrorBoundary()
```

Whilst the original interface remains, the non-lazy invocation also provides a similar method for obtaining the error boundary to the lazy version. This means it is easier to type.

An attempt has been made to help prevent footguns. For example, if you instantiate the plugin without a reference to React, but then pass the result of `getPlugin('react')` directly to a render, the following error is displayed:

image

## `@bugsnag/plugin-vue`

Similarly the Vue plugin has been updated to maintain backward compatibility, but also to support a framework reference later on.

### Existing
```js
Bugsnag.start({ plugins: [new BugsnagPluginVue(Vue)] })
// no use for the return type for the vue plugin
```

### New (lazy)
```js
Bugsnag.start({ plugins: [new BugsnagPluginVue()] })
Bugsnag.getPlugin('vue').installVueErrorHandler(Vue)
```

## Types

Along with these updates, better typing has been provided to the plugins.

In order to achieve good types for the`getPlugin()` method, rather than centralizing all of the types in the top level notifier, thereby requiring dependencies on React, Vue etc., for the top level notifiers I used the “module augmentation” method – as described in the [TypeScript handbook here](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).

In order to do this, I needed to undo some of the changes that were made in [this PR](https://github.com/bugsnag/bugsnag-js/pull/423), since there we added a prepublish step to each of the top-level notifiers and plugins to bundle the types locally inside their package.

So, with each notifier and plugin having a local copy of the `@bugsnag/core` types inside `dist/types/bugsnag-core` in its package and each having a reference to the `@bugsnag/core` module, it means we can’t use the module augmentation technique, since it will only augment its local copy.

The reason for applying this change in the past was to prevent each notifier having an install-time dependency on `@bugsnag/core`, just in case the consumers were using types. We’re now in the position where the alternative is for people installing `@bugsnag/js`, they’d also get `React`, `Vue`, `Express` etc., just in case they intend to use those plugins. So now it seems like a dependency on `@bugsnag/core` is the better approach.

The changes I have made mean that Bugsnag modules installed together on a project will reference the same single version of `@bugsnag/core`, enabling each of the plugin’s definitions to augment the definition of the client’s `getPlugin()` method to give it a more detailed type.

In a separate PR I will add improvements to the types for plugins that were not touched here.


packages/plugin-react/src/index.js
```diff
@@ -1,48 +1,33 @@
 module.exports = class BugsnagReactPlugin {
-  constructor (React = window.React) {
-    if (!React) throw new Error('cannot find React')
-    this.React = React
+  constructor (...args) {
     this.name = 'react'
+    this.lazy = args.length === 0 && !window.React
```
djskinner, line 4: Is there a use case where I'd want to use deferred loading even if `window.React` is defined?

djskinner: I supposed you could specifically pass `undefined` if you really wanted to do that

bengourley: I don't think passing `undefined` would achieve what you want there, because in that case `args.length = 1` so `this.lazy` would be `false`.

But I'm not sure, hopefully this a road we don't need to go down. If `window.React` is defined but that is not the reference to React that is desired, that implies the existence of _another_ version of React on the page.

If we want to support defering in that case, then I think we would need to special case the value of `args[0]` to allow something like `{ lazy: true }` or `'LAZY'`. To disambiguate.

packages/browser/package.json
```diff
@@ -2,27 +2,27 @@
...
   "description": "Bugsnag error reporter for browser JavaScript",
   "homepage": "https://www.bugsnag.com/",
   "repository": {
     "type": "git",
     "url": "git@github.com:bugsnag/bugsnag-js.git"
   },
   "browser": {
-    "dist/types/bugsnag": "./dist/bugsnag.js"
+    "types/bugsnag": "./dist/bugsnag.js"
   },
```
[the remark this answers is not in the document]
bengourley: https://github.com/bugsnag/bugsnag-js/pull/632

I don't entirely understand it, but it's one (of many) incantations required to make Angular happy.

PR #845, opened 2020-05-11
Fix version scheme

twometresteve:
This change adopts the following version pattern when publishing for CI (or locally, should you wish to us it):

<local version>-<ci flag>-<branch name>.<cmomit id>

Where:
- Local version is retrieved from `lerna.json` (x.y.x only)
- "ci-' is added to distinguish from local use and avoid publishing clashes


scripts/common.js
```diff
@@ -0,0 +1,31 @@
...
+    determineVersion: function determineVersion() {
+        // Form version string - branch name and commit id
+        let commitId = this.getCommitId();
+        let branchName;
+        if (process.env.BRANCH_NAME) {
+            console.log('Using BRANCH_NAME from the environment')
+            branchName = process.env.BRANCH_NAME
+        } else {
+            console.log('Using local Git repo to determine branch name (environment BRANCH_NAME may also be used)')
+            branchName = this.run('git branch --show-current').replace(/\//g,"-")
```
bengourley, line 19: this seems like a new feature of git that my version doesn't have. Can you use `git rev-parse --abbrev-ref HEAD` instead for wider compatibility?

scripts/common.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        // Distinguish from local use
+        let ciIndicator = (process.env.BUILDKITE ? "ci-" : "")
+
+        // Get the current version from lerna.json
+        const lerna = require("../lerna.json");
```
bengourley, line 26: `require` should be at the top of the file, though the linter may tell you that when it is enabled

scripts/common.js
bengourley, line 19: are there any other characters than `/` that are allowed in branch names that would break this?

twometresteve: There might be others, but I don't think we're in the habit of using them and `/` is the only character that the Buildkite agents replace when creating the `BRANCH_NAME` environment variable.  So I went with the same approach primarily for consistency.

scripts/common.js
twometresteve: Moved.

scripts/common.js
twometresteve: Of course.

twometresteve:
> Doesn't seem like the new files are linted on CI. I think they need to be added to the ci dockerfile `COPY scripts scripts`. I can run `npm run test:lint` locally to see the lint errors, or configure your editor's eslint integration.
 
Now added, IDE set up and violations resolved.

scripts/common.js
```diff
@@ -0,0 +1,31 @@
...
+    if (process.env.BRANCH_NAME) {
+      console.log('Using BRANCH_NAME from the environment')
+      branchName = process.env.BRANCH_NAME
+    } else {
+      console.log('Using local Git repo to determine branch name (environment BRANCH_NAME may also be used)')
+      branchName = this.run('git rev-parse --abbrev-ref HEAD').replace(/\//g, '-')
+    }
+
+    // Distinguish from local use
+    const ciIndicator = (process.env.BUILDKITE ? 'ci-' : '')
```
Cawllec, line 24: The `BUILDKITE` environment variable will need to be added into the `docker-compose.yml` in order for it to be passed through to the container.

twometresteve: Nice spot - I added that at the last minute and assumed too much.

PR #854, opened 2020-05-14
feat(plugin-react): allow FallbackComponent to reset ErrorBoundary

bengourley:
Retargetting #797 at the better types branch.


djskinner:
Looks good. I tested the types against a React TS project.

PR #857, opened 2020-05-19
Release v7.1.0

bengourley:
This update contains some substantial changes to plugin type definitions. If you are using TypeScript alongside a framework, you may need to make changes to your app. Please refer to the [upgrade guide](https://github.com/bugsnag/bugsnag-js/blob/fa1a3e85eaeb097b20ba3b0e91804ea369ae8de0/UPGRADING.md).

### Changed

- (plugin-react|plugin-vue): Support late passing of framework reference [#839](https://github.com/bugsnag/bugsnag-js/pull/839)

### Added

- (plugin-react): Add type definitions for `Bugsnag.getPlugin('react')` [#839](https://github.com/bugsnag/bugsnag-js/pull/839)
- (plugin-vue): Add type definitions for `Bugsnag.getPlugin('vue')` [#839](https://github.com/bugsnag/bugsnag-js/pull/839)
- (plugin-react): Add `clearError` prop to `ErrorBoundary`. [#797](https://github.com/bugsnag/bugsnag-js/pull/797)


UPGRADING.md
```diff
@@ -1,6 +1,70 @@
...
+
+// RECOMMENDED: to make use of the provided type definitions, update to the new api
+const ErrorBoundary = Bugsnag.getPlugin('react')!.createErrorBoundary()
+```
+
+_Note the use of the `!` operator._ The `getPlugin('react')` call will only return something if the react plugin was provided to `Bugsnag.start({ plugins: […] })`.
+
+### Plugins
+
+In order to work, The React and Vue plugin both require a reference to the respective framework to be passed in. This was required in the constructor, which meant there was no way to load Bugsnag _before_ the framework. To support this, we now support supplying the framework reference _after_ Bugsnag has started.
```
twometresteve, line 27:
```suggestion
In order to work, the React and Vue plugins both require a reference to the respective framework to be passed in. This was required in the constructor, which meant there was no way to load Bugsnag _before_ the framework. To support this, we now support supplying the framework reference _after_ Bugsnag has started.
```

PR #865, opened 2020-05-21
fix(expo-cli): Ensure version detection logic for @bugsnag/expo is correct

bengourley:
The existing logic to detect whether @bugsnag/expo v6 or v7 was used only worked for exactly v7.0.0, nothing higher.

This updates the logic to ensure the minimum version in the range is at least v7. I couldn't find a way to express "less than or equal to range", so I had to provide a high version of v6 that we will never conceivably publish, in order to be inclusive of v7.0.0 – if that makes sense.

### DIFF
--- a/packages/expo-cli/lib/insert.js
+++ b/packages/expo-cli/lib/insert.js
@@ -35,7 +35,7 @@ Bugsnag.start();
 
 const getCode = async (projectRoot) => {
   const manifestRange = await detectInstalled(projectRoot)
-  const isPostV7 = !manifestRange || semver.satisfies('7.0.0', manifestRange)
+  const isPostV7 = !manifestRange || semver.ltr('6.99.99', manifestRange)
   return code[isPostV7 ? 'postV7' : 'preV7']
 }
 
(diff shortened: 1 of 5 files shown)


packages/expo-cli/lib/insert.js
```diff
@@ -35,7 +35,7 @@ Bugsnag.start();
 
 const getCode = async (projectRoot) => {
   const manifestRange = await detectInstalled(projectRoot)
-  const isPostV7 = !manifestRange || semver.satisfies('7.0.0', manifestRange)
+  const isPostV7 = !manifestRange || semver.ltr('6.99.99', manifestRange)
```
twometresteve, line 38: What if someone used a range to skip 7 entirely for some reason?  E.g. `6.6.1 || >=8.0.0`.  We'd want v7 behaviour here but it would act as v6.

Is there some way to sidestep this entirely by cunning write a single piece of code to handle both scenarios?

twometresteve: Ok, discussed in person and agreed that that is not a realistic scenario due to the nature of CLI use.

PR #867, opened 2020-05-22
Add clearer error message when Bugsnag.start() not present

fractalwrench:
Adds a clearer error message when `Bugsnag.start()` is not present for the iOS and Android clients. Currently if a user does not add `Bugsnag.start()` to the appropriate class or removes it, then the app crashes with a reasonably cryptic error message. As this is something that can be detected we can improve the error messaging to look like below instead:

Screenshot 2020-05-22 at 10 11 21

### DIFF
--- a/packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
+++ b/packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
@@ -46,7 +46,7 @@ class BugsnagReactNative(private val reactContext: ReactApplicationContext) :
         val client = try {
             Bugsnag.getClient()
         } catch (exc: IllegalStateException) {
-            throw IllegalStateException("Failed to initialise Android Client, please check you have " +
+            throw IllegalStateException("Failed to initialise the native Bugsnag Android client, please check you have " +
             "added Bugsnag.start() in the onCreate() method of your Application subclass")
         }
         return try {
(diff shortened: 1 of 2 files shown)


packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
```diff
@@ -43,8 +43,13 @@ class BugsnagReactNative(private val reactContext: ReactApplicationContext) :
 
     @ReactMethod(isBlockingSynchronousMethod = true)
     fun configure(env: ReadableMap): WritableMap {
+        val client = try {
+            Bugsnag.getClient()
+        } catch (exc: IllegalStateException) {
+            throw IllegalStateException("Failed to initialise Android Client, please check you have " +
```
tomlongridge, line 49:
```suggestion
            throw IllegalStateException("Failed to initialise the native Bugsnag Android client, please check you have " +
```

packages/react-native/ios/BugsnagReactNative/BugsnagReactNative.m
```diff
@@ -56,7 +56,7 @@ @implementation BugsnagReactNative
     self.configSerializer = [BugsnagConfigSerializer new];
 
     if (![Bugsnag bugsnagStarted]) {
-        return nil;
+        [NSException raise:@"BugsnagException" format:@"Failed to initialise Cocoa Client, please check you have added Bugsnag.start() in the application:didFinishLaunchingWithOptions: method of your AppDelegate subclass"];
```
tomlongridge, line 59: This will most likely be Objective-C for a RN project won't it?

```suggestion
        [NSException raise:@"BugsnagException" format:@"Failed to initialise the Bugsnag Cocoa client, please check you have added [Bugsnag start] in the application:didFinishLaunchingWithOptions: method of your AppDelegate subclass"];
```

Issue #620, opened 2019-09-13
[plugin-express] body not showing up in request tab.


azizhk:
![Screenshot 2019-09-13 at 3 13 52 PM](https://user-images.githubusercontent.com/5094296/64853540-34deb000-d639-11e9-837a-907596391260.png)

So I have a POST request which throws an error, but body does not get logged as shown in the image.
The problem is that bugnsag generated the metadata in the middleware instead of the errorHandler. (Don't know why this is done, please don't add overhead and allow application code to run first)

The code over here expects req.body to be an object:
https://github.com/bugsnag/bugsnag-js/blob/7231ea7463e4d3bc63f46f9694fb694eb17fa01f/packages/plugin-express/src/request-info.js#L21

We use [body-parser](https://www.npmjs.com/package/body-parser) to create req.body into an object, but that middleware is after the bugsnag middleware. (As bugnsag recommends that the first middleware should be the bugnsags over [here](https://docs.bugsnag.com/platforms/javascript/express/))

I've moved the bugsnag middleware to be after body-parser for now. But I would ideally like if bugsnag can move the request metadata creation to the errorHandler.

mattdyoung:
Hi @azizhk 

Thanks for the report. We're taking a look and will decide on the best way forward.

phillipsam:
Hi @azizhk - we have released a fix for this in [v7.2.0](https://github.com/bugsnag/bugsnag-js/releases/tag/v7.2.0).

PR #872, opened 2020-06-01
plugin-express: fetch request body in onError

imjoehaines:
This fixes an issue where the body may not exist when the request handler runs and so the body will be missing from the error report in the dashboard

This can happen because the body may be parsed by middleware that runs after the `requestHandler`, which is fairly likely because Bugsnag should be installed as early as possible

See https://github.com/bugsnag/bugsnag-js/issues/620 for more details

I tried to write some tests for this but struggled to get anything working. I may be missing something but it seems like it's hard/impossible to test due to how the `requestHandler` works (happy to be corrected though!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## TBD
+
+### Fixed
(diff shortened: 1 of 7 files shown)


bengourley:
I tested this manually and it worked as desired.

I also figured a way to get the maze runner tests to verify this behaviour, so have pushed a test up for that. Tagging @twometresteve for a review from a testing POV.

PR #877, opened 2020-06-05
test(plugin-strip-project-root): convert to jest

djskinner:


packages/core/client.d.ts
```diff
@@ -38,8 +38,8 @@ interface Delivery {
  * This is currently used by the unit tests. These will be rolled into the
  * module itself once it is converted to TypeScript.
  */
-export default class ClientWithInternals extends Client {
-  public constructor(opts: Config, schema?: {[key: string]: any}, internalPlugins?: Plugin[])
+export default class ClientWithInternals<T extends Config = Config> extends Client {
```
djskinner, line 41: This change allows configs that extend the base `Config` type, such as `NodeConfig`. Required for the tests e.g. `new Client({ apiKey: 'api_key', projectRoot: '/app', plugins: [plugin] }` where `projectRoot` is not part of the base `Config` type.

PR #881, opened 2020-06-05
Add additional app and device fields

imjoehaines:
This PR adds a number of additional fields to the `app` and `device` notify keys, specifically:

- **plugin-browser-device**
    - `orientation` — the orientation of the browser using the [`Screen.orientation` API](https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation). This is fairly widely supported, but not universally (e.g. Safari doesn't support this). There is also fallback which compares the browser width & height, to support more devices
- **plugin-expo-device**
    - `totalMemory` — the total memory available on the device, fetched with the `expo-device` module, which is a new dependency
- **plugin-node-device**
    - `osName` — this uses `os.platform`, so reports the kernel name. This is still a good indicator and is much simpler and less error prone than trying to detect the OS
    - `osVersion` — this uses `os.release`, which is also kernel information
    - `totalMemory` — the total memory available to the node process (`os.totalmem`)
    - `freeMemory` — the free memory available to the node process (`os.freemem`)

Two new plugins have been added:
- **plugin-browser-app**:
    - Adds the `duration` field to `event.app`, like the existing `plugin-expo-app`
- **plugin-node-app**
    - Adds the `duration` field to `event.app`, like the existing `plugin-expo-app`

These plugins are currently identical so it may make sense to merge them into a `plugin-duration`, which could then be used in Expo too. I'm not sure if we'd want to keep them separate though, so that there's room for them to diverge in future if we add more app fields to the browser or Node


packages/core/test/client.test.ts
```diff
@@ -275,6 +275,30 @@ describe('@bugsnag/core/client', () => {
...
+
+          expect(payload.events[0].app.duration).toBeGreaterThanOrEqual(delayMs)
+          expect(payload.events[0].app.duration).toBeLessThanOrEqual(maximum)
+
+          done()
+        },
+        sendSession: () => {}
+      }))
+
+      setTimeout(() => client.notify(new Error('oh em eff gee')), delayMs)
```
djskinner, line 299: Not a big deal since it's only 10ms but an alternative to using an actual time delay is to consider using jest fake timers.

packages/core/client.js
```diff
@@ -58,6 +58,10 @@ class Client {
     this.Breadcrumb = Breadcrumb
     this.Session = Session
 
+    // store the time we were constructed so we can calculate the duration that
+    // the application has been running when an error is thrown
+    this._startTime = new Date()
```
djskinner, line 63: Would this be better outside the class, at the module level (as it was with expo) in order to give a more accurate start time? We suggest people start bugsnag first so in the simple case it likely won't make much difference. But for more complex cases the difference between the javascript being parsed and a client being newed could be significant.

packages/plugin-expo-app/app.js
```diff
@@ -1,11 +1,9 @@
 const Constants = require('expo-constants').default
 const { AppState } = require('react-native')
 
-const appStart = new Date()
```
djskinner, line 4: Similar to previous comment I wonder if it would be better to stick with `const appStart = new Date()` at the module level

packages/core/test/client.test.ts
imjoehaines: This looks possible in Jest 26 with ['modern' timer mocks](https://jestjs.io/docs/en/jest-object#jestusefaketimersimplementation-modern--legacy) but we're on version 25 — updating might be out of scope of this PR

packages/plugin-browser-device/device.js
```diff
@@ -3,13 +3,17 @@ const assign = require('@bugsnag/core/lib/es-utils/assign')
...
-module.exports = (nav = navigator) => ({
+module.exports = (nav = navigator, _screen = window.screen) => ({
   load: (client) => {
     const device = {
       locale: nav.browserLanguage || nav.systemLanguage || nav.userLanguage || nav.language,
       userAgent: nav.userAgent
     }
 
+    if (_screen && _screen.orientation && _screen.orientation.type) {
+      device.orientation = _screen.orientation.type
```
bengourley, line 14: The [`window.screen.orientation` API](https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation) isn't supported by all of the browsers in our support matrix, and for some it requires a vendor prefix (see the example on MDN).

It would be good to fall back to the result of this expression:

`document.documentElement.clientWidth > document.documentElement.clientHeight ? 'landscape' : 'portrait'`

which I tested in IE8

It would also be good to add a maze runner test for this. Asserting that the value is one of `landscape` `portrait` or one of the [`ScreenOrientation.type` values](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/type).

packages/plugin-expo-device/device.js
```diff
@@ -20,10 +22,22 @@ module.exports = {
     // get the initial orientation
     updateOrientation()
 
+    // Fetch the free disk space up front because it's an async API so we don't
+    // want to do this in the onError callback. This means the reported free
+    // disk space could be inaccurate as it may change between now and when an
+    // error occurs, however it's unlikely to be drastically different
+    let freeDisk
+    FileSystem.getFreeDiskStorageAsync().then(freeDiskStorage => { freeDisk = freeDiskStorage })
```
bengourley, line 30: This promise has no catch clause so it could cause an unhandled promise rejection.

It needs something like `.catch(er => client._logger.warn('Could not detect free disk space', er)`.

packages/plugin-expo-device/device.js
```diff
@@ -20,10 +22,22 @@ module.exports = {
     // get the initial orientation
     updateOrientation()
 
+    // Fetch the free disk space up front because it's an async API so we don't
+    // want to do this in the onError callback. This means the reported free
```
bengourley, line 26: We _can_ do async `onError` callbacks, but we strongly advise against it. But I think the disk space could quite conceivably fill up after the app has started, so either we need to convert the `onError` and `onSession` callbacks to be `async` OR set an interval to poll the free disk space while the app is running. I'm not sure what a good interval for that timeout would be though.

imjoehaines: Yeah I guess any interval we pick would still be wrong if the thing causing the crash also writes something large to disk

Is this better to leave out for now to avoid inaccurate data/making the `onError` async?

From a look at the source, it doesn't _seem_ like a super expensive API so a low interval may be ok, but that might be deceiving

PR #885, opened 2020-06-16
initial draft of next.js example

djskinner:
This draft PR adds a next.js example project with the aim of identifying best practices and setup for the next.js framework.

It aims to explore some requests raised by users, including:
- providing different environment variables for the server and client
- how to wrap API routes
- how to capture exceptions when code can be "client-rendered, server-rendered, statically generated, statically generated on-the-fly, etc."

# Approach

- Use two projects, one for errors submitted from node and a separate project for errors submitted from the browers.

# Areas for Discussion

## next.js executes code on build

Next actually `requires` each page during the build process. This causes `Bugsnag.start` to be called during build, which does not seem ideal. I haven't yet found a way to prevent code running during the build. The effect of this can be seen from the build output:

<details>
  <summary>Click to expand `npm run build` output</summary>

```
$ NEXT_PUBLIC_BUGSNAG_BROWSER_API_KEY=abc123 BUGSNAG_SERVER_API_KEY=def456 npm run build         

> nextjs-ts-example@0.0.0 build /Users/dan/bugsnag-js/examples/ts/nextjs
> next build

info  - Creating an optimized production build  
info  - Compiled successfully
info  - Collecting page data ..[bugsnag] Loaded!
info  - Collecting page data  
[==  ] info  - Generating static pages (0/2)[bugsnag] Loaded!
[bugsnag] Loaded!
info  - Generating static pages (2/2)
info  - Finalizing page optimization  

Page                                                           Size     First Load JS
┌ ○ /                                                          3.86 kB        78.4 kB
├   /_app                                                      0 B            74.5 kB
├ ○ /404                                                       3.58 kB        78.1 kB
├ λ /api/bad                                                   0 B            74.5 kB
├ λ /api/good                                                  0 B            74.5 kB
└ λ /bad-ssr                                                   354 B          74.9 kB
+ First Load JS shared by all                                  74.5 kB
  ├ chunks/2b0584c9c3b2247265ca83040edbd2ef06237df5.e90230.js  12.2 kB
  ├ chunks/3ef630e34cd10ba68f9d468ac363ff81c534e1e9.78a9f3.js  10.1 kB
  ├ chunks/framework.085e84.js                                 40 kB
  ├ chunks/main.181a33.js                                      6.77 kB
  ├ chunks/pages/_app.dc78e4.js                                4.66 kB
  └ chunks/webpack.ccf5ab.js                                   800 B

λ  (Server)  server-side renders at runtime (uses getInitialProps or getServerSideProps)
○  (Static)  automatically rendered as static HTML (uses no initial props)
●  (SSG)     automatically generated as static HTML + JSON (uses getStaticProps)
   (ISR)     incremental static regeneration (uses revalidate in getStaticProps)

Redirects

┌ source: /:path+/
├ destination: /:path+
└ permanent: true
``` 
</details>

Notice how `[bugsnag] Loaded!` is output three times during the process. I think this is once for each of the two pages plus the custom 404 page though what's really going on under the hood is quite opaque.

## No React tab for SSR render errors

If you hit `/bad-ssr` directly, the example tries to render React on the server but fails due to an exception thrown during render. In this situation no React additional meta data is present in the event.

If you navigate to `/bad-ssr` using the link from `/` the page is rendered on the browser and the resultant exception does include the React `componentStack`.

This is because the `ErrorBoundary` defined in `_app` is not involved in such server side errors, `_error` is the handler.

It's probably not possible to get the component stack in such a situation since React's error boundary handling doesn't seem to be involved.

## Errors caught by `_error` are handled.

Because we have to submit the error with `Bugsnag.notify` it comes through as `handled` when it should be `unhandled`.

We could do:

```
  var event = Event.create(error, true, {
    unhandled: true,
    severity: 'error',
    severityReason: {
      type: 'unknown',
    },
  }, 'unknown', 0);
  
  Bugsnag._notify(event);
```

Is that OK for the example?

## API Routes

- `lib/bugsnag` needs importing in every file as there is no way to centrally import modules for all the routes
- The express plugin can be used to capture additional request information
- Though the setup is fairly complex and needs to be done for each route.

### Error handler not used

It should be noted that the `_error` handler is not used in development mode and so the behaviour is different.

### [bugsnag] Bugsnag.start() was called more than once. Ignoring.

In development mode `[bugsnag] Bugsnag.start() was called more than once. Ignoring.` is output in the console. This is because `lib/bugsnag.ts` [gets evaluated more than once, which is not what you'd normally expect to happen](https://github.com/vercel/next.js/issues/253)). I'm not sure we can do anything about this without referring to internal methods (e.g. checking for _client) to subsequent calls. Even though `lib/bugsnag.ts` is only imported by `_app`, `_app` itself is also loaded multiple times.

# To Do

These areas still need work:

- Setting appVersion
- Source map uploads 

# Out of Scope

 - **using a custom express server**: To keep this example simple, using a custom express server with next.js is not included. For customised setups ideas from multiple examples would need to be combined.

# Test cases

(needs expanding)

- Hit `/bad-ssr` directly
- Navigate to `/bad-/ssr` from `/`
- Hit `/api/good` directly
- Hit `/api/bad` directly
- Send unhandled button
- Send handled button
- Trigger a React render error button
- Send an error from the server


examples/ts/nextjs/lib/bugsnag.js
```diff
@@ -0,0 +1,28 @@
...
+const sharedConfig = {
+  endpoints: {
+    notify: 'http://localhost:8000',
+    sessions: 'http://localhost:10000'
+  }
+}
+
+// TODO work out how to do appVersion
+
+if (typeof window === "undefined") {
```
djskinner, line 15: As discussed, separate projects for each execution environment (node/browser). `@bugsnag/plugin-express` would get pulled into the browser bundle. We could potentially resolve that by splitting into separate files with a conditional `require` or dynamic `import()`

It gets more complicated because for this example we have a custom express server, which also uses this file (hence JS not TS).

examples/ts/nextjs/pages/_app.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react'
+import '../lib/bugsnag'
+import Bugsnag from '@bugsnag/js'
+
+const ErrorBoundary = Bugsnag.getPlugin('react').createErrorBoundary(React);
```
djskinner, line 5: If `React` is not passed here (which is valid API usage) the plugin blows up on the server because we are currently using `window.React` as the default parameter. For this to work properly on node we would need to guard the evaluation of that expression with a `typeof window !== "undefined"` condition.

examples/ts/nextjs/pages/_error.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const error = err || new Error('Unknown error caught by CustomErrorPage.getInitialProps');
+
+  // Ignore 404 errors
+  if (statusCode === 404) {
+    return { statusCode }
+  }
+
+  Bugsnag.notify(error, (event) => {
+    // event.severity = 'error';
+    // event.unhandled = true;
```
djskinner, line 29: IMO these should be unhandled errors, just like if it were to hit the React ErrorBoundary

examples/ts/nextjs/pages/api/bad.ts
```diff
@@ -0,0 +1,36 @@
...
+        return reject(result)
+      }
+
+      return resolve(result)
+    })
+  })
+}
+
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
```
djskinner, line 22: This is an example of wrapping a next.js API route with Bugsnag. It's fairly involved but I'm not sure there are many alternatives as next.js doesn't seem to provide error handling here so you need to catch your own exceptions.

examples/ts/nextjs/pages/api/bad.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Run the middleware
+    // This must be the first piece of middleware in the stack.
+    // It can only capture errors in downstream middleware
+    await runMiddleware(req, res, middleware.requestHandler)
+
+    throw new Error('error in next.js API route');
+
+  } catch (error) {
+    middleware.errorHandler(error, req, res, () => {
+      throw error;
```
djskinner, line 33: otherwise the request hangs

examples/ts/nextjs/server.js
```diff
@@ -0,0 +1,28 @@
...
+var middleware = Bugsnag.getPlugin('express')
+
+app.prepare().then(() => {
+
+  express()
+    .use(middleware.requestHandler)
+    .get('/custom-server/bad', (req, res) => {
+      throw new Error('exception in custom express server (outside next.js)')
+    })
+    .use(middleware.errorHandler)
```
djskinner, line 20: Handle errors in the custom express server before we hit the next.js handler (as that handles it's own errors).

We might want to move this out the example to simplify things a bit and focus on the more common next.js use cases i.e. static and serverless.

lifeiscontent:
@djskinner any updates on getting this merged?

lifeiscontent:
CC: @mattdyoung

djskinner:
@lifeiscontent we don't have an ETA for this landing yet as there are some outstanding issues that warrant further discussion, which I expect we will make progress on in the coming weeks.

lifeiscontent:
@djskinner thanks for the update!

lifeiscontent:
@djskinner any progress?

djskinner:
Thanks to everyone for all the input into this issue. After some investigation, we have decided to revisit updating this example following some planned improvements to the notifier.

The examples we provide are a supplement to the docs and aim to exhibit the simplest integration with the framework. It is infeasible to cover every conceivable configuration (especially so with next.js). Both a custom express.js server and the new API routes are, I think, too niche to be included in the next.js example.

Here are some pointers I found during my investigation that I hope will answer a number of the questions raised in this discussion:

### Custom express server

For a custom express server, the [express integration guide docs](https://docs.bugsnag.com/platforms/javascript/express/) are a good start.

See [here](https://github.com/bugsnag/bugsnag-js/pull/885/commits/271ecca4854de0505ecb30acfbbc55e40b7eaae9#diff-ce50ec388b34427fe73613fbc1ee53889657914b916defe1cf8f421c0ff429dc) for what I came up with during the investigation.

### How to handle providing keys to the Bugsnag client(s) if the Next config doesn’t (or will not shortly) support differentiating between server and client?

[See here](https://github.com/bugsnag/bugsnag-js/pull/885/commits/271ecca4854de0505ecb30acfbbc55e40b7eaae9#diff-cb58cdc09578232bc663f9603270ae032791be57ed98e64841fd8e854b4ebdd7).

### How to properly wrap API routes?

> there doesn’t seem to be a “root” that middleware can be added to, so each API route will probably need to be wrapped with a connect-like middleware?

This is what I found too. [See here](https://github.com/bugsnag/bugsnag-js/pull/885/commits/271ecca4854de0505ecb30acfbbc55e40b7eaae9#diff-6dee61a30cfa87e36ef6cb9cbb11db787741e5e3d1e57f4e59e3bd56312bb5d3). I'm not sure there is a better way. It seems the API routes are designed to be used within the vercel ecosystem where typical middleware functionality is handled external to the next.js application.

### Pages can be client-rendered, server-rendered, statically generated, statically generated on-the-fly, etc. and I’m unsure what or where I need to trap exceptions to handle all of those and how I need to trap them such that the exceptions go to the right Bugsnag app with the right app type

This is where planned improvements to the notifier will help. Right now, if you want to forward errors caught in the next.js error handler, you'll need to make use of the private API. [See here](https://github.com/bugsnag/bugsnag-js/blob/9fe0c2f4d248d3e513d20a010d28f2fd4a4f9f9e/packages/plugin-react/src/index.js#L67-L74) for an example.

In terms of how this can be done in next.js, a combination of [this](https://github.com/bugsnag/bugsnag-js/pull/885/commits/271ecca4854de0505ecb30acfbbc55e40b7eaae9#diff-261d39b84d227a11e9eb70cae177610a37a2383c16e56420bed9fa7201e31522) and [this](https://github.com/bugsnag/bugsnag-js/blob/9fe0c2f4d248d3e513d20a010d28f2fd4a4f9f9e/packages/plugin-react/src/index.js#L67-L74) should do the trick, until the aforementioned notifier improvements are ready.

As pointed out, pages can be "client-rendered, server-rendered, statically generated, statically generated on-the-fly, etc." As with custom express servers and API routes, it might not be feasible to include all of the approaches in a single example. We will endevaour to target the primary use case when we revisit the example.

### What I would like to see though, is an explanation of why this error shows up in the example (or a fix): [bugsnag] Bugsnag.start() was called more than once. Ignoring.

As @lifeiscontent points out, next.js actually executes the code multiple times, including during the build step, and it does not appear to reset modules in between subsequent runs. This means that as next.js runs the `Bugsnag.start` code for a second time (even though you only wrote it once and might expect it to only run once in a normal node/JavaScript environment) the internal `_client` instance is populated from the previous run of it, hence the warning.

More on that in the PR description: https://github.com/bugsnag/bugsnag-js/pull/885

djskinner:
@lifeiscontent I hope the above response is of some use to you. Let me know if you have further questions and I'll try my best to be of assistance.

lifeiscontent:
@djskinner thanks for your write up, your work is very much appreciated. Thank you. On that note, is there a timeline for when the work related to the bugsnag client will be started/completed?

djskinner:
@lifeiscontent this work is something we've earmarked to look at over the coming months, but we don't yet have a firm timeline for release due to a number of competing priorities.

dennisroethig:
Hi 👋 what has happened since September with this? Any updates?

djskinner:
Hi @dennisroethig we have started work on a new example that takes advantage of new notifier features such as being able to mark events as unhandled. It's still a work in progress but it may still be of use to you. See https://github.com/bugsnag/bugsnag-js/pull/1363.

PR #896, opened 2020-06-29
test(react-native): Add event.app tests

bengourley:
Adds component tests for the `app` portion of the payload for react native.

Android-only until test infra is setup on Apple hardware too.


test/react-native/features/app_data.feature
```diff
@@ -0,0 +1,81 @@
+Feature: App data
```
twometresteve, line 1: I think it would be worth adding a check on `the event "unhandled" is` in each scenario.

test/react-native/features/app_data.feature
```diff
@@ -0,0 +1,81 @@
...
+  Then I wait to receive a request
+  And the exception "errorClass" equals "Error"
+  And the exception "message" equals "HandledCaughtError"
+  And the event "app.version" equals "1.2.3"
+  And the event "app.versionCode" equals 1
+  And the event "app.releaseStage" equals "production"
+  And the event "app.inForeground" is true
+  And the event "app.duration" is not null
+  And the event "app.durationInForeground" is not null
+
```
twometresteve, line 14: Looking at the notifier spec, I think I would expect a few more fields to be checked in each scenario?
- binaryArch
- buildUuid (possibly)
- codeBundleId
- id
- type
- inForeground

Some may require a bit more work (e.g. setting in config) or be less easy to predict.

test/react-native/features/app_data.feature
```diff
@@ -0,0 +1,81 @@
...
+  And the exception "message" equals "UnhandledError"
+  And the event "app.version" equals "1.2.3"
+  And the event "app.versionCode" equals 1
+  And the event "app.releaseStage" equals "production"
+  And the event "app.inForeground" is true
+  And the event "app.duration" is not null
+  And the event "app.durationInForeground" is not null
+
+  # Android
+  And the event "app.id" equals "com.reactnative"
```
twometresteve, line 37: You've probably seen already that this fails on v0.55 - possibly an issue for the fixture?

bengourley: Not sure how to resolve. @Cawllec could we get the package name to be `com.reactnative` on both versions? Or do we need to match against multiple strings here.

test/react-native/features/app_data.feature
bengourley: - `binaryArch` is only relevant for native native (NDK/C layer stuff) – we could manually set it to ensure it gets set, but I don't know how useful that is?
- `buildUuid` is generated by our gradle plugin so would have to add that to the fixture app if we wanted it
- `codeBundleId` is straightforward to test – I'll add a case for it
- `id` already tested
- `type` I'll add a case
- `inForegound` already tested

bengourley:
Ready for re-review, but also pending a solution to the differing app ids on fixtures – tests on 0.55 will fail due to this.

test/react-native/features/app_data.feature
twometresteve: If the answer to this isn't simple and in the interest of getting the PR merged, I would be happy to see this check commented out in the short term, with a reference to an internal ticket to be actioned before release.

test/react-native/features/fixtures/android-module/BugsnagModule.java
```diff
@@ -44,6 +49,38 @@ private Configuration createConfiguration(ReadableMap options) {
...
+      }
+
+      try {
+        Set<String> enabledReleaseStages = new HashSet<String>();
+        ReadableArray ar = options.getArray("enabledReleaseStages");
+        for (int i = 0; i < ar.size(); i++) enabledReleaseStages.add(ar.getString(i));
+        config.setEnabledReleaseStages(enabledReleaseStages);
+      } catch (NoSuchKeyException e) {
+        // ignore NoSuchKeyException
+      }
```
bengourley, line 82: definitely a cleaner way to do this…

test/react-native/features/app.feature
```diff
@@ -0,0 +1,118 @@
...
+Scenario: Setting releaseStage and enabledReleaseStages to enable delivery
+  When I run "AppConfigEnabledReleaseStagesScenario"
+  Then I wait to receive a request
+  And the exception "errorClass" equals "Error"
+  And the exception "message" equals "AppConfigEnabledReleaseStagesScenario"
+  And the event "unhandled" is false
+  And the event "app.releaseStage" equals "preprod"
+
+Scenario: Setting releaseStage and enabledReleaseStages to disable delivery
+  When I run "AppConfigEnabledReleaseStagesNoSendScenario"
```
bengourley, line 116: Ooops, this scenario is not defined, but it doesn't fail the tests, since it asserts that no error reports are delivered. Is there a good way to catch this?

twometresteve: I think it adds weight to crafting some form of positive check in scenarios, which we continue to discuss.  We have a few ideas, including inspecting device logs for "Bugsnag loaded" and adding more logging, but it involves non-trivial updates to Maze Runner.

For this particular issue (and perhaps others), I wonder if we could use our own Bugsnag.notify in the test fixture - so that a request is sent, failing the test.  That and reviewer diligence, of course.

test/react-native/features/fixtures/android-module/BugsnagModule.java
twometresteve: I'm generally in favour checking for existence first rather than ignoring exceptions, but the interface doesn't provide that - `ReadableMap` isn't the friendliest class in the world.

We could consider using its toHashMap() method first, although that has its own issues (e.g. casting) - or write a helper class to at least turn each block into one line.  I guess it depends how much we anticipate doing this Vs the extra effort.

test/react-native/features/app.feature
bengourley: ahh good idea, so in [`App.js`](https://github.com/bugsnag/bugsnag-js/pull/896/files#diff-d7f5039e8157e9d11859552ad734f896) we could do something like:

```diff
+ let Scenario = Scenarios[scenarioName]
+ if (!Scenario) {
+   Bugsnag.start(configuration)
+   Bugsnag.notify(new Error('Missing test scenario'))
+   return
+ }
- let scenario = new Scenarios[scenarioName](configuration, scenarioMetaData, jsConfig)
```

bengourley: would just have to be diligent making sure we always check the error messages for all errors received, otherwise we risk a different kind of false positive.

bengourley:
>  Seeing as we can't merge it yet anyway, can we let the dust settle on the target branch to resolve the test failure and the missing scenario to go in?

Not sure what you mean – the tests are passing and the missing scenario is added.

If there is something preventing it from being mergeable, that's fine but it might be a bit of a bottleneck – I'm basing the subsequent tests off of some of the tidying up I did on this branch. The alternative is that we land this and make a note to resolve the known issues.

test/react-native/features/app.feature
twometresteve: Yes, absolutely, that did also occur to me.  With a bit of thought we could potentially take it further by introducing some kind of stylised error message for Maze Runner to identify and immediately fail on.

test/react-native/features/fixtures/android-module/BugsnagModule.java
bengourley: I did look at factoring this out into generic methods for each type, but it didn't really end up much cleaner

```java
String appVersion = extractString("appVersion")
if (appVersion != null) {
  configuration.setAppVersion(appVersion)
}

String releaseStage = extractString("releaseStage")
if (releaseStage != null) {
  configuration.setReleaseStage(releaseStage)
}

Set<String> enabledReleaseStages = extractStringSet("enabledReleaseStages")
if (enabledReleaseStages != null) {
  configuration.setEnabledReleaseStages(enabledReleaseStages)
}

// ....

String extractString(String k, ReadableMap opts) {
  try {
    return opts.getString(k);
  } catch (NoSuchKeyException e) {
    return null
  }
}

Set<String> extractStringSet(String k, ReadableMap opts) {
  //...
}
```

twometresteve: Yeah, just leave it as it is, I think we're in danger of over-thinking it.

test/react-native/features/handled.feature
```diff
@@ -1,13 +1,13 @@
 Feature: Reporting handled errors
 
 Scenario: Calling notify() with a caught Error
-  When I run "HandledCaughtErrorScenario"
+  When I run "HandledJsErrorScenario"
   Then I wait to receive a request
   And the exception "errorClass" equals "Error"
-  And the exception "message" equals "HandledCaughtError"
+  And the exception "message" equals "HandledJsErrorScenario"
```
twometresteve, line 7: Bordering on repetitive, perhaps, but we could add checked on "unhandled" in this file and `unhandled.feature` too.

test/react-native/features/fixtures/app/scenarios/UnhandledNativeErrorScenario.js
```diff
@@ -1,8 +1,8 @@
...
 
-export default class UnhandledNativeErrorScenario extends Scenario {
-    run() {
-        NativeModules.BugsnagTestInterface.runScenario("UnhandledExceptionScenario", () => {})
-    }
-}
\ No newline at end of file
+export class UnhandledNativeErrorScenario extends Scenario {
+  run() {
+    NativeModules.BugsnagTestInterface.runScenario('UnhandledNativeErrorScenario', () => {})
```
twometresteve, line 6: Just a thought for a future changes, we could grab the scenario name from the class name here to avoid typos or copy/paste errors creeping in.

twometresteve:
> > Seeing as we can't merge it yet anyway, can we let the dust settle on the target branch to resolve the test failure and the missing scenario to go in?
> 
> Not sure what you mean – the tests are passing and the missing scenario is added.
> 
> If there is something preventing it from being mergeable, that's fine but it might be a bit of a bottleneck – I'm basing the subsequent tests off of some of the tidying up I did on this branch. The alternative is that we land this and make a note to resolve the known issues.

Sorry, just lot track of where we were for a moment.  I've now annotated a couple of internal tickets with actions from this to catch up on later.

PR #897, opened 2020-06-29
Refactor and tidy React Native example

twometresteve:
The main purpose of this change is to move the existing React Native example into a folder named after the version of RN that it uses, leaving us room to add further examples for different versions of RN.  In particular, so that we can add an example for RN 0.60+ that uses Cocoapods.

As well as the refactor, I have also removed the tvOS targets.  tvOS is not official supported and its presence may be misleading or confusing.

There is a known issue with running the example on a simulator, due to where XCode places DerivedData.  This is not made any better or worse by this change and will be addressed separately.  I have successfully run this change on a real device by opening the XCode project.


examples/reactnative/rn_0.59/README.md
```diff
@@ -1,6 +1,6 @@
 # React Native
 
-This is an example project showing how to use `@bugsnag/js` with a React Native project.
+This is an example project showing how to use `@bugsnag/js` with a React Native v0.59 project.
```
bengourley, line 3:
```suggestion
This is an example project showing how to use `@bugsnag/react-native` with a React Native v0.59 project.
```

PR #902, opened 2020-07-02
RN: breadcrumb tests

bengourley:
Inital set of tests for automatic/manual breadcrumbs.

Includes a couple of bug fixes that were spotted due to the new tests.
- bugsnag's own logs in breadcrumbs
- error appearing as a breadcrumb for itself


test/react-native/features/fixtures/android-module/scenarios/BreadcrumbsNativeManualScenario.kt
```diff
@@ -0,0 +1,13 @@
+package com.<ANDROID_PACKAGE_PATH>.scenarios
```
twometresteve, line 1: If you rebase the branch now that the pipeline refactor has gone in, you'll see that this package name can (needs to) now be explicitly set.

PR #903, opened 2020-07-03
Refactor RN test fixture build process

twometresteve:
This is the first of a few iterations to improve the build time and Docker caching ability for the React Native test fixtures.

In general, I have attempted to carefully craft the Dockerfile, cognisant of which files are being copied and which files are likely to change during developer workflow, which would lead to the cache being invalidated.

The main change here is to bring the initial elements of the Gradle process (downloading the Gradle zip) much earlier in the Docker build process and before most source files are copied in, which would invalidate the cache and cause the process to be repeated.  It does this with the help of a collection of Node scripts that detects all folders that contain a RN test fixture and "gathers" the few initial files on the host that are necessary to run the initial gradle command.  It then copies just these into the Docker image.

I have also simplified aspects of the build.  For example, it is no longer necessary to find and replace the package names.

Proposed changes for futures PRs:
- Use a multi-stage build to run the gathering script rather than a separate plugin and NPM install
- Consider bringing the NPM install and full Gradle build into the Docker build (but at some point doing this you lose the ability to parallelise the building of test fixtures for multiple RN versions).


test/react-native/scripts/build-android-locally.js
```diff
@@ -0,0 +1,22 @@
...
+requiredVars.forEach((envvar) => {
+  if (!process.env[envvar]) {
+    console.error(`Environment ${envvar} must be set`)
+    allPresent = false
+  }
+})
+if (!allPresent) return
+
+common.run('rm -rf docker-temp')
+common.run('mkdir -p docker-temp/fixtures')
```
bengourley, line 18: can you add this path to `.gitignore`, or put it in a system-wide tempdir so that it gets cleaned up by the OS?

Node docs for [`os.tmpdir()`](https://nodejs.org/dist/latest-v12.x/docs/api/os.html#os_os_tmpdir) and [`fs.mkdtemp()`](https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_mkdtemp_prefix_options_callback) if you want to go the system route.

twometresteve: I've already added it to `.gitignore`, but we might find the need for it to disappear altogether in future iterations.  I'm not particular fond of the temporary directory myself, but if we can keep it as is and I'll either remove it or change to an OS tempdir next time around if possible.

PR #913, opened 2020-07-08
test(react-native): Add tests for context

bengourley:
Setting context in various places, config, client, onError callback etc.


bengourley:
Hmm, seems to be a timing issue with this one https://buildkite.com/bugsnag/at-bugsnag-react-native/builds/317#8b674038-a67e-4f37-a479-9d83f778bf8d/419-426

PR #914, opened 2020-07-08
Simplify android maven repo setup for React Native

fractalwrench:
## Goal

Simplifies the setup of the maven repository for React Native.

Instead of the user having to specify a maven repository to find bugsnag's AAR files, they only have to apply a gradle file from within the node module. This allows for extra Gradle configuration to be added at a future date.

## Tests

Verified the change in the example app.


examples/reactnative/rn_0.59/android/app/build.gradle
```diff
@@ -77,6 +77,7 @@ project.ext.react = [
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
+apply from: "../../node_modules/@bugsnag/react-native/react.gradle"
```
tomlongridge, line 80: Is the name `react.gradle` a bit misleading? Perhaps we could go with `bugsnag.gradle`, `bugsnag-react-native.gradle` or `build.gradle`.

fractalwrench: happy with any of those - any particular favourite?

PR #915, opened 2020-07-08
Avoid needing to set header search paths manually

fractalwrench:
## Goal

Avoids the need for users to manually alter their header search path when using RN 0.59 or below.

## Changeset

Public header files are copied into the `ios/BugsnagReactNative/Bugsnag` directory, which is a directory that `react-native link` picks up by default. The podspec has been altered so that the public headers are found from this directory.

## Tests

Verified an example app compiled for RN 0.59 and RN 0.62


packages/react-native/BugsnagReactNative.podspec
```diff
@@ -13,9 +13,9 @@ Pod::Spec.new do |s|
   s.author       = { "Bugsnag" => "platforms@bugsnag.com" }
   s.platform     = :ios, "7.0"
   s.source       = { :git => "https://github.com/bugsnag/bugsnag-js.git", :tag => "v#{s.version}" }
-  s.source_files = "ios/BugsnagReactNative/**/*.{h,m}",
```
bengourley, line 16: Does this need to change? The glob as it was would ensure if we put anything in a subdirectory it would be picked up, so future proofing this file.

packages/react-native/BugsnagReactNative.podspec
```diff
@@ -13,9 +13,9 @@ Pod::Spec.new do |s|
...
+  s.source_files = "ios/BugsnagReactNative/*.{h,m}",
                    "ios/vendor/bugsnag-cocoa/**/*.{h,mm,m,cpp,c}",
-  s.public_header_files = "ios/vendor/bugsnag-cocoa/{#{bugsnag_cocoa_public_header_files.join(',')}}"
+  s.public_header_files = "ios/BugsnagReactNative/Bugsnag/*.{h}"
   s.header_dir = 'Bugsnag'
```
bengourley, line 19: did you try with this line removed? It seems somewhat off the beaten path, so would be good if we can remove it. I can't think what problem it solved, and since you've brought in a `Bugsnag/` directory, it might not be required anymore.

packages/react-native/BugsnagReactNative.podspec
fractalwrench: yep - otherwise the header files could be included twice as they're included in `ios/vendor/bugsnag-cocoa`

PR #918, opened 2020-07-09
test(react-native): Add device tests

bengourley:
Adds some basic payload assertions on the `device` section for handled/unhandled native and JS errors.


test/react-native/features/device.feature
```diff
@@ -0,0 +1,100 @@
...
+  And the event "device.jailbroken" is false
+  And the event "device.locale" equals "en_US"
+  And the event "device.manufacturer" is not null
+  And the event "device.model" is not null
+  And the event "device.osName" equals "android"
+  And the event "device.osVersion" matches "^\d+$"
+  And the event "device.runtimeVersions.androidApiLevel" is not null
+  And the event "device.runtimeVersions.reactNative" matches "^\d+\.\d+\.\d+$"
+  And the event "device.runtimeVersions.reactNativeJsEngine" matches "jsc"
+  And the event "device.totalMemory" is not null
```
twometresteve, line 21: I imagine this is the best we can do currently, but I would be concerned about the zero case.  Perhaps we could raise an internal ticket to add a new step along the lines of "and the event {string} is a number greater than {int}" to tighten this up in the future.

test/react-native/features/device.feature
```diff
@@ -0,0 +1,100 @@
...
+  And the exception "message" equals "DeviceJsUnhandledScenario"
+  And the event "unhandled" is true
+
+  # Android
+  And the event "device.id" is not null
+  And the event "device.jailbroken" is false
+  And the event "device.locale" equals "en_US"
+  And the event "device.manufacturer" is not null
+  And the event "device.model" is not null
+  And the event "device.osName" equals "android"
```
twometresteve, line 41: How's this and other lines going to integrate with the iOS work?  Do we have the tools in place now to handle it before merging, even if we don't add the iOS values just yet?

test/react-native/features/device.feature
bengourley: Yes, best we can do at present. I was on the cusp of making a variety of maze runner feature requests…

- `and the event {string} is an integer greater than {int}`
- `and the event {string} is an integer between {int} and {int}` (useful for `apiLevel`)
- `and the event {string} is a number` (general case where the number is allowed to be anything, but more type-guarantee than simply `not null`

test/react-native/features/device.feature
bengourley: as I understand it, @Cawllec is going to add the mechanism for this so can be merged as-is…

test/react-native/features/device.feature
twometresteve: No problem, just send the ticket reference our way.

test/react-native/features/device.feature
Cawllec: Yes, I'll handle this

PR #923, opened 2020-07-10
Improve test resilience

twometresteve:
1. Clear text fields before sending keys.  
2. Be prepared to wait longer (60s from default of 15s) before trying to enter text into the first field.


twometresteve:
> Seems to be a problem? https://buildkite.com/bugsnag/at-bugsnag-react-native/builds/337#afed6e43-55a9-4439-b727-8db3ef0b17e9/147-150

MazeRunner fixed and tests now passes (current known flake aside).

PR #927, opened 2020-07-13
React Native 0.63 example

twometresteve:
Adds a new example for React Native 0.63 replacing the example for 0.59, which is no longer supported.

Github has possibly been more helpful than needed and detected that a number of files have been moved and changed, despite me having performed the delete separately and after adding the new files.

I have tested the example on both Android 5 and iOS 10 devices, noting that React Native 0.63 itself does not support iOS 9.3 (our current earliest support iOS version).


examples/reactnative/rn_0.59/README.md
```diff
@@ -1,46 +0,0 @@
-# React Native
```
fractalwrench, line 1: It'd be worth copying this over to the 0.63 project and adapting where necessary

examples/reactnative/rn063example/android/app/build.gradle
```diff
@@ -73,11 +78,10 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "index.js"
+    enableHermes: false,  // clean and rebuild if changing
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
-apply from: "../../node_modules/@bugsnag/react-native/bugsnag-react-native.gradle"
```
fractalwrench, line 80: this is required to resolve the AAR dependency on Android

examples/reactnative/rn063example/android/app/src/main/java/com/rn063example/CrashyModule.java
```diff
@@ -1,4 +1,4 @@
-package com.reactnative;
+package com.rn063example;
 
 import com.facebook.react.bridge.ReactApplicationContext;
```
twometresteve, line 3: handledError() is missing from this class.

examples/reactnative/rn_0.59/README.md
twometresteve: Absolutely.  I actually meant to and forgot.

examples/reactnative/rn063example/android/app/build.gradle
twometresteve: It seemed to work fine without, but more than happy to add - especially if it's something that can fail intermittently.

twometresteve: Ok, it's coming in the next release.  I've added this and bumped the dependency to the next alpha release.  CI will fail initially, but once the alpha is created it should pass on retry.

examples/reactnative/rn063example/android/app/src/main/java/com/rn063example/CrashyModule.java
twometresteve: Now added.

PR #929, opened 2020-07-13
test: convert tests to jest

djskinner:
Convert the following packages' test files to TypeScript and jest:
- `plugin-node-in-project`
- `plugin-node-device`
- `plugin-node-surrounding-code`
- `plugin-node-uncaught-exception`

### DIFF
--- a/packages/plugin-node-surrounding-code/test/surrounding-code.test.ts
+++ b/packages/plugin-node-surrounding-code/test/surrounding-code.test.ts
@@ -1,17 +1,16 @@
-const { describe, it, expect } = global
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import fs from 'fs'
+import plugin from '../'
+import { join } from 'path'
+import Event from '@bugsnag/core/event'
+import Client from '@bugsnag/core/client'
 
-const fs = require('fs')
 let createReadStreamCount = 0
 const originalReadStream = fs.createReadStream
 fs.createReadStream = function () {
   createReadStreamCount++
-  return originalReadStream.apply(fs, arguments)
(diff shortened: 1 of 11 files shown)


packages/plugin-node-surrounding-code/test/surrounding-code.test.ts
```diff
@@ -1,38 +1,37 @@
-const { describe, it, expect } = global
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
djskinner, line 1: I wonder if it's worth disabling this rule across all test files as its usage is often legitimate and useful.

Often there are cases where a previous assertion asserts the existence of something. Unfortunately the typings of `expect` aren't sufficient to avail the TS compiler of such knowledge (perhaps it is possible to do with better typings that use assertion signatures - I'm not sure). Even without the benefit of previous assertions, it seems legitimate to me to make assumptions about the existence of things because the test would just fail otherwise.

FWIW we agreed that it would be OK disable this rule in the dashboard-js project though we haven't got round to doing it yet. That might be because in that project we are more relaxed about overriding rules on a per-line or entire file basis where judged appropriate.

On a related note I've seen a few places in the tests where the optional chaining `?.` operator would be useful. Though currently jest doesn't like it when used. I think we might need to update our TypeScript/jest/babel dependencies if we want to make use of that one.

bengourley: I think it's a reasonable thing to assert. Since the implementation is JS, the type system doesn't guarantee that such things are genuinely non-null.

Maybe updating the test dev dependencies should come in via another PR. The `?.` and also `!` are nice operators for terse TS so it would be good to have access to them.

PR #931, opened 2020-07-13
RN: Add metadata tests

bengourley:


test/react-native/features/metadata.feature
```diff
@@ -0,0 +1,19 @@
...
+  And the event "metaData.jsdata.some_data" equals "set via config"
+  And the event "metaData.jsdata.some_more_data" equals "set via client"
+  And the event "metaData.jsdata.even_more_data" equals "set via event"
+
+Scenario: Setting metadata (native)
+  When I run "MetadataNativeScenario"
+  Then I wait to receive a request
+  And the exception "errorClass" equals "java.lang.RuntimeException"
+  And the exception "message" equals "MetadataNativeScenario"
+  # And the event "metaData.nativedata.some_data" equals "set via config"
```
bengourley, line 17: Need to add metadata parsing to the native app in order to pass this through to the scenario.

Cawllec: I'll prioritise this feature as the next to be picked up, and look at adding metadata configuration for both Android and iOS

Cawllec:
In addition we should probably add an unhandled native check, due to past issues with metadata not being synced to the native crash handler.

PR #938, opened 2020-07-17
React-native iOS test fixture

Cawllec:
Adds a test fixture and appropriate pipeline to run react-native end-to-end tests against Cocoa targets.

Makes a couple of changes that may effect running things locally:
- The `build` method in `scripts/react-native-helper.js` has been renamed to `buildAndroid`
- A step has been added `And the payload fields match one of the following sets:` to allow testing across multiple payloads where the order they are received may be an issue.


test/react-native/features/app.feature
```diff
@@ -30,63 +34,84 @@ Scenario: Unhandled JS error
   And the exception "message" equals "AppJsUnhandledScenario"
   And the event "unhandled" is true
   And the event "app.version" equals "1.2.3"
-  And the event "app.versionCode" equals 1
+  # Parameter not present on iOS devices
```
bengourley, line 37: I do see this for iOS events on the dashboard, though it may only be included in a release build?

image

test/react-native/features/steps/react-native-steps.rb
```diff
@@ -55,3 +55,12 @@
   valid_values = table.raw.flatten
   assert_true(valid_values.include?(actual_value), "#{field_path} value: #{actual_value} did not match the given list: #{valid_values}")
 end
+
+Then("the payload fields match one of the following sets:") do |data_table|
```
bengourley, line 59: I think moving to this step loses us some assurances compared with the individual assertions before:

- sometimes the order is important
- we now no longer are guaranteed that every entry in the table was used – the same input could have been received each time and would still pass

I think if we were to use something like this, it would be better to collect all the requests and say "given assertions xyz, do I have a request that matches each?"

test/react-native/features/app.feature
Cawllec: I understood from [here](https://github.com/bugsnag/platforms-notifier-spec/blob/master/app.md#versioncode-1) that it was Android specific.  @fractalwrench do you know if that's the case?

fractalwrench: The versionCode does appear to be populated on iOS events so the iOS implementation and spec are inconsistent.

test/react-native/features/steps/react-native-steps.rb
```diff
@@ -55,3 +55,27 @@
...
```
bengourley, line 59: I think this should now be removed since it's unused.

PR #943, opened 2020-07-21
Re-enable temporarily disabled linting and resolve violations early

twometresteve:


Cawllec:
It looks like we need to re-enable the first stage in the `pipeline.yml` file, as that outputs the `min_packages.tar` that the lint stage depends on.

twometresteve:
> It looks like we need to re-enable the first stage in the `pipeline.yml` file, as that outputs the `min_packages.tar` that the lint stage depends on.

Great, thanks - I'd been looking at that and couldn't immediately see why it was depending on something.

PR #944, opened 2020-07-21
RN: Native notifier updates

bengourley:
- Updates both native notifiers to include latest fixes
- Update JS event-sync module to use the provided app/device metadata from android


packages/react-native/ios/vendor/bugsnag-cocoa/Bugsnag.podspec.json
```diff
@@ -9,7 +9,7 @@
   },
   "source": {
     "git": "https://github.com/bugsnag/bugsnag-cocoa.git",
-    "tag": "v6.1.0"
+    "tag": "v6.1.1"
```
twometresteve, line 12: 6.1.2 is imminent - did you want to handle that separately?

bengourley: I landed the changes from the `react-native-plugin` integration branch, are the 6.1.2 changes not on that?

twometresteve: Sorry, got the wrong end of the stick as to how the update script works.

twometresteve:
We have a ticket to investigate and fix the UnhandledJsPromiseRejectionScenario failure, so I'm happy for this to be force-merged.

bengourley:
> We have a ticket to investigate and fix the UnhandledJsPromiseRejectionScenario failure, so I'm happy for this to be force-merged.

Oh – I just pushed a commit that ensures that scenario restarts, in case it does crash the app.

PR #945, opened 2020-07-21
Leave working tree clean when running tests

twometresteve:
Delete and ignore NPM/Pod lock files for the React Native test fixtures to avoid dirtying the working tree when running tests.


twometresteve:
> Did you look at invoking `npm` with the `--no-package-lock` option, or see if cocoapods has an equivalent?

I tried with `--no-package-lock` in the first instance and the build failed (but on reflection that might have been due to poor internet).  I looked at the Cocoapods options and couldn't see one, which is what then led me to this approach - so that it was consistent across the two technologies.

Cawllec:
As an additional change I would add:
```
git checkout test/react-native/features/fixtures/rn*/package.json test/react-native/features/fixtures/rn*/ios/Podfile.lock
git clean -xfdf test/react-native/features/fixtures/rn*
```
Somewhere in the build scripts after Android/iOS has built.  At the moment if you wish to do repeated builds for iOS it errors if you don't clean those changes/files up.

PR #948, opened 2020-07-22
test(plugin-browser): convert type tests to jest

djskinner:
Converts the type tests being run by tsc to basic jest integration tests that aim to replicate the type checking of the previous tests.


bengourley:
This looks good in terms of testing all the things the previous set up did.

My only concern is it now appears to make network requests, unless I'm missing some mocking or configuration somewhere?

If this is the case then I think it should be updated to either point to an endpoint on the local server or to mock/stub out the network calls made by the delivery layer, whichever is most straightforward to implement and least confusing for future travellers.

djskinner:
> My only concern is it now appears to make network requests

I've gone with mocking XHR which allows us to make assertions too. It makes for a fairly decent (if minimal) set of integration tests.

PR #951, opened 2020-07-23
docs(example): add missing pod install step for React Native example

twometresteve:
Details the required pod install step, which was previously missing having moved the example from RN 0.59.

### DIFF
--- a/examples/reactnative/rn063example/README.md
+++ b/examples/reactnative/rn063example/README.md
@@ -31,10 +31,10 @@ This project was bootstrapped with `react-native init`.
 
     The API key can be found in the Bugsnag settings for your project.
 
-1. For iOS only, install the project dependencies using: 
+1. For iOS only, install the project dependencies from the `ios/` directory using: 
 
 ```
-pod install --project-directory=ios
+pod install 
 ```
 
 1. Run the app on either Android or iOS:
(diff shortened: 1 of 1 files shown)


examples/reactnative/rn063example/README.md
```diff
@@ -31,6 +31,12 @@ This project was bootstrapped with `react-native init`.
 
     The API key can be found in the Bugsnag settings for your project.
 
+1. For iOS only, install the project dependencies using:
```
tomlongridge, line 34: My preference would be to follow the docs and suggest they `cd` into the directory and run `pod install`. It's a bit less magic and more understandable.

Unless @bengourley feels otherwise, in which case we should update the docs.

twometresteve: The reason why I didn't was because the next line (I believe) relies on being in the directory above - so it would need two extra statements to be added.  Happy to change it if we're happy with that though.

tomlongridge: That seems like overkill and I think would be obvious to a RN developer. How about making it less explicit?

```suggestion
1. For iOS only, install the project dependencies from the `ios/` directory using: 
```

twometresteve: Wording changed as suggested and option removed from the command.

PR #952, opened 2020-07-23
chore(example): add buildUUID to RN example

twometresteve:
Demonstrates displaying the buildUUID in the dashboard (free by-product of manual testing).


examples/reactnative/rn063example/android/app/src/main/AndroidManifest.xml
```diff
@@ -26,5 +26,7 @@
                  android:value="0abcdef000000abcdef000000abcdef0"/>
       <meta-data android:name="com.bugsnag.android.DETECT_NDK_CRASHES"
                  android:value="true"/>
+        <meta-data android:name="com.bugsnag.android.BUILD_UUID"
+                   android:value="fc217a04-99f7-4cc4-8585-51ea769f501b"/>
```
tomlongridge, line 30: I think this could do with an explanatory comment if we're going to include it in the example.

Something like:
```
<!-- Generated by the Bugsnag Gradle Plugin: see https://docs.bugsnag.com/build-integrations/gradle/#build-uuids -->
```

cc @fractalwrench - that's correct isn't it?

twometresteve: Happy to add a comment.

PR #957, opened 2020-07-24
docs(react-native): Upgrade guide

bengourley:
Upgrade guide for `bugsnag-react-native` to `@bugsnag/react-native`.

### DIFF
--- a/packages/react-native/UPGRADING.md
+++ b/packages/react-native/UPGRADING.md
@@ -4,7 +4,7 @@
 
 As of `v7.3` of the [`bugsnag-js` monorepo](https://github.com/bugsnag/bugsnag-js) it contains Bugsnag's SDK for React Native. This additional notifier joins `@bugsnag/js` and `@bugsnag/expo` in its unified version scheme, so the first version of `@bugsnag/react-native` is `v7.3.0`.
 
-The previous React Native SDK – [`bugsnag-react-native`](https://github.com/bugsnag/bugsnag-react-native) currently at `v2.23` – continues to be available and will receive critical bug fixes, but it is no longer under active development and won't receive new features.
+The previous React Native SDK – [`bugsnag-react-native`](https://github.com/bugsnag/bugsnag-react-native) – continues to be available and will receive critical bug fixes, but it is no longer under active development and won't receive new features.
 
 `@bugsnag/react-native` is a complete rewrite and has no backwards compatibility with `bugsnag-react-native`. Therefore, the upgrade steps essentially consist of these three steps:
 
--- a/packages/react-native/UPGRADING.md
+++ b/packages/react-native/UPGRADING.md
@@ -82,7 +82,7 @@ Previously in `bugsnag-react-native` it was possible to initialize in a variety 
 
 In each of your native projects you need to initialize Bugsnag. This was optional before, but it means that now crashes before the JS layer loads will always be caught.
 
-To facilitate this, and to get Bugsnag started as early as possible, most configuration can only be done in the native layer. Some JS-specific options are allowed in the JS layer.
+To facilitate this, and to get Bugsnag started as early as possible, most configuration can only be done in the iOS and Android projects. Some JS-specific options are allowed in the JS layer.
 
 ##### iOS
 
--- a/packages/react-native/UPGRADING.md
+++ b/packages/react-native/UPGRADING.md
@@ -12,11 +12,9 @@ The previous React Native SDK – [`bugsnag-react-native`](https://github.com/bu
 2. adding `@bugsnag/react-native` to your project
 3. updating any usage of Bugsnag throughout your codebase
 
-#### A quick note on supported versions of React Native
-
 The oldest version of React Native supported by `@bugsnag/react-native@7.3` is `0.60`. If your project is running an older version than this, you should upgrade that first, or continue to use `bugsnag-react-native` until such time as you can upgrade.
 
-### Removing `bugsnag-react-native`
+## Removing `bugsnag-react-native`
 
 Since `@bugsnag/react-native` is an entirely new module, `bugsnag-react-native` should be removed first. It's not possible to run them both alongside one-another, so the replacement can't be done incrementally.
 
@@ -41,7 +39,7 @@ For iOS, open up the project in XCode `open ios/<ProjectName>.xcworkspace`, and 
 
 For Android, run `cd android && ./gradlew clean`.
 
-### Adding `@bugsnag/react-native`
+## Adding `@bugsnag/react-native`
 
 Full instructions are available at the [React Native integration guide](https://docs.bugsnag.com/platforms/react-native/react-native#installation) but here is a summary.
 
@@ -53,7 +51,7 @@ npm install --save @bugsnag/react-native
 yarn add @bugsnag/react-native
 ```
 
-#### Android
+### Android
 
 Add the following to your `android/app/build.gradle` to integrate Bugsnag into your Anroid project:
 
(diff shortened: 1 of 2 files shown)


packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+
+```
+- <key>BugsnagAPIKey</key>
+- <string>YOUR_API_KEY</string>
++ <key>bugsnag</key>
++ <dict>
++     <key>apiKey</key>
++     <string>YOUR_API_KEY</string>
++ </dict>
+```
```
tomlongridge, line 127: We want to emphasise use of the plist/manifest as much as possible - can you move this to after the `[Bugsnag start]` snippet and reference the new ability to set other configuration options? Then have the in-code snippet afterwards.

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+##### Android
+
+Replace the following import which will likely be in your `MainApplication.java` file:
+
+```diff
+- import com.bugsnag.android.BugsnagReactNative;
++ import com.bugsnag.android.Bugsnag;
+```
+
+Inside your application's `onCreate` method, replace any Bugsnag initialization code with the updated version:
```
tomlongridge, line 138: As above, it would be good to reference the file-loading config mechanism here.

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+- config.setReleaseStage("beta");
+- BugsnagReactNative.startWithConfiguration(this /* app context */, config);
++ Configuration config = Configuration.load(this);
++ config.setReleaseStage("beta");
+Bugsnag.start(this /* app context */, config);
+```
+
+###### JS
+
+The JS interface has been updated to match our new notifier spec standard – that means it looks the same as our JS notifier, and uses the same names and concepts as our other platforms.
```
tomlongridge, line 151: Do we want to refer to the spec?

```suggestion
As part of the `bugsnag-js monorepo`, the JS interface now looks the same as our JS notifier, and uses the same names and concepts as our other platforms.
```

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+
+When the JS layer loads, __the native layers and notifiers must already have been configured__. It you attempt to initialize the JS layer without having first configured the native layer, it __will not work__. This is new behaviour – previously we allowed configuration of the native layer to be done lazily and automatically, but this meant native errors on startup could be missed.
```
tomlongridge, line 153:
```suggestion
When the JS layer loads, __the native iOS/Android Bugsnag client must already have been configured__. If you attempt to initialize the JS layer without having first configured the native layer, it __will not work__. This is new behavior – previously we allowed configuration of the native layer to be done lazily and automatically, but this meant native errors on startup could be missed.
```

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+
+Since the JS layer hooks in to an already running native layer, there are fewer configuration options available. Most configuration must be done in the native layer, with only a small subset of JS-specific options allowed in the JS layer.
```
tomlongridge, line 155: Move underneath proceeding snippet as it's referring to configuration, not starting?

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+###### Handled errors
+
+To report a handled error call the `notify()` method:
+
+```diff
++ client.notify(new Error('uh oh'))
++ Bugsnag.notify(new Error('uh oh'))
+```
+
+The second argument is an `onError` callback which will receive the error report that will be sent to Bugsnag. This is similar, but previously the data structure was a `report` – now it is an `event`. See the [Customizing error reports](https://docs.bugsnag.com/platforms/react-native/react-native/customizing-error-reports/#updating-events-using-callbacks) for full information. An example of  adding metadata is shown below:
```
tomlongridge, line 210:
```suggestion
The second argument is an `onError` callback which will receive the error report that will be sent to Bugsnag. This is similar to the previous syntax, but previously the data structure was a `report` – now it is an `event`. See the [Customizing error reports](https://docs.bugsnag.com/platforms/react-native/react-native/customizing-error-reports/#updating-events-using-callbacks) for full information. An example of adding metadata to an event is shown below:
```

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+
+```diff
+- configuration.registerBeforeSendCallback(fn)
+- configuration.unregisterBeforeSendCallback(fn)
++ Bugsnag.start({ onError: fn })
+  // or
++ Bugsnag.addOnError(fn)
++ Bugsnag.removeOnError(fn)
+```
+
```
tomlongridge, line 234: Add a quick sentence to go with the snippet and link to docs?

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+```diff
+- client.startSession()
+- client.pauseSession()
+- client.resumeSession()
++ Bugsnag.startSession()
++ Bugsnag.pauseSession()
++ Bugsnag.resumeSession()
+```
+
```
tomlongridge, line 254: Add a quick sentence to go with the snippet and link to docs?

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+###### Adding user information
+
+```diff
+- client.setUser('1234', 'Jessica Jones', 'jess@example.com')
++ Bugsnag.setUser('1234', 'jess@example.com', 'Jessica Jones')
+```
+
+__Note: the argument order has changed__ from `id,name,email` to `id,email,name`
+
```
tomlongridge, line 243: Add a quick sentence to go with the snippet and link to docs?

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
++ Bugsnag.start({
++   codeBundleId: '1.2.3-r091283'
++ })
+```
+
+See the [React Native configuration options page](https//docs.bugsnag.com/platforms/react-native/react-native/configuration-options) for reference.
+
+#### General usage
+
+##### iOS
```
tomlongridge, line 180: The subheadings are getting quite nested here, to the extent that the bold body text stands out more than the heading.

Could we have an H5 referring to native usage followed by H5's for each of the JS sections?

packages/react-native/UPGRADING.md
```diff
@@ -0,0 +1,279 @@
...
+```
+cd ios
+pod install
+```
+
+### Updating usage
+
+Full integration instructions are available on the [React Native integration guide](https://docs.bugsnag.com/platforms/react-native/react-native#installation). This section will focus on specific areas that have changed.
+
+Since this update involved a major bump to Bugsnag's JS, Android and iOS notifiers – bringing all of their APIs in sync, virtually every usage of Bugsnag's API in your application will need updating.
```
tomlongridge, line 77: Is there a more positive way to word this?

packages/react-native/UPGRADING.md
bengourley: I figured it was ok to refer to in passing, but happy with your suggestion.

packages/react-native/UPGRADING.md
bengourley: I think I have addressed this.

packages/react-native/UPGRADING.md
bengourley: Have attempted to be more cheery about this. Let me know what you think.

PR #958, opened 2020-07-24
RN: Array metadata test

bengourley:
Ensure a bug in bugsnag-react-native is not present in the rewrite, via an automated test to catch regressions.

### DIFF
--- a/test/react-native/features/fixtures/app/scenarios/MetadataJsScenario.js
+++ b/test/react-native/features/fixtures/app/scenarios/MetadataJsScenario.js
@@ -14,6 +14,7 @@ export class MetadataJsScenario extends Scenario {
     Bugsnag.addMetadata('jsdata', 'some_more_data', 'set via client')
     Bugsnag.notify(new Error('MetadataJsScenario'), (event) => {
       event.addMetadata('jsdata', 'even_more_data', 'set via event')
+      event.addMetadata('jsarraydata', 'items', [ 'a', 'b', 'c' ])
     })
   }
 }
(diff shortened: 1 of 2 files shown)


test/react-native/features/fixtures/app/scenarios/MetadataJsScenario.js
```diff
@@ -14,6 +14,7 @@ export class MetadataJsScenario extends Scenario {
     Bugsnag.addMetadata('jsdata', 'some_more_data', 'set via client')
     Bugsnag.notify(new Error('MetadataJsScenario'), (event) => {
       event.addMetadata('jsdata', 'even_more_data', 'set via event')
+      event.addMetadata('jsarraydata', 'items', [ 'a', 'b', 'c' ])
```
twometresteve, line 17: Would there be a case for having something other than strings in the array?

test/react-native/features/metadata.feature
```diff
@@ -8,8 +8,9 @@ Scenario: Setting metadata (JS)
   And the event "metaData.jsdata.some_data" equals "set via config"
   And the event "metaData.jsdata.some_more_data" equals "set via client"
   And the event "metaData.jsdata.even_more_data" equals "set via event"
+  And the payload field "events.0.metaData.jsarraydata.items" is an array with 3 elements
```
twometresteve, line 11: I'm guessing the bug was that the array wasn't present at all?  I mean, would there be some merit is checking what the 3 values actually are?

test/react-native/features/fixtures/app/scenarios/MetadataJsScenario.js
twometresteve, line 17: Why not apply the same change to the two native scenarios?  Set array metadata in the JS layer and ensure is comes out in handled and unhandled native exceptions.

test/react-native/features/metadata.feature
bengourley: the bug was that it came through as:

```js
{ '0': 'a', '1': 'b', '2': 'c' }
```

instead of


```js
[ 'a', 'b', 'c' ]
```

twometresteve:
> Also approved it, but I think we can make the tests better without a lot of effort.

PR #962, opened 2020-07-27
Update Session tests for JS/Native cross-compatability

Cawllec:
Adds a bunch of assertions targeting:
- The `session.events.handled` count in error payloads
- The behaviour of started, paused, and resumed sessions when triggered from:
  - JS with JS errors
  - JS with Native errors
  - Native with JS errors

It's a bit messy, and the only way of improving the reliability if flakes do occur would be increasing timeouts (which is never a great thing), but I think the extra time take on the test is worth the extra coverage it gives


test/react-native/features/fixtures/app/scenarios/SessionJsControlledManualJsScenario.js
```diff
@@ -0,0 +1,39 @@
...
+              setTimeout(() => {
+                Bugsnag.resumeSession()
+                setTimeout(() => {
+                  Bugsnag.notify(new Error('SessionJsControlledManualJsScenarioC'), () => {}, () => {
+                    setTimeout(() => {
+                      Bugsnag.pauseSession()
+                      setTimeout(() => {
+                        Bugsnag.startSession()
+                        setTimeout(() => {
+                          Bugsnag.notify(new Error('SessionJsControlledManualJsScenarioD'))
```
bengourley, line 26: if you promisify setTimeout, this pyramid of doom would go away

e.g. https://stackoverflow.com/a/33292942

Cawllec: Hopefully that's been resolved now, pending CI

PR #980, opened 2020-08-04
Allow overriding api key in RN iOS via event callback

fractalwrench:
## Goal

Allows overriding the API key in React Native iOS in an event callback by reading the value passed down from the JS layer. This only applies to the iOS layer for now.

Added an E2E scenario to cover the change.

### DIFF
--- a/test/react-native/features/fixtures/app/scenarios/EventApiKeyOverrideScenario.js
+++ b/test/react-native/features/fixtures/app/scenarios/EventApiKeyOverrideScenario.js
@@ -2,13 +2,8 @@ import Scenario from './Scenario'
 import Bugsnag from '@bugsnag/react-native'
 
 export class EventApiKeyOverrideScenario extends Scenario {
-  constructor(configuration, extraData, jsConfig) {
-    super()
-  }
-
   run() {
     Bugsnag.notify(new Error('EventApiKeyOverrideScenario'), event => {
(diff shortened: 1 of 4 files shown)


test/react-native/features/fixtures/app/scenarios/EventApiKeyOverrideScenario.js
```diff
@@ -0,0 +1,15 @@
+import Scenario from './Scenario'
+import Bugsnag from '@bugsnag/react-native'
+
+export class EventApiKeyOverrideScenario extends Scenario {
+  constructor(configuration, extraData, jsConfig) {
+    super()
+  }
+
```
bengourley, line 8:
```suggestion
```

Empty constuctor can be omitted.

test/react-native/features/fixtures/app/scenarios/EventApiKeyOverrideScenario.js
```diff
@@ -0,0 +1,15 @@
...
+  run() {
+    Bugsnag.notify(new Error('EventApiKeyOverrideScenario'), event => {
+      event.setUser('123', 'bug@sn.ag', 'Bug Snag')
```
bengourley, line 11: Is this line relevant to the test? If not I think it should be removed.
```suggestion
```

PR #981, opened 2020-08-04
[PLAT-4704] RN: Improve metadata synchronisation

bengourley:
- Separates out the add and clear use cases for metadata into their own methods
- Properly process updates from the native layer (each update sends the entire new state of metadata, which was previously merged so nothing ever got removed)

Android update includes changes from https://github.com/bugsnag/bugsnag-android/pull/918


fractalwrench:
LGTM

PR #984, opened 2020-08-06
Prevent invalid event.apiKey message logging for JS calls

fractalwrench:


packages/react-native/ios/BugsnagReactNative/BugsnagEventDeserializer.m
```diff
@@ -85,9 +85,12 @@ - (BugsnagEvent *)deserializeEvent:(NSDictionary *)payload {
                                                     threads:[self deserializeThreads:payload[@"threads"]]
                                                     session:session];
     event.context = payload[@"context"];
-    event.apiKey = payload[@"apiKey"];
     event.groupingHash = payload[@"groupingHash"];
 
+    if (payload[@"apiKey"]) {
```
bengourley, line 90: Does this get coerced to a boolen for the if statement? just wondering why it doesn't need `!= nil` check like some other fields in this method.

kattrali: yes - if the key doesn't exist, the statement resolves to nil, which is 'falsy'. Just about any other value is truthy.

twometresteve: On that basis I'm happy to approve this myself.

PR #1002, opened 2020-08-17
[PLAT-4848] RN: Add type definitions to published package files

bengourley:
The type definitions were not inlude in the `files` list in the React Native package's manifest.


henrymoulton:
Thanks guys for the speedy fix!

PR #1006, opened 2020-08-19
test(delivery-node, delivery-x-domain-request): convert tests

djskinner:


babel.config.js
```diff
@@ -8,7 +8,8 @@ module.exports = api => {
   if (api && api.env('test')) {
     presets.push('@babel/preset-typescript')
     plugins.push(['@babel/plugin-proposal-class-properties', { loose: true }])
-    plugins.push(['@babel/plugin-transform-modules-commonjs'])
+    plugins.push('@babel/plugin-transform-modules-commonjs')
+    plugins.push('@babel/plugin-proposal-optional-chaining')
```
djskinner, line 12: allows optional chaining in tests e.g. `expect(err?.meesage).toBe(` which is what the lint rules recommend

packages/delivery-node/delivery.d.ts
```diff
@@ -0,0 +1,6 @@
+import { Delivery } from '@bugsnag/core/client'
```
djskinner, line 1: This works but I don't think is ideal. It's easy enough to use the internal types within the core package but externall we either need to export them or do this. I'm inclined to prefer exporting the `Delivery` type publically but this works for now.

packages/delivery-node/test/delivery.test.ts
```diff
@@ -1,10 +1,18 @@
-const { describe, it, expect } = global
+import delivery from '../'
+import http from 'http'
+import { Client } from '@bugsnag/core'
+import { EventDeliveryPayload } from '@bugsnag/core/client'
```
djskinner, line 4: Same here RE internal types

twometresteve (commented):
Happy that tests have been considered here (it's the whole point!).

packages/delivery-node/delivery.d.ts
bengourley: discussed elsewhere and agreed this is fine

bengourley (approved):

PR #1010, opened 2020-08-20
[PLAT-4691] Fix Expo config type definition

### LINKED ISSUE #953: Expo/TypeScript: apiKey required when calling Bugsnag.start()
According to the Basic Configuration section in the [Expo Manual Setup doc](https://docs.bugsnag.com/platforms/react-native/expo/manual-setup/#basic-configuration), it is recommended to set the Bugsnag API key in `app.json` rather than when calling `Bugsnag.start`. However, the TypeScript definitions require `apiKey` when passing a configuration object to `Bugsnag.start`. 

Here is the TypeScript error from my codebase:
```
Property 'apiKey' is missing in type '{ enabledReleaseStages: string[]; releaseStage: string; logger: null; onError: (event: Event) => false | undefined; }' but required in type 'Config'.

  node_modules/@bugsnag/core/types/common.d.ts:7:3
      apiKey: string
        ~~~~~~
    'apiKey' is declared here.
```

There is a trivial workaround: just use a type assertion when calling `start`:
```
Bugsnag.start({
    apiKey: undefined!,
    // ...
})
```

bengourley:
Fixes #953.

The config object does not require the `apiKey` on Expo, since it may be supplied in `app.json`.


djskinner (approved):
I haven't tested manually but am confident the change and corresponding type test will work as expected

twometresteve (approved):

PR #1012, opened 2020-08-20
[PLAT-4872] Swap to a peer dependency on bugsnag/core in plugins

imjoehaines:
## Goal

Currently our plugins have a hard dependency on `@bugsnag/core`. This can lead to issues when a plugin is updated separately to `@bugsnag/core`, because NPM will dutifully install multiple copies of `@bugsnag/core` — the original version will remain and the plugin will get a newer version of its own. Typically this shouldn't be a problem because changes will be backwards compatible, but it can cause issues for type declarations as a newer `@bugsnag/core` may be incompatible with an old `@bugsnag/core`

To prevent having multiple copies of `@bugsnag/core`, we can use `peerDependencies` instead of a hard dependency. This should mean that there is only ever one top-level version installed, which will be pulled in by `@bugsnag/js`, `@bugsnag/expo` or `@bugsnag/react-native`

## Design

I went with a version constraint of `^7.0.0` to be as compatible as possible, but I think this could be `^7.3.0` (or even the version this is released in (`^7.4.0` ?)) because there shouldn't be a scenario where a new plugin version is installed with an old version of core — not 100% sure this is a good idea though!

## Changeset

- All `delivery-` and `plugin-` packages no longer have a hard dependency on `@bugsnag/core`
- All `delivery-` and `plugin-` packages have a dev dependency on `@bugsnag/core`
- All `delivery-` and `plugin-` packages have a peer dependency on `@bugsnag/core`

## Testing

This should be covered by existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - (plugin-angular): Update bundles and package entrypoints to support the Ivy renderer. [#994](https://github.com/bugsnag/bugsnag-js/pull/994)
 - (react-native): Add `codeBundleId` to config type definition. [#1011](https://github.com/bugsnag/bugsnag-js/pull/1011)
+- Use a peer dependency on @bugsnag/core in plugins [#1012](https://github.com/bugsnag/bugsnag-js/pull/1012)
 
 ## 7.3.2 (2020-08-17)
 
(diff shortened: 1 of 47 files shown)


twometresteve:
Approved from the test perspective, but I have not performed a code review.

PR #1018, opened 2020-08-25
test(delivery-xml-http-request): convert tests

djskinner:
## Goal

## Design

See previous discussions

## Changeset

Converted `delivery-xml-http-request` tests from jasmine to jest.

## Testing


twometresteve (approved):

bengourley (approved):
is the `XMLHttpRequest` not available like `Window` is to these tests? I guess if it were the fake implementation would have to mock more in this test's setup.

djskinner:
> is the XMLHttpRequest not available like Window is to these tests? I guess if it were the fake implementation would have to mock more in this test's setup.

The types are there but it is done like this because the mock implementation is different from the real thing. For example, the mock implementation has a `method: string | null` property for easy assertions but the real interface doesn't have this.

I will rename the interface to MockXMLHttpRequest to make this distinction clearer.

PR #1020, opened 2020-08-25
test(plugin-express, plugin-koa, plugin-restify): convert tests

djskinner:
## Goal

## Design

See previous discussions

## Changeset

- Converted `plugin-express` tests from jasmine to jest.
- Converted `plugin-koa` tests from jasmine to jest.
- Converted `plugin-restify` tests from jasmine to jest.
- Updated internal `__sessionDelegate`  types and updated client and `plugin-browser-device` tests to reflect this.

## Testing


twometresteve (approved):
Fine from a test perspective, but I did wonder if the deletion of the package-lock.json files was intentional.

bengourley (approved):
Since this rule is ignored 3 times, how do you feel about removing it?

`@typescript-eslint/no-non-null-assertion`?

Thinking about the types and dependencies, I guess each of the frameworks should have a peer dependency on the types… but only if the user is using typescript?

djskinner:
> Since this rule is ignored 3 times, how do you feel about removing it?

Agreed. I don't think it has value in tests. I will remove in a followup PR I'm planning, which will also bump the linting dependencies where appropriate.

> Thinking about the types and dependencies, I guess each of the frameworks should have a peer dependency on the types… but only if the user is using typescript?

Possibly, yeah. We can discuss it separately and address it in its own PR. Conditional dependencies for TypeScript would be nice but I'm not sure it is possible. On the other hand, we're essentially doing that now. If using TypeScript in an express app then @types/express will almost definitely be a required dependency so one could argue we're not adding much by making it a peer dependency ourselves. The implication is that non-TypeScript users don't get the dependency they don't need.

> I did wonder if the deletion of the package-lock.json files was intentional.

It doesn't look like any package-lock.json files have been deleted in this PR, though a lot of lines have been removed as a consequence of the reduction in dependencies.

PR #1021, opened 2020-08-25
test(delivery-react-native): convert tests

djskinner:
## Goal

## Design

See previous discussions

## Changeset

Converted `delivery-react-native` tests from jasmine to jest.

## Testing


twometresteve (approved):

packages/delivery-react-native/test/delivery.test.ts
```diff
@@ -1,13 +1,33 @@
...
+  }>
+  severity: string
+  severityReason: {
+    type: string
+    [key: string]: any
+  }
+  unhandled: boolean
+  app: App
+  device: Device
+  threads: unknown[]
```
bengourley, line 17: `Thread` is defined in [`common.d.ts`](https://github.com/bugsnag/bugsnag-js/blob/9f74bf2157260cb2579b7caa95da0fb325a1f76e/packages/core/types/common.d.ts#L126-L133)

In fact, could you build `NativeClientEvent` this type by doing `Pick<EventWithInternals, 'errors' | 'severity' | 'severityReason' etc. >`?

bengourley (approved):
Looks good. Small comment on the types which if there's good justification can be left as-is.

packages/delivery-react-native/test/delivery.test.ts
djskinner: Yes this is much better. Slightly more complicated than I'd hoped but definitely an improvement.

PR #1024, opened 2020-08-26
[PLAT-3907] Use the `latest-v2-cli` branch and `reset_with_timeout` method

Cawllec:
## Goal
- Uses the `latest-v2-cli` release of Maze-runner to ensure future breaking changes do not effect the tests here
- Replaces `reset` with `reset_with_timeout` to add resilience against report duplication.


twometresteve (commented):
Does bin/local_test_util.js also need updating for the Docker tag change?  I tend to myself, but don't know if it is necessary.

twometresteve (approved):
LGTM

dockerfiles/Dockerfile.react-native
```diff
@@ -1,5 +1,5 @@
 # CI test image for unit/lint/type tests
-FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:typed-platform-checks-ci
```
twometresteve, line 2: Slightly concerned this made it onto master in the first place.  No real harm done, but would be good to get this merged asap.

PR #1025, opened 2020-08-27
Add `BugsnagSourceMapUploaderPlugin` to nuxtjs example project

steve-nester-uk:
## Goal

Add `BugsnagSourceMapUploaderPlugin` to nuxtjs example project

### DIFF
--- a/examples/js/nuxtjs/nuxt.config.js
+++ b/examples/js/nuxtjs/nuxt.config.js
@@ -55,7 +55,20 @@ module.exports = {
     ** You can extend webpack config here
     */
     extend (config, ctx) {
+      if (!isDev && isClient) {
 
+        config.devtool = '#source-map'
+
+        config.plugins.push(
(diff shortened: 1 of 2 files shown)


examples/js/nuxtjs/nuxt.config.js
```diff
@@ -55,7 +55,20 @@ module.exports = {
     ** You can extend webpack config here
     */
     extend (config, ctx) {
+      if (!isDev && isClient) {
 
+        config.devtool = '#source-map'
```
bengourley, line 60: I think I saw somewhere in the logs that # was stripped because it was extraneous? https://webpack.js.org/configuration/devtool/#devtool

steve-nester-uk: Yes, you're right - I removed the #

bengourley (approved):

twometresteve (commented):
All good from a test perspective, naturally.

Issue #998, opened 2020-08-15
React native:  No Bugsnag API Key set, js engine: hermes


v3solutions:
Hi,
Since upgrading bugsnag to latest version, can't make it work on adroid. Crashes on opening, however it does send a crash report to my bugsnag account. Bugsnag error: "No Bugsnag API Key set, js engine: hermes". It works no problem with ios.
Api key is set inside AndroidManifest.xml. I've tried inside App.js with and without apiKey:

```
import Bugsnag from '@bugsnag/react-native';
Bugsnag.start({
  apiKey: '90d03f98aa81640e********',
});
```

"react-native": "0.63.2"
"@bugsnag/react-native": "^7.3.1"

using with hermes.

phillipsam:
Hi @v3solutions - are you able to share more about your reproduction case with us?

steve-nester-uk:
@v3solutions we have been able to reproduce a similar issue, which is possibly related to the use of minifyEnabled / proguard for release builds.  We are still investigating but would be interested to know if your issue only happens during release builds (e.g. --variant=release) and whether you are using minifyEnabled?

v3solutions:
@steve-nester-uk sorry for a delay. I'm using minifyEnabled, it does work as a debug. Seems only release build has this issue. If I comment out 
```
Bugsnag.start({
});
``` 
it works too.
Let me know if you need anything. I'll be more active this week.

tobiasbu:
Hi, I recently upgrade React Native to 0.63.2 and Bugsnag to latest version and I started to receiving the "No Bugsnag API Key set" error in Android.

As suggested by @steve-nester-uk disabling `enableProguardInReleaseBuilds`/`minifyEnabled` option worked for me in release variant. Thanks!

mattdyoung:
@tobiasbu - thanks for the info. Yes, we can confirm the bug can affect projects whether or not they use Hermes. We're working on the fix.

johnkiely1:
This has now been fixed and released in [v7.3.4](https://github.com/bugsnag/bugsnag-js/releases/tag/v7.3.4)

PR #1030, opened 2020-09-02
[PLAT-4921] Fix @bugsnag/react-native proguard compatibility

bengourley:
Addresses #998.

- Updates the vendored `bugsnag-android` to include [this fix](https://github.com/bugsnag/bugsnag-android/pull/927).
- Modifies the tests to use proguard on the application so that CI would catch any future regressions in this area
- Removes the duplicate proguard rules from this repo – they are inculded in [`bugsnag-android/bugsnag-plugin-react-native`](https://github.com/bugsnag/bugsnag-android/blob/master/bugsnag-plugin-react-native/proguard-rules.pro).


fractalwrench (approved):
LGTM pending CI

twometresteve (commented):
Looks good from a test perspective.

PR #1031, opened 2020-09-08
Check if CDN uploads are successful

imjoehaines:
## Goal

Knox doesn't error if the request is unsuccessful, so we need to check the status code ourselves

Ultimately we may want to move to a maintained library but, with this check, debugging should be less of a problem in future

## Testing

Manually tested by attempting to upload with invalid credentials — before the change the script silently does nothing; now it fails with an error

### DIFF
--- a/packages/browser/bin/cdn-upload
+++ b/packages/browser/bin/cdn-upload
@@ -25,9 +25,20 @@ const headers = {
   'x-amz-acl': 'public-read'
 }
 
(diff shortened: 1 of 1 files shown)


bengourley (approved):
Looks good and behaves as expected when I set an invalid key.

Would be good to exit with a non-zero status code when it fails, but it didn't do that anyway so not really in the scope of this PR.

PR #1048, opened 2020-09-16
fix(react-native): update bugsnag-cocoa

djskinner:
Updates the vendored bugsnag-cocoa to 6.1.4 primarily to incorporate the following fixes:
- Copy the metadata observer list rather than mutating it directly.

## Goal

So that react-native users can benefit from the _copy the metadata observer list rather than mutating it directly_ bugfix found in bugsnag-cocoa 6.1.4

## Changeset

- Checkout out bugsnag-cocoa master branch as a sibling of bugsnag-js
- Executed `./update-ios.sh` in the packages/react-native directory
- Committed the resulting changes

## Testing

No manual testing has been performed.


bengourley (approved):
LGTM pending CI

PR #1051, opened 2020-09-18
Remove API Keys from example

xander-jones:
## Goal
Remove API key from RN example. When running this example for the first time, there should not be an API key in the example.

## Design
N/A

## Changeset
* `Info.plist`
* `AndroidManifest.xml`

## Testing
N/A


twometresteve (commented):
Looks good from a test perspective.

bengourley (commented):
The example in there is a "valid" key which is not tied to any project. It prevents a warning from being logged about api key format.

xander-jones:
Closed: this API key is intentionally in the example for test purposes.

PR #1053, opened 2020-09-18
Use MacOS 10.15 build servers

twometresteve:
## Goal

Move the Mac-based Buildkite build steps to the new naming convention for Buildkite queues, unlocking access to our new MacOS 10.15 servers whilst maintaining access to the 10.14 servers where appropriate.

For JS, the RN steps may be serviced by either a 10.14 (while they exist) or 10.15 server.  Expo steps will be served only by 10.15 servers as a result of this change, as breaking changes were needed in its setup (unable to install Turtle at the package level on 10.15 but unable to install it global on the 10.15 server).

## Design

Buildkite agents running on Macs used to have a single build queue - opensource-mac. With the addition of 10.15 servers to the existing 10.14 fleet (which will be retired at some point soon) we have moved to tying agents to multiple queues, in order that we can be more targeted in which agents pick up which jobs. There are now the following queues:
- opensource-mac (deprecated)
- opensource-mac-cocoa
- opensource-mac-expo
- opensource-mac-rn

## Changeset
- (Expo) Use a global Turtle install
- Agent queue names in the Buildkite pipelines.
- Remove broken and outdated local iOS build script.

## Testing

Naturally tested via a passing CI pipeline.


package.json
```diff
@@ -67,7 +67,6 @@
     "test:build-browser-container": "docker-compose up --build minimal-packager && docker-compose build --pull browser-maze-runner",
     "test:build-node-container": "docker-compose up --build minimal-packager && docker-compose build --pull node-maze-runner",
     "test:build-expo-android-container": "test/expo/scripts/build-android-locally.sh",
-    "test:build-expo-ios-container": "EXPO_RELEASE_CHANNEL=localtest test/expo/scripts/build-ios-locally.sh",
```
Cawllec, line 70: This script is referenced in https://github.com/bugsnag/bugsnag-js/pull/1053/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R76, so we should amend that to prevent errors.

test/expo/scripts/build-ios.sh
```diff
@@ -5,10 +5,7 @@ set -e
 # Lets make sure the build folder was cleared out correctly
 rm -rf $BUILDKITE_BUILD_CHECKOUT_PATH/build/*
 cd test/expo/features/fixtures/test-app
-npm i turtle-cli@0.14 bunyan
-perl -0777 -i.original -pe "s/entitlements\\['aps-environment'\\] =[^;]+;//gs" node_modules/\@expo/xdl/build/detach/IosNSBundle.js
```
Cawllec, line 9: Have we confirmed that we no longer need to remove the push notification entitlements by running a build locally and then using the produced artefact in a test run?

Cawllec: Resolved through Slack conversation.

package.json
twometresteve: Now removed and README updated.  I've raised an internal ticket to reinstate the ability to run Expo tests locally against an iOS device.

Cawllec (approved):

PR #1060, opened 2020-09-22
fix: Xcode 12 compatibility

radko93:
## Goal

Latest Xcode 12 fails to build without a module to depend on `React-Core` directly hence this change is necessary for all native modules on iOS. This change requires React Native 0.60.2 or newer. For more details please check: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

## Testing

It's not easy to reproduce this, sometimes it might require to create a clean project.


mattdyoung:
@radko93 
Under what circumstances is this change needed?

I've just tested creating a clean React Native (0.63.2) project, integrating bugsnag-js (v.7.3.5) and building in XCode 12 which works without issue.

mattdyoung:
Closing as we've had no response, but happy to re-open if you can share steps to reproduce the issue.

PR #1061, opened 2020-09-22
[PLAT-5002] Include native stacktraces JS errors caused by native promise rejections

bengourley:
When a native module method is called from JS and the native promise is rejected, the JS layer receives the native stacktrace.

This PR brings together the native and JS changes required to include this stacktrace in the event as an additional entry in the errors array.

**Note: landing these changes but with the JS part commented out. Awaiting pipeline update before this feature is enabled.**


twometresteve (commented):
Without looking through the whole changeset in too much detail, I'm content that we have paid adequate attention to testing on this piece of work.

bengourley:
I've tested this in debug/prod builds on both platforms, using the rn0.63 example in this repo. All working as expected.
However, I can't approve the PR since I originally raised it 🙃

twometresteve (approved):
Reviewed primarily from the point of view that the right tests have been added and are being executed in CI.  Approved on the basis that changeset has already been reviewed by PRs into the integration branch.

imjoehaines (approved):

PR #1062, opened 2020-09-22
[PLAT-5000] Include native promise rejection stack traces in Cocoa notifier

nickdowell:
When a native module method is called from JS and the native promise is rejected, the JS layer receives the native stacktrace.

This PR introduces Cocoa notifier changes to process these native stacktraces and include them in error reports.

Note: this currently pulls in an unreleased version of of the Cocoa notifier.


packages/react-native/ios/BugsnagReactNative/BugsnagEventDeserializer.m
```diff
@@ -98,6 +100,21 @@ - (BugsnagEvent *)deserializeEvent:(NSDictionary *)payload {
...
         [event attachCustomStacktrace:error[@"stacktrace"] withType:@"reactnativejs"];
     }
+    
+    id nativeStack = payload[@"nativeStack"];
+    if ([nativeStack isKindOfClass:[NSArray class]] &&
+        [nativeStack filteredArrayUsingPredicate:
+         [NSPredicate predicateWithFormat:@"NOT SELF isKindOfClass: %@", [NSString class]]].count == 0) {
+        NSArray<BugsnagStackframe *> *stackframes = [BugsnagStackframe stackframesWithCallStackSymbols:nativeStack];
+        if (stackframes != nil) {
+            BugsnagError *nativeError = [BugsnagError new];
```
fractalwrench, line 110: The `type` needs to be set to `BSGErrorTypeCocoa` as it's ultimately used by the [Error Reporting API](https://bugsnagerrorreportingapi.docs.apiary.io/#reference/0/notify/send-error-reports) to ensure that stacktraces render correctly for each platform. Using the `initWithErrorClass:errorMessage:errorType:stacktrace:` construct would guarantee that all properties on the object are initialized correctly.

fractalwrench (commented):
Looks good - I think this can be approved once a minor comment has been resolved and the CI build has been fixed: https://buildkite.com/bugsnag/at-bugsnag-react-native/builds/760#782aa1a1-2763-475c-87ed-a3fa261fcef5/46-445

twometresteve (commented):
We'll perform our test-perspective review of all of this once `integration/rn-native-stack` heads for `next`.

fractalwrench (approved):

PR #1065, opened 2020-09-22
Add plugin for React Native Navigation

imjoehaines:
## Goal

This PR adds a plugin for [React Native Navigation](https://wix.github.io/react-native-navigation). This plugin sets the [context](https://docs.bugsnag.com/platforms/react-native/react-native/customizing-error-reports/#setting-context) to the name of the currently rendered component and leaves a breadcrumb on each navigation change

Usage:

```js
import { Navigation } from 'react-native-navigation'
import Bugsnag from '@bugsnag/react-native'
import BugsnagPluginReactNativeNavigation from '@bugsnag/plugin-react-native-navigation'

Bugsnag.start({
  plugins: [new BugsnagPluginReactNativeNavigation(Navigation)]
})
```

### DIFF
--- a/packages/plugin-react-native-navigation/package.json
+++ b/packages/plugin-react-native-navigation/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@bugsnag/plugin-react-native-navigation",
+  "version": "7.3.5",
+  "main": "react-native-navigation.js",
+  "description": "@bugsnag/react-native plugin for integration with react-native-navigation",
+  "homepage": "https://www.bugsnag.com/",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:bugsnag/bugsnag-js.git"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "*.js"
+  ],
+  "author": "Bugsnag",
+  "license": "MIT",
+  "devDependencies": {
+    "@bugsnag/core": "^7.3.5"
+  },
+  "peerDependencies": {
+    "@bugsnag/core": "^7.0.0",
+    "react-native-navigation": "^7.0.0"
+  }
+}
--- a/packages/plugin-react-native-navigation/package.json
+++ b/packages/plugin-react-native-navigation/package.json
@@ -2,6 +2,7 @@
   "name": "@bugsnag/plugin-react-native-navigation",
   "version": "7.3.5",
   "main": "react-native-navigation.js",
+  "types": "types/react-native-navigation.d.ts",
   "description": "@bugsnag/react-native plugin for integration with react-native-navigation",
   "homepage": "https://www.bugsnag.com/",
   "repository": {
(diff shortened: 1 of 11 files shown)


packages/plugin-react-native-navigation/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "*.js"
+  ],
+  "author": "Bugsnag",
+  "license": "MIT",
+  "devDependencies": {
+    "@bugsnag/core": "^7.3.5"
+  },
+  "peerDependencies": {
+    "@bugsnag/core": "^7.0.0",
+    "react-native-navigation": "^7.0.0"
```
bengourley, line 24: Is this incompatible with older versions? Could this version constraint be more loose?

packages/plugin-react-native-navigation/react-native-navigation.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  load (client) {
+    let lastComponent
+
+    this.Navigation.events().registerComponentDidAppearListener(event => {
+      client.setContext(event.componentName)
+
+      if (lastComponent !== event.componentName) {
+        client.leaveBreadcrumb(
```
bengourley, line 13: This ought to honour the `enabledBreadcrumbTypes` config option…

```js
if (client._config.enabledBreadcrumbTypes && client._config.enabledBreadcrumbTypes.includes('navigation')) {
```

bengourley (changes requested):
Looks good. Nice and simple.

- Have you tested it manually on an example app?
- As a top-level plugin it will need types (there will be a similar example when I push my branch in a bit)
- A docs PR is needed

packages/plugin-react-native-navigation/package.json
imjoehaines: Widened it down to v2 as the same API has existed since then

imjoehaines:
> Have you tested it manually on an example app?

Yeah, I've tested v2, v3, v4, v5, v6 & v7

> As a top-level plugin it will need types (there will be a similar example when I push my branch in a bit)

Added in 5d64d9006cce77ffe79521041861256586e781fe

> A docs PR is needed

👍

bengourley (approved):
Looks good, manually verified working as expected on an example project.

twometresteve (commented):
No comments from a test perspective.

Cawllec:
I tripped up creating the test fixtures because I'd not passed `Navigation` into the `new BugsnagReactNativeNavigation()` call.

It took a while to debug due to not having an error message, just a silent failure.  Can we add some sort of message to that?

imjoehaines:
> I tripped up creating the test fixtures because I'd not passed `Navigation` into the `new BugsnagReactNativeNavigation()` call.
> 
> It took a while to debug due to not having an error message, just a silent failure. Can we add some sort of message to that?

We could but it definitely shouldn't be a silent failure; as soon as `load` is called it will try to call a method on `Navigation`. If it's not passed in that will throw, with a stacktrace pointing to the plugin. Were javascript errors being suppressed somehow in the way the tests are setup?

I think it's fine to add an error anyway though — what do you think @bengourley?

bengourley:
It's not silent, but I guess crucially the error happens after the mistake was made. I think it would be a reasonable improvement to add a more friendly error in the constructor.

Though this got me thinking, is there a use case for passing in `Navigation` later, in the way that we do with the React and Vue plugins?

PR #1067, opened 2020-09-23
[PLAT-5010]: Create React Navigation plugin

bengourley:
This PR adds a plugin that integrates with [React Navigation](https://reactnavigation.org/).

The new plugin hooks in to `onStateChange` events to perform the following actions:
- Sets `context` so that when an error happens, so the developer can immediately see which screen it happened on
- Leaves breadcrumbs, so the developer can see what navigation events preceeded an error

Note, end to end tests to follow in separate PRs as it requires a significant amount of @bugsnag/platforms-test-automation input.


packages/plugin-react-navigation/react-navigation.js
```diff
@@ -0,0 +1,57 @@
+const React = require('react')
+
+class Plugin {
```
imjoehaines, line 3: Should this have a real name? I'm not sure if babel/uglify mangles it anyway, but class names get exposed in some places so it could be useful

packages/plugin-react-navigation/types/bugsnag-plugin-react-navigation.d.ts
```diff
@@ -0,0 +1,22 @@
+import { Plugin } from '@bugsnag/core'
+import { NavigationContainer } from '@react-navigation/native'
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface BugsnagPluginReactNavigation extends Plugin { }
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+declare class BugsnagPluginReactNavigation {
+  constructor()
+}
```
imjoehaines, line 9: Is there a reason to prefer this over

```ts
declare class BugsnagPluginReactNavigation implements Plugin { ... }
```

?

packages/plugin-react-navigation/react-navigation.js
```diff
@@ -0,0 +1,57 @@
...
+    const createNavigationContainer = (NavigationContainer) => React.forwardRef((props, ref) => {
+      const { onReady, onStateChange, ...forwardProps } = props
+      const navigationRef = ref || React.useRef(null)
+      const routeNameRef = React.useRef(undefined)
+
+      const wrappedOnReady = (...args) => {
+        const currentRoute = navigationRef.current ? navigationRef.current.getCurrentRoute() : null
+        if (currentRoute) {
+          routeNameRef.current = currentRoute.name
+          client.setContext(currentRoute.name)
```
imjoehaines, line 18: Should this leave a bradcrumb too? I'm not sure what the lifecycle is like so it might not be necessary if there's an `onStateChange` soon after, but it might be useful to see the initial route in case that crashes before an `onStateChange` — e.g. if a user isn't logged in they'd see the login screen, but otherwise would have a different view and it might be useful for that to be reflected in the breadcrumbs

React Native Navigation calls `componentDidAppear` for the initial route, so you get a breadcrumb for the first rendered view with that plugin

packages/plugin-react-navigation/test/react-navigation.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+import Client from '@bugsnag/core/client'
+import TestRenderer from 'react-test-renderer'
+import * as React from 'react'
+import { NavigationContainer, NavigationContainerRef } from '@react-navigation/native'
+
+jest.mock('@react-navigation/native')
+
+afterEach(() => jest.clearAllMocks())
+
+describe('plugin: react navigation', () => {
```
imjoehaines, line 11: It'd be good to have a test where we don't pass `ref`, so the plugin creates one itself

packages/plugin-react-navigation/test/react-navigation.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    const MockedNavigationContainerRender = (NavigationContainer as any).render as jest.MockedFunction<React.ForwardRefRenderFunction<any, any>>
+    TestRenderer.create(<App/>)
+
+    expect(MockedNavigationContainerRender).toBeCalledTimes(1)
+
+    expect(c.getContext()).toBeUndefined()
+
+    MockedNavigationContainerRender.mock.calls[0][0].onReady()
```
imjoehaines, line 71: I think the tests would be more readable if we pulled out variables instead of referring to this in full every time, i.e.

```ts
const passedProps = MockedNavigationContainerRender.mock.calls[0][0]
const passedRef = MockedNavigationContainerRender.mock.calls[0][1]

props.onReady()
expect(c.getContext()).toBe('home')
```

packages/plugin-react-navigation/types/bugsnag-plugin-react-navigation.d.ts
bengourley: It's so that the class dynamically recieves all of the members of the `Plugin` interface. More context/discussion here https://github.com/bugsnag/bugsnag-js/pull/1067#discussion_r494207534 resulting in less duplication.

Since the duplication is minimal (just the `load` signature) I think it would be less confusing to duplicate so I'll switch.

packages/plugin-react-navigation/test/react-navigation.test.tsx
bengourley: The ref has to be passed in in order to mock the `getCurrentRoute()` behaviour of the navigation lib.

imjoehaines: Ah yeah; that makes sense

imjoehaines (approved):

twometresteve (commented):
No comments from a test pespective.

cristianoccazinsp:
Just curious, what happens with this change if we were already leaving breadcrumbs and setting context? Are we going to end up with duplicate logs? Should be listed as a breaking change if so.

imjoehaines:
> Just curious, what happens with this change if we were already leaving breadcrumbs and setting context? Are we going to end up with duplicate logs? Should be listed as a breaking change if so.

This plugin has to be installed manually (it doesn't get pulled in automatically by `@bugsnag/react-native`) so it won't interfere with your custom context/breadcrumbs 🙂

If you'd like to migrate to using the plugin instead of your custom solution, [the docs are here](https://docs.bugsnag.com/platforms/react-native/react-native/navigation-libraries/)

PR #1068, opened 2020-09-24
Add example app for React Native 0.60.0

fractalwrench:
## Goal

Adds an example app for React Native 0.60.0, created using `react-native init`. This helped aid with debugging differences in behaviour between RN 0.60.0 and the latest example app, so should come in use for future development.

Additionally this adds a button to both example apps which triggers a promise rejection, to display JS/native stack trace capture to users.

## Testing

Verified by running the example app on iOS + Android and sending each type of error to the dashbboard.


bengourley (approved):
Looks good at a glance.

PR #1069, opened 2020-09-24
dep: update RN android dependency

fractalwrench:
## Goal

Updates the React Native Android dependency to use the changeset from the following PR: https://github.com/bugsnag/bugsnag-android/pull/940

This fixes how RN < 0.63 parses native stack traces in the case of promise rejections.


fractalwrench:
Merging as Android PR has been approved

PR #1072, opened 2020-09-28
[PLAT-5050] Add anonymous device ID

imjoehaines:
## Goal

Record an anonymous ID which persists across events & sessions

## Changeset

- add a new config option "generateAnonymousId" to core (defaults to true)
- if "generateAnonymousId" is enabled, the browser device plugin will now generate an anonymous ID and store it in local storage. This will be added to all subsequent events & sessions. If persistence fails, the ID will not be sent

## Testing

Unit tests and additions to the browser device E2E tests


test/browser/features/steps/browser_steps.rb
```diff
@@ -101,3 +101,37 @@
     And the payload has a valid sessions array
   }
 end
+
+Then("the event device ID is valid") do
```
imjoehaines, line 105: This feels bad, but I'm not sure we can do much better. Safari 10 doesn't have local storage available on browserstack, so the device ID is always missing. Selenium can't run javascript via the Ruby API on the mobile browsers, so we can't detect if local storage is available

With browserstack currently we could assume that local storage is available if we're not able to run javascript to check. That would let us assert against device ID on the mobile browsers, but might cause flakes in future if local storage stops being available on them (I might have just been lucky with devices on this PR so far)

Happy to do either — I opted for no flakes but I'm not sure how likely it is to break in future so it might be safe to assume local storage is present instead

test/browser/features/steps/browser_steps.rb
```diff
@@ -101,3 +101,37 @@
...
+    $logger.info('Local storage is not supported in this browser, assuming device ID is null')
+    step('the event "device.id" is null')
+  else
+    # if there's no local storage and we can't run javascript from Selenium, we
+    # don't know if the device ID is null or not, so we can't assert against it
+    $logger.warn('Selenium cannot run JavaScript in this browser. Device ID will not be checked!')
+  end
+end
+
+When("I save {string} in local storage as {string}") do |key, value|
```
imjoehaines, line 131: Not having local storage available seems like grounds to skip this test instead ?

packages/plugin-browser-device/device.js
```diff
@@ -19,6 +43,15 @@ module.exports = (nav = navigator, screen = window.screen) => ({
           : 'portrait'
     }
 
+    if (client._config.generateAnonymousId) {
+      const id = getDeviceId()
+
+      // If there's no ID, it's likely because localStorage is disabled
+      if (id) {
+        device.id = id
```
bengourley, line 51: as long as `getDeviceId()` returns an `id` or `undefined`, I think it's safe enough and terser to do:

`device.id = getDeviceId()`

bengourley (approved):
One comment that might trim a few bytes, otherwise implementation looks good.

test/browser/features/steps/browser_steps.rb
bengourley: I think we ought to assume that it can run. At the moment this gives us "false" positives for this feature on mobile browsers without actually testing it.

test/browser/features/steps/browser_steps.rb
bengourley: it does, but as established, this check failing doesn't mean local storage is not available, on mobile devices it just means you can execture arbitrary JS over the selenium interface

test/browser/features/steps/browser_steps.rb
imjoehaines: Yeah I think I agree; I'll change it to assume we can use local storage unless we can prove otherwise

test/browser/features/steps/browser_steps.rb
imjoehaines: I changed this to use a new HTML page to setup local storage instead, so that we don't need to run javascript via Selenium to run this test

It now also skips if we know local storage isn't available (i.e. Safari 10)

bengourley (approved):

PR #1073, opened 2020-09-28
Add React-Navigation test fixture to pipeline

Cawllec:
## Goal

To refactor the existing fixture build system to enable different test fixtures to be built using a single "base" app and multiple JS sets/installation scripts.

This change _doesn't_ however add actual tests for those breadcrumbs, but once this is passing on CI I'll create a couple of sample feature files that can be edited at leisure.

## Design

Maintaining multiple unique fixtures is a road I don't believe any of us wish to go down, so making the basic features complete skeletons and copying everything specific in at build time seems like the best solution.

By separating the JS files into sub-directories at a higher level we can in theory add as many different build setups as we'd like and simply change the `JS_SCENARIO_DIR` environment variable on different build steps to get a completely different test fixture. 

## Testing

Running locally appears to work, the real test will be ensuring it runs on CI.


.buildkite/react-native-pipeline.yml
```diff
@@ -59,6 +59,28 @@ steps:
 
   - wait
 
+  - label: ':android: Build react-navigation 0.63 apk'
+    env:
+      REACT_NATIVE_VERSION: "rn0.63"
```
twometresteve, line 64: I think we also need to build for RN 0.60?

scripts/react-native-helper.js
```diff
@@ -54,15 +61,25 @@ module.exports = {
     common.run(`npm install --registry ${registryUrl}`, true)
 
     // Install notifier
-    const command = `npm install @bugsnag/react-native@${version} --registry ${registryUrl}`
+    let command = `npm install @bugsnag/react-native@${version}  --registry ${registryUrl}`
+    common.run(command, true)
+
+    // Install navigation tracker
+    command = `npm install @bugsnag/plugin-react-navigation@${version} --registry ${registryUrl}`
```
twometresteve, line 68: How's the `plugin-react-native-navigation` going to fit in?

twometresteve:
I'm getting slightly dizzy thinking about all the combinations, but are we heading towards the following test fixtures:
On Android:
- RN 0.60, no nav tracker
- RN 0.63, no nav tracker
- RN 0.60, React Navigation
- RN 0.63, React Navigation
- RN 0.60, React Native Navigation
- RN 0.63, React Native Navigation
On iOS:
- As the above, repeated (!)

And what about the test scenarios?  Do we now have two sets of feature files:
- Those that already exist and run on the test fixtures with no navigation built in
- "Navigation breadcrumb" features to be written and run on all test fixtures with navigation in?  If so, is it exactly the same set of tests run on each of these?

.buildkite/react-native-pipeline.yml
Cawllec: Yes, this is more a test run than anything else atm

scripts/react-native-helper.js
Cawllec: That's a good question, I'd assumed they were the same module until now

PR #1077, opened 2020-09-30
test(delivery-expo): convert tests

djskinner:
## Goal

## Design

See previous discussions

## Changeset

Converted `delivery-expo` tests from jasmine to jest.

## Testing


unimodules-core.d.ts
```diff
@@ -0,0 +1,8 @@
+// @unimodules/core is referenced by the expo-file-system types.
+// It is stubbed out in order to avoid depending on a chain of other
+// modules just for the type definitions.
+
+declare module '@unimodules/core' {
+  export type EventEmitter = any
+  export type Subscription = any
+}
```
djskinner, line 8: Without this, repo dev dependencies of `react-native-unimodules` and `@types/react-native` are required, which doesn't work because `@types/react-native` conflicts with the existing `@types/react`.

This will get resolved more satisfactorily in the next phase of moving to individual extended tsconfigs for each package.

bengourley: does it have to live at the root? or can it live inside of `delivery-expo`?

djskinner: It can live anywhere, I wasn't sure where to put it as it's not just the `delivery-expo` package that requires it, but likely some of the other expo modules.

bengourley: hmm ok. maybe if it went in the expo package, can all the others reference it from there?

twometresteve (commented):
No comments from Test.

bengourley (approved):

PR #1079, opened 2020-10-01
test: improve jest configuration

djskinner:
## Goal

Simplify the jest testing configuration and fix the following warning in the test output:

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

      at get (node_modules/react-native/Libraries/Core/setUpXHR.js:28:30)
      at Object.getValue [as Blob] (node_modules/react-native/Libraries/Utilities/defineLazyObjectProperty.js:42:16)
          at Array.forEach (<anonymous>)
```

## Design

Remove unnecessary configuration. Specify own minimal test setup.

## Changeset

- Removal of `setUpGlobals.js` and `setUpXHR.js` test setup for react-native tests
- Define `window` in test setup, which is the only thing we used from `setUpGlobals.js`
- Remove unnecessary ts configuration

## Testing

Automated tests pass

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -54,8 +54,6 @@ module.exports = {
         testsForPackage('plugin-react-native-session')
       ],
       setupFiles: [
-        require.resolve('react-native/Libraries/Core/setUpGlobals.js'),
-        require.resolve('react-native/Libraries/Core/setUpXHR.js'),
         '<rootDir>/packages/react-native/src/test/setup.js'
       ]
     },
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -48,10 +48,7 @@
     //   "node_modules/@types",
     //   "packages/core/types"
     // ],
-    "types": [                                /* Type declaration files to be included in compilation. */
-      "node",
-      "jest"
-    ],
+    // "types": []                                /* Type declaration files to be included in compilation. */
     // "allowSyntheticDefaultImports": true,  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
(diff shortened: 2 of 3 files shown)


jest.config.js
```diff
@@ -54,8 +54,6 @@ module.exports = {
         testsForPackage('plugin-react-native-session')
       ],
       setupFiles: [
-        require.resolve('react-native/Libraries/Core/setUpGlobals.js'),
-        require.resolve('react-native/Libraries/Core/setUpXHR.js'),
```
bengourley, line 58: apparently i thought this was necessary for the rn integration tests https://github.com/bugsnag/bugsnag-js/commit/83bbb38ed617cc5d835e4013600c2b9f53832eec#diff-2d0cd5d10b9604941c38c6aac608178a

but since they still pass on this branch, it seems it is not required

tsconfig.json
```diff
@@ -48,10 +48,7 @@
     //   "node_modules/@types",
     //   "packages/core/types"
     // ],
-    "types": [                                /* Type declaration files to be included in compilation. */
-      "node",
-      "jest"
-    ],
+    // "types": []                                /* Type declaration files to be included in compilation. */
```
bengourley, line 51: how come these are no longer required?

bengourley (approved):

tsconfig.json
djskinner: I'm afraid I can't remember fully. I think added when I was working through the clashing of `@types/react` and `@types/react-native`

PR #1109, opened 2020-10-21
Ensure every buildkite step has a timeout

twometresteve:
## Goal

Adds timeouts to Buildkite steps that were missing them.

## Design

Every step should have a timeout, even if far in excess of what might be needed.  Whilst the infrastructure will scale, the number of instances are capped - potentially adding to CI congestion and there is a cost implication too.

## Changeset

React Native pipeline - Expo and the main JS ones were ok.

## Testing

Covered by standard CI.

### DIFF
--- a/.buildkite/react-native-pipeline.yml
+++ b/.buildkite/react-native-pipeline.yml
@@ -1,5 +1,6 @@
 steps:
   - label: ':docker: Build and publish JS packages'
+    timeout_in_minutes: 30
     plugins:
       - docker-compose#v3.3.0:
           build: publisher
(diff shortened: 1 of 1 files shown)


bengourley (approved):
Lots of retries required on buildkite, but assuming unrelated?

twometresteve:
> Lots of retries required on buildkite, but assuming unrelated?

Yes, absolutely - the majority were due to the MazeRunner image disappearing, which has now been dealt with.  The Android 9 tests do certainly seem to need some attention separately in due course though.

PR #1113, opened 2020-10-26
test(expo): convert tests

djskinner:
## Goal

## Design

See previous discussions

## Changeset

Converted `expo-cli` tests from jasmine to jest.

## Testing


packages/expo/test/index.test.ts
```diff
@@ -0,0 +1,166 @@
...
+import { Breadcrumb, Session } from '../types/bugsnag'
+
+jest.mock('expo-constants', () => ({
+  default: {
+    platform: {},
+    manifest: {}
+  }
+}))
+
+jest.mock('../../plugin-expo-device/node_modules/expo-constants', () => ({
```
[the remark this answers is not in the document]
bengourley: it's probably a side-effect of the monorepo setup where there are multiple instances of that module on the filesystem

packages/expo/test/index.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      maxBreadcrumbs: 20,
+      enabledBreadcrumbTypes: ['manual', 'log', 'request'],
+      user: null,
+      metadata: {},
+      logger: undefined,
+      redactedKeys: ['foo', /bar/]
+    })
+
+    Bugsnag.notify(error, onError)
+    expect(true).toBe(true)
```
bengourley, line 161: I know a lot of it is mocked, but there must be a better assertion we can make about the client here?

djskinner: I'm struggling to know what to do here. I want to mock fetch and assert it was called. However, for reasons I don't fully understand, but I think is to do with the monorepo setup, I'm unable to mock the fetch that `delivery-expo` imports.

djskinner: Ok I've resorted to mocking delivery-expo and asserting on that. It's better. For some reason though, I cannot assert on the exceptions part of the payload. It's there in the jest diff but when I try to access `.exceptions`, it's undefined.

bengourley: ah, I did a little digging and this seems to be due to the fact that the `event` object at runtime has a `.errors` property, but to match the API, when this gets [JSON stringified it gets renamed to `exceptions`](https://github.com/bugsnag/bugsnag-js/blob/ac7172ad6f794e68833ca2cb22227ae585912abd/packages/core/event.js#L82).

So when Jest output the actual value when it doesn't match the expected value, it looks like it JSON-stringifies the actual value.

I added the following and it worked:

```diff
      expect(_delivery.sendEvent).toHaveBeenCalledWith(expect.objectContaining({
        apiKey: '030bab153e7c2349be364d23b5ae93b5',
        events: expect.arrayContaining([
          expect.objectContaining({
            app: expect.objectContaining({ releaseStage: 'production', version: '1.2.3' }),
            breadcrumbs: [],
            device: expect.objectContaining({ manufacturer: 'Google', model: 'Pixel 4', modelNumber: undefined, osName: 'android', totalMemory: undefined }),
+             errors: expect.arrayContaining([
+               expect.any(Object)
+            ])
          })
        ]),
        notifier: { name: 'Bugsnag Expo', url: 'https://github.com/bugsnag/bugsnag-js', version: '7.5.2' }
      }), expect.any(Function))
```

bengourley (approved):

PR #1115, opened 2020-10-28
docs(expo): Outline manual dependency sync process

bengourley:
## Goal

Document the process so that engineers without prior knowledge of the project's nuances can successfully keep Expo deps in sync.

## Design

Keeping it simple and local to the project, and subject to the same review process for changes/updates. 

## Changeset

Added a CONTRIBUTING.md to the `@bugsnag/expo` package.

## Testing

N/A


packages/expo/CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
...
+- `expo-constants` (`@bugsnag/plugin-expo-app`)
+- `expo-device` (`@bugsnag/plugin-expo-device`)
+
+If you add a new dependency please add it to this list.
+
+To check what native module versions are bundled with Expo, check this file:
+
+https://github.com/expo/expo/blob/master/packages/expo/bundledNativeModules.json
+
+To check whan JS-only module versions are depended on by Expo, check this file:
```
imjoehaines, line 22:
```suggestion
To check what JS-only module versions are depended on by Expo, check this file:
```

packages/expo/CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
...
+
+https://github.com/expo/expo/blob/master/changelogVersions.json
+
+## Updating the CLI to install a compatible notifier version
+
+When the version of the bundled native modules changes, making the notifier incompatible with previous Expo SDKs we need to update the CLI using the established pattern. See `packages/expo-cli/commands/install.js` for details.
```
imjoehaines, line 28: I initially read this as "When the version of the bundled native modules changes, the notifier _might_ be incompatible with previous Expo SDKs", when I think it means "When the version of the bundled native modules changes, the notifier _will_ be incompatible with previous Expo SDKs"

Something like this might be clearer?

```suggestion
When the version of the bundled native modules changes the notifier will be incompatible with previous Expo SDKs. To prevent installing the conflicting versions, we need to update the CLI using the established pattern in [`packages/expo-cli/commands/install.js`](../expo-cli/commands/install.js).
```

imjoehaines (approved):

bengourley:
Mergining in spite of CI failure. This was due to an issue that is now fixed in #1116. This PR only contains a markdown file.

PR #1125, opened 2020-11-11
React Native CI and local build speed improvements

twometresteve:
## Goal

Makes some improvements to build speeds on CI and locally for React Native on Android , whilst also ironing out a couple of wrinkles in the process.  The elapsed time on CI for the first end-to-end tests to start running is now 8 minutes (down from almost 20).

Further improvements can still be made, but these will follow in subsequent PRs.

## Changeset

Separating the non-notifier-specific Docker content into its own image allows us to be less careful about the invalidation of Docker layer caches and as a result, the confusing "gathering" and "prepare" steps are no longer needed.

The triggering of the Expo/RN pipelines have been brought to the front of the JS pipeline.  The previous placement was unnecessarily low down and artificially prolonged feedback time.

Locally, the Docker image is no longer used to build the test fixture.  It's far quicker to build it on the local file system and, having published the notifier, the test fixture builds from scratch in about a minute.

Some other small efficiency improvements realised in the process.  For example, the iOS subfolders were all being `rsync`ed into place, unnecessarily.

## Testing

Covered by passing CI - but I've also verified that building the test fixture locally still works.


dockerfiles/Dockerfile.android-builder-base
```diff
@@ -0,0 +1,28 @@
...
+ENV PATH="${PATH}:/sdk/platform-tools"
+
+ENV ANDROID_HOME="/sdk/"
+ENV GRADLE_OPTS="-Dorg.gradle.daemon=false"
+
+# Pre-download Gradle
+ENV GRADLE_USER_HOME="/app/gradle"
+WORKDIR /app/gradle
+
+RUN wget -q https://services.gradle.org/distributions/gradle-5.1.1-all.zip \
```
bengourley, line 26: are these the versions known statically? Or will this get out of date when we add a new RN fixture, and (potentially) unknowingly cause a different version of gradle to be downloaded on-demand way later in the build?

bengourley (approved):

dockerfiles/Dockerfile.android-builder-base
twometresteve: You're right, they could get out of date and cause a download later in the build.  It's not a massive issue on CI as it's so fast out there to download things, but it is an area to work on improving.

PR #1128, opened 2020-11-13
Add Bugsnag-Integrity header to browser XML HTTP delivery

imjoehaines:
## Goal

This PR adds the Bugsnag-Integrity header to the browser XML HTTP delivery

## Testing

- Unit tests check for the header being present
- the Maze Runner "the request is a valid browser payload for the _x_ API" steps now also check for the new header, if the browser supports arbitrary headers

### DIFF
--- a/packages/delivery-xml-http-request/delivery.js
+++ b/packages/delivery-xml-http-request/delivery.js
@@ -1,3 +1,4 @@
+const bytesize = require('@bugsnag/bytesize')
 const payload = require('@bugsnag/core/lib/json-payload')
 
 module.exports = (client, win = window) => ({
@@ -8,12 +9,17 @@ module.exports = (client, win = window) => ({
       req.onreadystatechange = function () {
         if (req.readyState === win.XMLHttpRequest.DONE) cb(null)
       }
+
+      const json = payload.event(event, client._config.redactedKeys)
+
       req.open('POST', url)
       req.setRequestHeader('Content-Type', 'application/json')
       req.setRequestHeader('Bugsnag-Api-Key', event.apiKey || client._config.apiKey)
+      req.setRequestHeader('Bugsnag-Integrity', 'simple ' + bytesize(json))
       req.setRequestHeader('Bugsnag-Payload-Version', '4')
       req.setRequestHeader('Bugsnag-Sent-At', (new Date()).toISOString())
-      req.send(payload.event(event, client._config.redactedKeys))
+
+      req.send(json)
     } catch (e) {
       client._logger.error(e)
     }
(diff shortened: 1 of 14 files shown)


test/browser/features/steps/browser_steps.rb
```diff
@@ -119,3 +123,12 @@
...
   end
 end
+
+Then("the Bugsnag-Integrity header is valid") do
+  raw_request = Server.current_request[:request]
+
+  type, value = raw_request['Bugsnag-Integrity'].split(' ')
+
+  assert_equal('simple', type)
+  assert_equal(raw_request.body.bytesize, value.to_i)
```
bengourley, line 133: Since this uses the ruby bytesize method we could leverage it further to test that the `@bugsnag/bytesize` module you behaves in the same way for some of the complicated cases.

It would be good to add a feature file and some fixtures that send some combinations of multibyte characters, rather like you have in the unit tests for the module itself.

packages/delivery-xml-http-request/delivery.js
```diff
@@ -8,12 +9,17 @@ module.exports = (client, win = window) => ({
       req.onreadystatechange = function () {
         if (req.readyState === win.XMLHttpRequest.DONE) cb(null)
       }
+
+      const json = payload.event(event, client._config.redactedKeys)
+
       req.open('POST', url)
       req.setRequestHeader('Content-Type', 'application/json')
       req.setRequestHeader('Bugsnag-Api-Key', event.apiKey || client._config.apiKey)
+      req.setRequestHeader('Bugsnag-Integrity', 'simple ' + bytesize(json))
```
bengourley, line 18: minor nit – backticks are prefferd syntax for string interpolation (but it definitely gets compiled down to what you wrote)

```suggestion
      req.setRequestHeader('Bugsnag-Integrity', `simple ${bytesize(json)}`)
```

packages/delivery-xml-http-request/delivery.js
```diff
@@ -25,12 +31,17 @@ module.exports = (client, win = window) => ({
       req.onreadystatechange = function () {
         if (req.readyState === win.XMLHttpRequest.DONE) cb(null)
       }
+
+      const json = payload.session(session, client._config.redactedKeys)
+
       req.open('POST', url)
       req.setRequestHeader('Content-Type', 'application/json')
       req.setRequestHeader('Bugsnag-Api-Key', client._config.apiKey)
+      req.setRequestHeader('Bugsnag-Integrity', 'simple ' + bytesize(json))
```
bengourley, line 40: same again

```suggestion
      req.setRequestHeader('Bugsnag-Integrity', `simple ${bytesize(json)}`)
```

bengourley (approved):
Some nits/suggestions if there's time, but otherwise 👍

test/browser/features/steps/browser_steps.rb
imjoehaines: I did verify the original implementation with Ruby's bytesize method, but that is a good idea to make sure it stays equivalent

Should I make a new ticket for that though, as this integration branch won't be merged that soon?

imjoehaines: Added some MR tests with the examples in 90595bd28d540524a889e0d4678f1402bde7eea3

PR #1130, opened 2020-11-14
Move Browser and Node tests to MazeRunner v3

twometresteve:
## Goal

Update the main pipeline to use MazeRunner v3.

## Design

The MazeRunner Docker image is based on Ubuntu rather than Alpine, which seems to provide greater stability for the BrowserStackLocal binary (which does not directly support Alpine).  It also removes the need for copying in various glibc artefacts.

## Changeset

In addition to the move to MazeRuner v3, I've also taken the opportunity to review the platform version coverages.  One of the scenarios is failing on the later version of Chrome, Firefox and Safari, so I have put a soft fail on these pending further investigation.

## Testing

Basic move to v3 is covered by CI.  As noted above, there is a scenario failing on the newer browsers that requires further investigation.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -135,8 +135,6 @@ steps:
       BROWSER: "chrome_latest"
     concurrency: 5
     concurrency_group: 'browserstack'
-    soft_fail:
-      - exit_status: "*"
 
   - label: ':ie: v8 Browser tests'
     depends_on: "browser-maze-runner-image"
@@ -255,8 +253,6 @@ steps:
       BROWSER: "safari_13"
     concurrency: 5
     concurrency_group: 'browserstack'
-    soft_fail:
-      - exit_status: "*"
 
   - label: ':desktop_computer: Opera v12 Browser tests'
     depends_on: "browser-maze-runner-image"
@@ -336,8 +332,6 @@ steps:
       BROWSER: "firefox_latest"
     concurrency: 5
     concurrency_group: 'browserstack'
-    soft_fail:
-      - exit_status: "*"
 
   - label: ':node: Node 4'
     depends_on: "node-maze-runner-image"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -135,6 +135,8 @@ steps:
       BROWSER: "chrome_latest"
     concurrency: 5
     concurrency_group: 'browserstack'
+    soft_fail:
+      - exit_status: "*"
 
   - label: ':ie: v8 Browser tests'
     depends_on: "browser-maze-runner-image"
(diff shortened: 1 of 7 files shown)


imjoehaines (changes requested):
The test failure is because [browser_errors.yml](https://github.com/bugsnag/bugsnag-js/blob/next/test/browser/features/fixtures/browser_errors.yml) needs to be updated to know about the new browsers; they're probably the same as the older versions

It's probably worth improving [the function that uses the data from that file](https://github.com/bugsnag/bugsnag-js/blob/eeb2292723fa0843b8f2ca44eb476ca6f1a741c7/test/browser/features/support/env.rb#L28-L30) to check if the browser exists and add a useful error message if it doesn't

Something like

```ruby
def get_error_message id
  browser = ENV['BROWSER']

  raise "The browser '#{browser}' does not exist in 'browser_errors.yml'" unless ERRORS.has_key?(browser)

  ERRORS[browser][id]
end
```

twometresteve:
> The test failure is because [browser_errors.yml](https://github.com/bugsnag/bugsnag-js/blob/next/test/browser/features/fixtures/browser_errors.yml) needs to be updated to know about the new browsers; they're probably the same as the older versions
> 
> It's probably worth improving [the function that uses the data from that file](https://github.com/bugsnag/bugsnag-js/blob/eeb2292723fa0843b8f2ca44eb476ca6f1a741c7/test/browser/features/support/env.rb#L28-L30) to check if the browser exists and add a useful error message if it doesn't
> 
> Something like
> 
> ```ruby
> def get_error_message id
>   browser = ENV['BROWSER']
> 
>   raise "The browser '#{browser}' does not exist in 'browser_errors.yml'" unless ERRORS.has_key?(browser)
> 
>   ERRORS[browser][id]
> end
> ```

Updated as suggested.  the `errorMessage` for `unhandled_syntax` needed changing in each case, but I was happy with the change given the messages for other browsers (variations on "unexpected token: '!'").

imjoehaines (approved):

bengourley (approved):
Looks good.

I'm cautious about testing the latest versions of browsers with our current set up. It has the potential to be quite brittle and can result in test failures that are not due to problems with the notifier (as you noticed by having to tweak the specifics of the line number and error message expectations for unhandled errors).

However, that said – we haven't added recent browsers to the matrix in a long time, so happy to try this out and see how brittle it ends up being in practice, which may be a reasonable tradeoff for the benefits of automatically testing against the latest browsers.

twometresteve:
> Looks good.
> 
> I'm cautious about testing the latest versions of browsers with our current set up. It has the potential to be quite brittle and can result in test failures that are not due to problems with the notifier (as you noticed by having to tweak the specifics of the line number and error message expectations for unhandled errors).
> 
> However, that said – we haven't added recent browsers to the matrix in a long time, so happy to try this out and see how brittle it ends up being in practice, which may be a reasonable tradeoff for the benefits of automatically testing against the latest browsers.

Yes, agreed and happy to receive that feedback.  I was cautiously optimistic about trying to roll with the latest, but wouldn't be upset to see us fix to a specific version and update periodically.

PR #1132, opened 2020-11-17
Separate pipelines for Browser, Node, Expo and RN

twometresteve:
## Goal

Spilt the JS build into separate pipelines for Browsers, Node, Expo and React Native.  

## Design

A new pipeline for the React Native CLI will slot in next to these and the main pipeline contains "general" steps such as package/publishing and unit tests.  The package step may eventually disappear if Node/Browsers are switched over to using the NPM published artifacts.

## Changeset

As usual with changes like this the diffs look far worse than the real changes.  It was almost copy/paste other than:
- Moving the what was RN-specific publish step to the main pipeline.
- Adding the `build` attribute on artifact download steps
- Step dependencies reviewed

## Testing

Covered by CI passing, combined with a sanity check that no build steps have been lost as part of the refactor.  The browser tests currently fail, as an update to our S3 secrets bucket is needed in order to access the required BrowserStack credentials.

### DIFF
--- a/.buildkite/browsers-pipeline.yml
+++ b/.buildkite/browsers-pipeline.yml
@@ -1,0 +1,254 @@
+steps:
+
+  - label:  ':docker: Build browser maze runner image'
+    key: "browser-maze-runner-image"
+    timeout_in_minutes: 20
+    plugins:
+      - artifacts#v1.3.0:
+          download: min_packages.tar
+          build: ${BUILDKITE_TRIGGERED_FROM_BUILD_ID}
+      - docker-compose#v3.7.0:
+          build:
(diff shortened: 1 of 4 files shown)


Cawllec (commented):
Looks good, only suggestion is that we change the Browser, Node & Expo build steps to use the pre-packaged distro, as that would likely save several minutes from each.

Cawllec (approved):

twometresteve:
> Looks good, only suggestion is that we change the Browser, Node & Expo build steps to use the pre-packaged distro, as that would likely save several minutes from each.

Yes, agreed - I'll raise an internal ticket to address this before too long.

PR #1133, opened 2020-11-18
Add Bugsnag Integrity header to Expo delivery

imjoehaines:
## Goal

Adds the Bugsnag-Integrity header to Expo delivery, using `expo-crypto` to calculate the sha1 hash of the payload

### DIFF
--- a/packages/delivery-expo/package.json
+++ b/packages/delivery-expo/package.json
@@ -18,7 +18,8 @@
   "license": "MIT",
   "dependencies": {
     "@react-native-community/netinfo": "5.9.6",
-    "expo-file-system": "~9.2"
+    "expo-file-system": "~9.2",
+    "expo-crypto": "~8.3.0"
   },
   "devDependencies": {
     "@bugsnag/core": "^7.3.5"
--- a/packages/delivery-expo/delivery.js
+++ b/packages/delivery-expo/delivery.js
@@ -2,6 +2,7 @@ const payload = require('@bugsnag/core/lib/json-payload')
 const UndeliveredPayloadQueue = require('./queue')
 const NetworkStatus = require('./network-status')
 const RedeliveryLoop = require('./redelivery')
+const Crypto = require('expo-crypto')
 
 module.exports = (client, fetch = global.fetch) => {
   const networkStatus = new NetworkStatus()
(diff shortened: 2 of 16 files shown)


packages/delivery-expo/package.json
```diff
@@ -18,7 +18,8 @@
   "license": "MIT",
   "dependencies": {
     "@react-native-community/netinfo": "5.9.6",
-    "expo-file-system": "~9.2"
+    "expo-file-system": "~9.2",
+    "expo-crypto": "~8.3.0"
```
bengourley, line 22: How did you wind up with this version?

As of the latest Expo (v40) the latest version of the crypto module is 8.1.0: https://github.com/expo/expo/blob/c326745ae16d7b6bbad86de69a61ea9606085b7a/changelogVersions.json#L27

I think we ought to pin it to that.

bengourley (approved):
Looks good aside from maybe the exact version.

packages/delivery-expo/package.json
imjoehaines: I don't think SDK 40 has been released, has it? This page suggests 39 is the latest version https://docs.expo.io/workflow/upgrading-expo-sdk-walkthrough/

This matches the version constraint for Expo 39: https://github.com/expo/expo/blob/sdk-39/packages/expo/bundledNativeModules.json#L25

bengourley: huh, how confusing, the depedency for the unrelease v40 is _lower_ than that of 39 😕

Ok, as long as it matches v39 I'm happy.

imjoehaines: Just noticed we linked different files — for SDK 39 the version constraint in [`bundledNativeModules` is `~8.3.0`](https://github.com/expo/expo/blob/sdk-39/packages/expo/bundledNativeModules.json#L25) but in [`changelogVersions` it's `8.1.0`](https://github.com/expo/expo/blob/sdk-39/changelogVersions.json#L27)

Expo Crypto has native components (https://github.com/expo/expo/tree/sdk-39/packages/expo-crypto) so I think `bundledNativeModules` is the right one to use (?) I'm not sure why it's in both files & why they are different versions though :s

PR #1134, opened 2020-11-18
Placeholder pipeline for React Native CLI

twometresteve:
## Goal

Bare minimum PR to prepares the ground for the forthcoming React Native CLI.

## Changeset

New, empty, pipeline for RN CLI, triggered from the main pipeline.

I noticed an inconsistency in the naming of the browser pipeline compared to the `test/browser` folder, so ironing that out now as well.

## Testing

Covered by passing CI and I will also check that the appropriate Github status checks are in place.


bengourley (approved):
I can't see the output of "echo Placeholder" in the placeholder pipeline output. If that is to be expected, ignore me :)

twometresteve:
> I can't see the output of "echo Placeholder" in the placeholder pipeline output. If that is to be expected, ignore me :)

As expected - the step is skipped to avoid using any resources and potentially delaying the build whilst waiting for an agent.  But Buildkite still wanted me to put _something_ in there.

PR #1136, opened 2020-11-19
[PLAT-4935] Add automate-symbolication command (Android)

bengourley:
Add a command to automate source map upload when a React Native build happens.

- `git fetch && git checkout feature/rn-cli-automate-symbolication-cmd`
- `npm i`
- `npm run bootstrap`
- `npm run build`
- `npm link` (this symlinks the local package to your global npm modules)

On a React Native project:

- run `bugsnag-react-native-cli automate-symbolication`

### DIFF
--- a/packages/react-native-cli/src/lib/Gradle.ts
+++ b/packages/react-native-cli/src/lib/Gradle.ts
@@ -13,7 +13,7 @@ export async function modifyRootBuildGradle (projectRoot: string, logger: Logger
   try {
     await insertValueAfterPattern(
       topLevelBuildGradlePath,
-      /[\r\n]\s*classpath\("com.android.tools.build:gradle:\d+\.\d+\.\d+"\)/,
+      /[\r\n]\s*classpath\(["']com.android.tools.build:gradle:.+["']\)/,
       GRADLE_PLUGIN_IMPORT,
       logger
     )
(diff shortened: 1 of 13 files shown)


packages/react-native-cli/src/lib/Gradle.ts
```diff
@@ -0,0 +1,100 @@
...
+const DOCS_LINK = 'https://docs.bugsnag.com/build-integrations/gradle/#installation'
+
+export async function modifyRootBuildGradle (projectRoot: string, logger: Logger): Promise<void> {
+  logger.debug('Looking for android/build.gradle')
+  const topLevelBuildGradlePath = path.join(projectRoot, 'android', 'build.gradle')
+  logger.debug('Importing com.bugsnag:bugsnag-android-gradle-plugin')
+  try {
+    await insertValueAfterPattern(
+      topLevelBuildGradlePath,
+      /[\r\n]\s*classpath\("com.android.tools.build:gradle:\d+\.\d+\.\d+"\)/,
```
fractalwrench, line 16: There are a couple of ways this dependency could be declared, particularly if a user has a brownfield app that they've manually setup themselves. It'd be good to consider the following tweaks as hopefully they'd be relatively little effort and capture some common use cases:

1. Gradle can use either single or double quotes for declaring dependencies, so this could try and check both
2. Gradle dependency versions can be expressed in [various different ways](https://docs.gradle.org/current/userguide/single_versions.html) and can even be a [string interpolated groovy variable](https://github.com/bugsnag/bugsnag-android/blob/master/build.gradle#L15), so it's probably best to assume that anything that comes after `com.android.tools.build:gradle:` is a valid version string

fractalwrench (approved):
LGTM, pending a couple of tweaks to wording/detection of gradle snippets

PR #1138, opened 2020-11-23
[PLAT-5412] Cocoa mixed-type stacktraces

nickdowell:
## Goal

A change in the design of how React Native native promise rejections would be represented, from an event with multiple errors to a single event with stack frames of mixed types.

This depends on https://github.com/bugsnag/bugsnag-cocoa/pull/902

## Changeset

* Updated bugsnag-cocoa
* Changed `nativeStack` parsing to append native stack trace to top of stacktrace

## Testing

* Manually verified using React Native example app


bengourley (approved):
New `bugsnag-stacktrace-types` header present and exception/stackframe `type` values all as expected for RN promise rejections.

PR #1141, opened 2020-11-24
[PLAT-4932] feat(react-native-cli): Add install command

bengourley:
Adds `bugsnag-react-native-cli install` command.

- `git fetch && git checkout feature/rn-cli-install-cmd`
- `npm i`
- `npm run bootstrap`
- `npm run build`

On a React Native project:

- run `bugsnag-react-native-cli install`
- verify that Bugsnag is installed (note that it will be integrated in another command `insert` so won't actually do anything until then)

### DIFF
--- a/packages/react-native-cli/src/commands/InstallCommand.ts
+++ b/packages/react-native-cli/src/commands/InstallCommand.ts
@@ -1,0 +1,41 @@
+import prompts from 'prompts'
+import logger from '../Logger'
+import { install as npmInstall, detectInstalled, guessPackageManager } from '../lib/Npm'
+import { install as podInstall } from '../lib/Pod'
+import onCancel from '../lib/OnCancel'
+
+export default async function run (argv: string[], opts: Record<string, unknown>): Promise<void> {
+  const projectRoot = process.cwd()
+
+  try {
+    const alreadyInstalled = await detectInstalled('@bugsnag/react-native', projectRoot)
+    if (alreadyInstalled) {
+      logger.warn('@bugsnag/react-native is already installed, skipping')
+    } else {
+      logger.info('Adding @bugsnag/react-native dependency')
+      const { packageManager } = await prompts({
+        type: 'select',
+        name: 'packageManager',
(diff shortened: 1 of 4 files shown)


packages/react-native-cli/src/commands/InstallCommand.ts
```diff
@@ -0,0 +1,41 @@
...
+      const { packageManager } = await prompts({
+        type: 'select',
+        name: 'packageManager',
+        message: 'Using yarn or npm?',
+        choices: [
+          { title: 'npm', value: 'npm' },
+          { title: 'yarn', value: 'yarn' }
+        ],
+        initial: await guessPackageManager(projectRoot) === 'npm' ? 0 : 1
+      })
```
imjoehaines, line 25: I think every tool that I've used will just use yarn if there's a yarn lockfile and use NPM otherwise; do we need to ask or can we just use one?

packages/react-native-cli/src/lib/Pod.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!iosDirList.includes('Podfile')) {
+      logger.warn('No Podfile found in ios directory, skipping')
+      return
+    }
+  } catch (e) {
+    if (e.code === 'ENOENT') {
+      logger.warn('No ios directory found in project, skipping')
+      return
+    }
+    throw e
```
imjoehaines, line 18: I don't think there's a test for this case (where we re-throw)

imjoehaines (approved):

packages/react-native-cli/src/commands/InstallCommand.ts
bengourley: Good point, I'll create a follow-on ticket to update this (and the other location this logic is used) to just go ahead and use the detected package manager.

I figured there was no harm in asking, but when running `init` which will iterate through all of the steps, it will end up asking this twice, as the answer isn't shared between the `install` step and the `automate-symbolication` step (which wants to install `@bugsnag/source-maps`. And that would be annoying.

PR #1144, opened 2020-11-25
[PLAT-4934] Add configure command to RN CLI

bengourley:
Adds `bugsnag-react-native-cli install` command.

- `git fetch && git checkout feature/rn-cli-configure-cmd`
- `npm i`
- `npm run bootstrap`
- `npm run build`

On a React Native project:

- run `bugsnag-react-native-cli configure`
- verify that the Bugsnag API key that you were prompted for is added correctly to the Info.plist/AndroidManifest.xml

### DIFF
--- a/packages/react-native-cli/src/commands/ConfigureCommand.ts
+++ b/packages/react-native-cli/src/commands/ConfigureCommand.ts
@@ -1,0 +1,30 @@
+import prompts from 'prompts'
+import logger from '../Logger'
+import onCancel from '../lib/OnCancel'
+import { addApiKey as addApiKeyAndroid } from '../lib/AndroidManifest'
+import { addApiKey as addApiKeyIos } from '../lib/InfoPlist'
+
+export default async function run (argv: string[], opts: Record<string, unknown>): Promise<void> {
+  const projectRoot = process.cwd()
+
+  try {
+    const { apiKey } = await prompts({
+      type: 'text',
+      name: 'apiKey',
+      message: 'What is your Bugsnag API key?',
+      validate: value => {
+        return value.length > 1
+          ? true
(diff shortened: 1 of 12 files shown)


packages/react-native-cli/src/commands/ConfigureCommand.ts
```diff
@@ -0,0 +1,30 @@
...
+export default async function run (argv: string[], opts: Record<string, unknown>): Promise<void> {
+  const projectRoot = process.cwd()
+
+  try {
+    const { apiKey } = await prompts({
+      type: 'text',
+      name: 'apiKey',
+      message: 'What is your Bugsnag API key?',
+      validate: value => {
+        return value.length > 1
```
fractalwrench, line 16: Is there value in validating that the apiKey is a hexadecimal string of the correct length?

fractalwrench (approved):

packages/react-native-cli/src/lib/InfoPlist.ts
```diff
@@ -0,0 +1,41 @@
...
+    xcodeprojDir = (await fs.readdir(iosDir)).find(p => p.endsWith('.xcodeproj'))
+    if (!xcodeprojDir) {
+      logger.warn(UNLOCATED_PROJ_MSG)
+      return
+    }
+  } catch (e) {
+    logger.warn(UNLOCATED_PROJ_MSG)
+    return
+  }
+  const plistPath = path.join(iosDir, xcodeprojDir.replace(/\.xcodeproj$/, ''), 'Info.plist')
```
nickdowell, line 32: This works fine with react native's project template but may break if developers reorganise their project structure. Determining the path 100% correctly would involve parsing the Xcode project and finding the `INFOPLIST_FILE` build setting for the app target... tricky to implement, so this current solution is probably good enough!

nickdowell (approved):

PR #1148, opened 2020-11-26
Allow overriding unhandled state in callbacks

imjoehaines:
## Goal

This PR allows callbacks to modify the handled-ness of an event. This propagates to the associated session (if one exists) so that its handled/unhandled count is incremented _after_ the handled-ness has been changed. We also send the `unhandledOverridden` flag to the API, so that it knows a callback has modified the handled-ness

## Testing

Unit tests and MR tests for browser & node

### DIFF
--- a/packages/core/client.js
+++ b/packages/core/client.js
@@ -273,14 +273,18 @@ class Client {
     event._user = assign({}, event._user, this._user)
     event.breadcrumbs = this._breadcrumbs.slice()
 
-    if (this._session) {
-      this._session._track(event)
-      event._session = this._session
+    const trackSession = event => {
+      if (this._session) {
+        this._session._track(event)
+        event._session = this._session
+      }
     }
 
     // exit early if events should not be sent on the current releaseStage
     if (this._config.enabledReleaseStages !== null && !includes(this._config.enabledReleaseStages, this._config.releaseStage)) {
       this._logger.warn('Event not sent due to releaseStage/enabledReleaseStages configuration')
+      trackSession(event)
+
       return cb(null, event)
     }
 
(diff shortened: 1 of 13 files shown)


packages/core/client.js
```diff
@@ -273,14 +273,18 @@ class Client {
...
+      if (this._session) {
+        this._session._track(event)
+        event._session = this._session
+      }
     }
 
     // exit early if events should not be sent on the current releaseStage
     if (this._config.enabledReleaseStages !== null && !includes(this._config.enabledReleaseStages, this._config.releaseStage)) {
       this._logger.warn('Event not sent due to releaseStage/enabledReleaseStages configuration')
+      trackSession(event)
```
bengourley, line 286: I think this is unnecessary since the notifier will never send events nor sessions with this configuration.

packages/core/client.js
```diff
@@ -298,6 +302,8 @@ class Client {
 
       if (!shouldSend) {
         this._logger.debug('Event not sent due to onError callback')
+        trackSession(event)
```
bengourley, line 305: I'm tempted to say that this is unnecessary too, but the spec is unclear so I've started an internal discussion to get clarity.

packages/core/client.js
```diff
@@ -314,6 +320,13 @@ class Client {
         event._handledState.severityReason = { type: 'userCallbackSetSeverity' }
       }
 
+      if (event.unhandled !== event._handledState.unhandled) {
+        event._handledState.severityReason.unhandledOverridden = true
+        event._handledState.unhandled = event.unhandled
+      }
+
+      trackSession(event)
```
bengourley, line 328: If my other comments are ratified then you can just inline contents of `trackSession` here as the logic is only used once.

bengourley (commented):
Some spec-related minutia to resolve, but otherwise everything else looks good.

bengourley (approved):

PR #1149, opened 2020-11-26
[PLAT-4933] Add insert command to RN CLI

bengourley:
Adds `bugsnag-react-native-cli insert` command.

- `git fetch && git checkout feature/rn-cli-insert-cmd`
- `npm i`
- `npm run bootstrap`
- `npm run build`

On a React Native project:

- run `bugsnag-react-native-cli insert`
- verify that the Bugsnag initialisation has been correctly inserted into `index.js`, `MainApplication.java` and `AppDelegate.m`

### DIFF
--- a/packages/react-native-cli/src/lib/Insert.ts
+++ b/packages/react-native-cli/src/lib/Insert.ts
@@ -12,7 +12,7 @@ const COCOA_APP_LAUNCH_REGEX = /(-\s*\(BOOL\)\s*application:\s*\(UIApplication\s
 
 const BUGSNAG_JAVA_IMPORT = 'import com.bugsnag.android.Bugsnag;'
 const BUGSNAG_JAVA_INIT = 'Bugsnag.start();'
-const JAVA_APP_ON_CREATE_REGEX = /(public void onCreate\s*\(\)\s*\{[.\s]*super\.onCreate\([.\s]*\);(\s*))\S/
+const JAVA_APP_ON_CREATE_REGEX = /(public void onCreate\s*\(\)\s*\{[^]*super\.onCreate\(\);(\s*))\S/
 
 const DOCS_LINK = 'https://docs.bugsnag.com/platforms/react-native/react-native/#basic-configuration'
 const FAIL_MSG = (filename: string) =>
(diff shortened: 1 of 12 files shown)


nickdowell (approved):
Ran `bugsnag-react-native-cli insert` on a freshly created project.

LGTM 👍

packages/react-native-cli/src/lib/Insert.ts
```diff
@@ -0,0 +1,112 @@
...
+`import Bugsnag from "@bugsnag/react-native";
+Bugsnag.start();`
+
+const BUGSNAG_COCOA_IMPORT = '#import <Bugsnag/Bugsnag.h>'
+const BUGSNAG_COCOA_INIT = '[Bugsnag start];'
+const COCOA_APP_LAUNCH_REGEX = /(-\s*\(BOOL\)\s*application:\s*\(UIApplication\s\*\)\s*application\s+didFinishLaunchingWithOptions:\s*\(NSDictionary\s*\*\)launchOptions\s*\{\s*)\S/
+
+const BUGSNAG_JAVA_IMPORT = 'import com.bugsnag.android.Bugsnag;'
+const BUGSNAG_JAVA_INIT = 'Bugsnag.start();'
+const JAVA_APP_ON_CREATE_REGEX = /(public void onCreate\s*\(\)\s*\{[.\s]*super\.onCreate\([.\s]*\);(\s*))\S/
```
fractalwrench, line 15: There are a couple of scenarios I thought of which might fail if a user has customized their code a bit:

1. The user might decide to run some of their own code before calling `super.onCreate()`:

```
public void onCreate() {
    logSomething("onCreate() called");
    super.onCreate();
}
```

2. The `@Override` annotation is on the same line for some code styles:

```
@Override public void onCreate() {
```

My understanding is that it should be possible to support these use cases with the same approach - if not, they're probably not important enough to worry about.

fractalwrench (approved):

packages/react-native-cli/src/lib/Insert.ts
bengourley: 1. should already be catered for with `[.\s]*` before `super\.onCreate` in the regex?
2. should also already be catered for? The regex doesn't care what comes before `public void onCreate`

perhaps worth adding a test for these cases to demonstrate

bengourley: Regex tweaked and tests added to for these cases.

PR #1152, opened 2020-11-27
Build iOS test fixtures for React Native CLI testing

twometresteve:
## Goal

Starts to lay the tracks for testing that apps can build and run, sending reports to Bugsnag, after having the `react-native-cli` command run against them.

## Design

All steps in the React Native CLI pipeline are currently skipped, as there is no point adding time to the build and using CI resource in the short term.  However, I want to get the change in as it refactors aspects of the existing pipelines - and the PR is already getting quite large.

If the skip is removed, the IPA steps all pass (which you can see if you look in the build history).  The Android build needs more work and will be the subject on my next PR.  Later PRs will also then install the CLI and invoke `react-native-cli init` before building the app.

## Changeset

- New steps to build IPAs for RN 0.60 to 0.63
- Flipper removed from 0.62 onwards (to speed up build)
- Base Android build moved to main pipeline, as it will be needed by both the React Native and React Native CLI pipelines
- Minor style inconsistencies resolved
- Initial infrastructure added for Android, but further work is need to get it working properly

## Testing

No new tests as this stage, as long as the standard CI passes.


.buildkite/pipeline.yml
```diff
@@ -59,7 +75,10 @@ steps:
     async: true
 
   - label: 'Trigger React Native CLI pipeline'
-    depends_on: "publish-js"
+    depends_on:
+# TODO: Reinstate later
+#      - 'publish-js'
```
imjoehaines, line 80: Is later now or even later?

imjoehaines (approved):

.buildkite/pipeline.yml
twometresteve: Even later - I realised as I was checking through that I should have made it cleaner.  It should happen naturally though when we come in install and run the CLI.

PR #1153, opened 2020-11-30
Add docker setup for RN CLI testing

imjoehaines:
## Goal

Adds a basic docker setup for running Maze Runner against the RN fixtures

## Changeset

- `bin/minify` - remove dependency on `bash` so containers don't need to install it anymore
- Add `react-native-cli` docker-compose service
- Add ` dockerfiles/Dockerfile.react-native-cli` which builds the CLI and copies the packaged version into each RN fixture (*this uses the [Maze Runner integration branch](https://github.com/bugsnag/maze-runner/pull/185) for now*)
- Add a dockerfile for RN fixtures that installs `bugsnag-react-native-cli` globally and launches a shell
- Add a bare minimum `install.feature` which proves the CLI runs; this is just a template to build on

## Testing

Build the new service:

```sh
$ docker-compose build react-native-cli
```

and run it:

```sh
$ docker-compose run --use-aliases react-native-cli
```

The react native fixture can be chosen with the `REACT_NATIVE_VERSION` environment variable when running, e.g.

```sh
$ REACT_NATIVE_VERSION=rn0_63 docker-compose run --use-aliases react-native-cli
```


dockerfiles/Dockerfile.react-native-cli
```diff
@@ -0,0 +1,27 @@
...
+RUN npm pack --verbose packages/react-native-cli/
+
+# The maze-runner test image
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:integration-interactive_cli-cli as react-native-cli-maze-runner
+
+WORKDIR /app
+
+COPY test/react-native-cli test/react-native-cli
+COPY --from=react-native-cli-feature-builder /app/bugsnag-react-native-cli-*.tgz ./
+RUN for d in test/react-native-cli/features/fixtures/*/; do cp /app/*.tgz "$d"; done
```
twometresteve, line 25: We'll want this to install from the package we publish to Artifactory, although that may be a task for when it's integrated into the pipeline.  Perhaps you could add a TODO here for now?

test/react-native-cli/features/fixtures/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:lts-alpine
+
+COPY . /app
+
+WORKDIR /app
+
+RUN npm i -g bugsnag-react-native-cli-*.tgz
```
twometresteve, line 10: My comment above would also apply here, of course.

dockerfiles/Dockerfile.react-native-cli
twometresteve: ...the benefit of having it like this for now is that it will just work when running locally.  To run from a published artifact we'd need to facilitate and document overriding the version (like we've done for React Native).

imjoehaines: This is copied from [the node dockerfile](https://github.com/bugsnag/bugsnag-js/blob/next/dockerfiles/Dockerfile.node#L28-L29); is that changing too?

twometresteve: Eventually, yes, there's a ticket to update it at a later point.

twometresteve: The history on this was that we could use `pack` for Node and it was easier - but React Native forced our arm into using the published artifacts, due to Lerna not providing a `pack` command.  Publish/install is more representative of real life and of course we'd like it to be eventually consistent.

test/react-native-cli/features/install.feature
```diff
@@ -0,0 +1,6 @@
+Feature: install command
+
+Scenario: running 'install' on an empty fixture
+    When I run the service "rn0_60" interactively
```
twometresteve, line 4: Would be great if the RN version could be set by environment variable, then we'd have separate Buildkite jobs for each RN version, running in parallel.

imjoehaines: 👍 done in https://github.com/bugsnag/bugsnag-js/pull/1153/commits/f2404dc65f60d7cc04af5a94584f1f71c619a7a6

Checked that this runs the 0.63 fixture:

```
$ REACT_NATIVE_VERSION=rn0_63 DEBUG=true docker-compose run --use-aliases react-native-cli
```

and this runs 0.60

```
$ REACT_NATIVE_VERSION=rn0_60 DEBUG=true docker-compose run --use-aliases react-native-cli
```

twometresteve (approved):
LGTM

PR #1154, opened 2020-11-30
test(expo): Don't match against harcoded version in tests

bengourley:
Stop unit tests from breaking after a release (or prerelease, where it was noticed)


twometresteve (approved):
LGTM

PR #1158, opened 2020-12-01
feat(react-native-cli): Allow user to select BAGP version

bengourley:
To enable testing the RN CLI before an update to BAGP is released proper, it's useful to be able to specify the version desired. This PR adds a prompt for it, defaulting to the previously hardcoded value (`5.+`).


tomlongridge (approved):
LGTM

twometresteve (approved):
LGTM.  I've tested it locally and whilst I hit problems later they're not something that we would seek to address here.

PR #1159, opened 2020-12-01
Update RN Interactive CLI tests to use an NPM registry

twometresteve:
## Goal

Update the interactive React Native CLI tests to consume the packages we publish to Artifactory as part of CI.  

## Design

This is more representative of real life use and will eventually lead to a consistent approach across the notifier repo.

## Changeset

TBD

## Testing

Will be covered by passing CI.


test/react-native-cli/features/fixtures/Dockerfile
```diff
@@ -1,15 +1,36 @@
...
 
-RUN npm i -g bugsnag-react-native-cli-*.tgz
+# Artifcatory authentication
+ARG REGISTRY_URL
+ARG REG_BASIC_CREDENTIAL
+ARG REG_NPM_EMAIL
+RUN npm config set registry $REGISTRY_URL
+RUN npm config set _auth $REG_BASIC_CREDENTIAL
+RUN npm config set email $REG_NPM_EMAIL
+RUN npm config set always-auth = true
```
imjoehaines, line 19: Artifactory hosts packages as `tgz` files, so I wonder whether we can keep the old way of installing the package, but provide a different tarball on CI? That way we can build the package locally without having to deal with artifactory at all — I think it's more valuable to be easier to run locally than on CI vs the other way around

e.g. currently to run all of these tests locally, all you need to do is run two commands & the only setup is to login to docker to get access the maze runner images (which will always be needed);

```sh
$ docker-compose build react-native-cli
$ REACT_NATIVE_VERSION=rn0_63 docker-compose run react-native-cli
```

If we do go this route then how to get this to work needs to be documented in `TESTING.md`

docker-compose.yml
```diff
@@ -141,20 +141,24 @@ services:
...
+      REGISTRY_URL:
+      REG_NPM_EMAIL:
+      REG_BASIC_CREDENTIAL:
     networks:
       default:
         aliases:
           - maze-runner
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
+      - ./test/react-native-cli/features/:/app/features/
```
imjoehaines, line 161: Doesn't this mean any changes that the tests make will be reflected in your local file system? The tests will make _a lot_ of changes to files, so we probably don't want this unless there's a way to exclude the `fixtures` directory

```suggestion
```

docker-compose.yml
```diff
@@ -141,20 +141,24 @@ services:
...
-  react-native-cli:
-    build:
-      context: .
-      dockerfile: dockerfiles/Dockerfile.react-native-cli
+  react-native-cli-tool-maze-runner:
+    # TODO Use Maze Runner release
+    image: 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:integration-interactive_cli-cli
     environment:
       DEBUG:
+      VERBOSE:
```
imjoehaines, line 149: Is `VERBOSE` used by anything? It looks the same as running without it to me

```suggestion
```

test/react-native-cli/features/fixtures/Dockerfile
```diff
@@ -1,15 +1,36 @@
...
+
+ARG REACT_NATIVE_VERSION
+WORKDIR /app/$REACT_NATIVE_VERSION
+
+COPY test/react-native-cli/features/fixtures/$REACT_NATIVE_VERSION $REACT_NATIVE_VERSION
+
+# Install the CLI
+ARG NOTIFIER_VERSION
+ENV NOTIFIER_VERSION=$NOTIFIER_VERSION
+RUN run-func ../scripts/react-native-cli-helper.js installCli
```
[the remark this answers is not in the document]
twometresteve: Yep, I think this was a big hangover from a previous setup where we didn't have a node-based image.

test/react-native-cli/features/steps/steps.rb
```diff
@@ -21,6 +21,6 @@
   ERROR
 end
 
-When('I run the React Native service interactively') do
-  step("I run the service '#{current_fixture}' interactively")
+When('I run the React Native CLI service interactively') do
+  step("I run the service 'react_native_cli' interactively")
 end
```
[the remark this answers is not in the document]
twometresteve: Agreed - and it leaves the rest of the file deserving a more conventional home.  Have moved to `support/env.rb`.

docker-compose.yml
twometresteve: VERBOSE and DEBUG are synonymous in Maze and I'm just following the established pattern here.  We should probably just take one of them when we next make a breaking change to reduce the amount of config kicking around.

docker-compose.yml
twometresteve: The intention is that the Maze Runner container (which this configures) only reads the `.feature` files.  The test fixture directories are copied into their own Docker images as part of running the tests.  So I wouldn't expect this to result in changes to the local file system when running locally - but will check!

test/react-native-cli/features/fixtures/Dockerfile
twometresteve: Speaking to @bengourley a couple of times, the general feeling was that NPM publish and install from a registry, rather than using `pack` is more representative of real life.  It also means we can work towards a consistent approach across the repo (React Native cannot use `pack`).  It's also worth noting that you don't have to use Artifactory for that - Verdaccio is a lightweight NPM registry that serves nicely for local use.

That all said, we could consider providing both as options, assuming the plumbing didn't become too convoluted.  And you're certainly right about updating `TESTING.md` - it has been on my mind to do, but I'd been waiting for things to settle a bit and it probably becomes easier to do once the MazeRunner changes are released.

BTW, if you update the `Gemfile` locally to point at the MR integration branch (or local checkout), you don't need to build the docker image at all - you can just run:
```
REACT_NATIVE_VERSION=rn0_63 bundle exec maze-runner
```

docker-compose.yml
imjoehaines: I'm not sure that's true; running with `VERBOSE` doesn't seem to do anything

twometresteve: Ah, I was referring to the core MazeRunner logger itself, which contains:
```
  def initialize
    if ENV['VERBOSE'] || ENV['DEBUG']
      super(STDOUT, level: Logger::DEBUG)
```

But in this context it'll be limited by the absence of `VERBOSE` from the secondary `docker-compose.yml` file - for the test fixtures.  TBH I only put it in for consistency with existing entries, so let's talk IRL and see if we can just get rid of it from everywhere.

docker-compose.yml
imjoehaines: Yeah, I think you're right 👍

test/react-native-cli/features/fixtures/Dockerfile
```diff
@@ -2,14 +2,34 @@ FROM node:lts-alpine
...
+RUN npm config set registry $REGISTRY_URL
+RUN npm config set _auth $REG_BASIC_CREDENTIAL
+RUN npm config set email $REG_NPM_EMAIL
+RUN npm config set always-auth = true
 
-# TODO(PLAT-5515) 'pod install' is currently always run but will only work on
-# macOS, so we stub it out with a script that does nothing
-RUN printf '#!/usr/bin/env sh\nprintf "this is a stub of $0\n"' > /usr/bin/pod && chmod +x /usr/bin/pod
+# Files needs for installing the CLI
+COPY .git .git
```
imjoehaines, line 21: I guess we need this because of the `installCli` command? Can we remove it after or put it somewhere away from the fixtures (outside `/app`) as it will break the tests — the CLI checks if we're in a git repo and the tests assume that we're not unless it creates one

twometresteve: Ah, ok. interesting and good spot.  I think there are a few approaches we could potentially take:
- Me find another way of passing the version in explicitly using `NOTIFIER_VERSION` (it's `determineVersion` that will use it)
- Remove the .git context in the docker file afterwards
- Create a Cucumber step to explicitly remove any git context it might find

I'm happy to look at the first, but I think the last also has some merit in terms of making the test robust.

imjoehaines: I think you could also copy this & run `installCli` somewhere other than `/app` for it to work — `installCli` runs globally so it doesn't have to happen in the same directory as the fixture projects

twometresteve: Yeah, somehow I skipped over you telling me that in the initial comment.  I've moved it all into `/test` now, with just the fixture going into `/app`.

twometresteve:
We'll come back to this another day.

PR #1162, opened 2020-12-02
fix(react-native-cli): Ensure android init receives "this" arg

bengourley:
Add missing `this` param.


twometresteve (approved):
Haven't specifically tested this, but LGTM based on the change I needed to make to get the app to build.

PR #1166, opened 2020-12-02
RN-CLI: Tests for automate-symbolication command

bengourley:
## Goal

Adds tests for the RN CLI 'automate-symbolication' command:

- successful modification
- successful modification, selecting desired `@bugsnag/source-maps` version
- opting out of iOS modification
- opting out of Android modification
- opting out of both

## Testing

Manually run against the four RN fixtures

PR #1168, opened 2020-12-03
[PLAT-5501] RN-CLI: Have confidence in the guessed npm client

bengourley:
Remove the confirmations as it's an unncessary blocking step.

Tests in #1166 and ~#1157~ will need to be updated when this lands on the integration branch.


imjoehaines (approved):
I just merged the install command tests so they'll need to be updated in this PR (sorry!), otherwise this looks good

PR #1169, opened 2020-12-03
Run React Native CLI tests in pipeline

twometresteve:
## Goal

Automates the running of the React Native CLI test in the pipeline.

## Design

This uses the `npm pack` approach for installing the notifier.  Switching to npm (Lerna) publish/install will follow in a later PR.

## Changeset

- New pipeline steps
- Docker Compose Buildkite plugin bumped
- Docker Compose/Dockerfile renamed to distinguish from the forthcoming "build and run" tests

## Testing

Now covered by passing CI.

### DIFF
--- a/.buildkite/react-native-cli-pipeline.yml
+++ b/.buildkite/react-native-cli-pipeline.yml
@@ -57,14 +57,6 @@ steps:
     env:
       REACT_NATIVE_VERSION: "rn0_63"
 
-  - label: ':runner: Broken tests'
-    depends_on: "cli-maze-image"
-    timeout_in_minutes: 20
-    plugins:
-      docker-compose#v3.7.0:
-        run: react-native-cli-tool-maze-runner
-        use-aliases: true
-
   #
   # Built app test fixtures
   #
--- a/.buildkite/react-native-cli-pipeline.yml
+++ b/.buildkite/react-native-cli-pipeline.yml
@@ -24,6 +24,7 @@ steps:
       docker-compose#v3.7.0:
         run: react-native-cli-tool-maze-runner
         use-aliases: true
+        command: ["-e", "features/built_app.feature"]
     env:
       REACT_NATIVE_VERSION: "rn0_60"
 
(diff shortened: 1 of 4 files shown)


.buildkite/react-native-cli-pipeline.yml
```diff
@@ -1,23 +1,81 @@
...
+  # CLI tests
+  #
+  - label: ':runner: RN 0.60 CLI tests'
+    depends_on: "cli-maze-image"
+    timeout_in_minutes: 20
+    plugins:
+      docker-compose#v3.7.0:
+        run: react-native-cli-tool-maze-runner
+        use-aliases: true
+        command: ["-e", "features/built_app.feature"]
```
twometresteve, line 27: I haven't added this file yet, but it's coming in a later change.

imjoehaines, line 27: We might want to add `--retry` here to avoid the known flake

imjoehaines (approved):

.buildkite/react-native-cli-pipeline.yml
twometresteve: Ok, we can consider adding that as and when needed.

docker-compose.yml
```diff
@@ -141,10 +141,10 @@ services:
       - ./test:/app/test
       - ./features:/app/features
 
-  react-native-cli:
+  react-native-cli-tool-maze-runner:
     build:
       context: .
-      dockerfile: dockerfiles/Dockerfile.react-native-cli
+      dockerfile: dockerfiles/Dockerfile.react-native-cli-tool
```
bengourley, line 147: what's with the new name?

bengourley (approved):

docker-compose.yml
twometresteve: I'm anticipating needing to add a separate Dockerfile for the MazeRunner "end-to-end" (i.e. built and run-on-device) tests, so this is seeking to distinguish from that.

PR #1173, opened 2020-12-03
fix(react-native): Fix propagation of enabledReleaseStages from Cocoa notifier

nickdowell:
## Goal

Fixes the propagation of `enabledReleaseStages` from the Cocoa notifier.

`enabledReleaseStages` is an NSSet, which is not a JSON compatible type.

## Changeset

Converting `enabledReleaseStages` to an array fixes the issue.

## Testing

Manually verified using a JavaScript remote debugger to ensure `enabledReleaseStages` is correctly populated on the JS side.


bengourley (approved):
Tested locally and I now see the `enabledReleaseStages` config in the JS layer when populated 👍

PR #1177, opened 2020-12-04
Allow notifier time to synch state with native layer

twometresteve:
## Goal

Attempts to resolve a persistent flaking test that requires JS layer state to be synched with the native layer.

## Testing

The scenario in question was flaking for me 3/5 before the change.  With it in it has passed 10/10 times.

### DIFF
--- a/.buildkite/react-native-pipeline.yml
+++ b/.buildkite/react-native-pipeline.yml
@@ -223,7 +223,6 @@ steps:
         - --fail-fast
     env:
       SKIP_NAVIGATION_SCENARIOS: "true"
-      DEBUG: "true"
     concurrency: 9
     concurrency_group: 'browserstack-app'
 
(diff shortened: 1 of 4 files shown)


nickdowell (approved):
Nice 🙏

Any idea why it was flaking with rn0.63 but not rn0.60?

fractalwrench (approved):

twometresteve:
> Nice 🙏
> 
> Any idea why it was flaking with rn0.63 but not rn0.60?

Not specifically, but if it's a race condition that we're talking about I can believe that the plumbing and mechanics of rn0.63 simply take more time that 0.60.

Sadly though the story doesn't seem to be over on this PR, as it's flagged up a persistent failure in at least one other scenario, which wasn't included in my trials.  Very surprising.

nickdowell (approved):

fractalwrench (approved):

PR #1178, opened 2020-12-04
RN CLI: Add TODOs to steps that should be in Maze Runner

imjoehaines:
## Goal

Add TODOs to steps that should be in Maze Runner (see PLAT-5565)

Some of these steps will need less ambiguous names, e.g. "I press enter" could also be valid in a web browser


imjoehaines:
@twometresteve just checking we agree which steps should/shouldn't be in Maze Runner. Any suggestions for better names would be great too!

twometresteve (approved):
Agreed that these belong in MR.  I think we can defer the choice of step naming until we actually come to move them.

PR #1189, opened 2020-12-10
test: tidy test output and fix open handles

djskinner:
- Set the `roots` property in the jest config as a potential performance improvement. I noticed some haste-map warnings coming from the test/ director. I'm not sure why jest is looking in there but this might prevent it.
- Mock console to clean up the test output. We might want to consider asserting on calls being made in some cases though I haven't done that currently.
- Ensures open handles are closed, which fixes the "A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks." warning

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,100 +1,86 @@
 const testsForPackage = (packageName) => `<rootDir>/packages/${packageName}/**/*.test.[jt]s?(x)`
 
-module.exports = {
+const package = (displayName, packageNames, config = {}) => ({
   roots: ['<rootDir>/packages'],
+  displayName,
+  testMatch: packageNames.map(testsForPackage),
+  ...config
+})
+
+module.exports = {
   projects: [
-    {
-      displayName: 'core',
-      testMatch: [
-        testsForPackage('core')
-      ]
-    },
-    {
-      displayName: 'shared plugins',
-      testMatch: [
-        testsForPackage('plugin-app-duration')
-      ]
-    },
-    {
-      displayName: 'browser',
-      testMatch: [
-        testsForPackage('browser'),
-        testsForPackage('delivery-x-domain-request'),
-        testsForPackage('delivery-xml-http-request'),
-        testsForPackage('plugin-react'),
-        testsForPackage('plugin-vue'),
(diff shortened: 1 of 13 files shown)


packages/plugin-angular/package.json
```diff
@@ -25,7 +25,6 @@
     "build": "npm run clean && npm run build:esm2015 && npm run build:esm5",
     "build:esm2015": "ngc -p tsconfig.json",
     "build:esm5": "ngc -p tsconfig.esm5.json",
-    "test": "npm run build",
```
djskinner, line 28: I'm assuming we don't need to run this on test but I could be wrong. It can be reinstated along with the related package.json change.

bengourley: This "tests" the package by checking it compiles. We could do `tsc --noEmit` instead to make it more explicit?

jest.config.js
```diff
@@ -1,6 +1,7 @@
 const testsForPackage = (packageName) => `<rootDir>/packages/${packageName}/**/*.test.[jt]s?(x)`
 
 module.exports = {
+  roots: ['<rootDir>/packages'],
```
bengourley, line 4: According to the CI run, this doesn't prevent a lot of jest-haste-map collisions being output: https://buildkite.com/bugsnag/bugsnag-js/builds/3362#12b5bdbe-8113-4fa1-ba74-7d2ff119670d/122-128

We could give all those package.jsons unique names, but maybe there is some more tweaking of jest config to be done that would prevent it from looking there…

bengourley (changes requested):
Looks good. Couple of tweaks needed I think.

jest.config.js
djskinner: I've sorted this now. roots needed setting for every jest project

packages/plugin-angular/package.json
djskinner: I've changed the script name from 'test' to 'test:types' so it gets run with the other type tests

bengourley (approved):

PR #1191, opened 2020-12-10
RN CLI: Install Bugsnag Android Gradle plugin in the 'install' command

imjoehaines:
## Goal

Currently when the user answers "no" to automatic source map upload, the Bugsnag Android Gradle plugin is not installed. This leaves the user with a broken project, because BAGP is required for the notifier to work (it adds a mavenLocal repo so that the gradle build picks up the vendored android notifier).

Now the `install` command will add BAGP to the project and the `automate-symbolication` command will only enables the `uploadReactNativeMappings` flag in the Gradle config (either by adding a new `bugsnag` section, or adding the flag to an existing one)

### DIFF
--- a/packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
+++ b/packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
@@ -1,9 +1,9 @@
 import prompts from 'prompts'
 import logger from '../Logger'
-import { modifyRootBuildGradle, modifyAppBuildGradle } from '../lib/Gradle'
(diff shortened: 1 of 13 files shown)


bengourley (changes requested):
Implementation looks good.

The `enableReactNativeMappings` function isn't covered by unit tests at all:

image

Can you add some tests to`Gradle.test.ts` for at least:

- asserting changes are made as expected
- asserting no changes are made if the pattern can't be detected and warning is output
- asserting no changes are made if the changes already exist

bengourley (approved):

PR #1192, opened 2020-12-10
Add E2E test for changing unhandled in React Native

fractalwrench:
## Goal

Adds an E2E test that verifies `event.unhandled` can be overridden on React Native. Currently this only runs on Android - a separate PR will be raised to achieve this for Cocoa.

## Changeset

- Bumped bugsnag-android to use the following changeset: https://github.com/bugsnag/bugsnag-android/pull/1032
- Added React Native scenarios which override the unhandled state in both directions
- Ensured that the tests run against Android only, as the Cocoa changeset has not been integrated into bugsnag-js yet


test/react-native/features/fixtures/app/scenario_js/app/scenarios/HandledOverrideJsErrorScenario.js
```diff
@@ -0,0 +1,12 @@
+import Scenario from './Scenario'
+import Bugsnag from '@bugsnag/react-native'
+
+export class HandledOverrideJsErrorScenario extends Scenario {
+  async run () {
+    Bugsnag.startSession()
+    await this.timeout(750)
```
bengourley, line 7: Are these timeouts just to ensure the event is always the second request received by the mock server?

bengourley (approved):

test/react-native/features/fixtures/app/scenario_js/app/scenarios/HandledOverrideJsErrorScenario.js
fractalwrench: Yep, that's the case

PR #1197, opened 2020-12-14
Bump to Cocoa library to 6.4.1

kstenerud:
Changes:

- Ran update.ios script to import cocoa
- Modified BugsnagEventDeserializer.m to update for changed API


packages/react-native/ios/BugsnagReactNative/BugsnagEventDeserializer.m
```diff
@@ -156,9 +157,11 @@ - (BugsnagHandledState *)deserializeHandledState:(NSDictionary *)payload {
 
     BSGSeverity severity = BSGParseSeverity(payload[@"severity"]);
     BOOL unhandled = [payload[@"unhandled"] boolValue];
+    BOOL unhandledOverridden = [severityReason[@"unhandledOverridden"] boolValue];
     return [[BugsnagHandledState alloc] initWithSeverityReason:reason
                                                       severity:severity
                                                      unhandled:unhandled
+                                           unhandledOverridden:unhandledOverridden
```
kstenerud, line 164: Changes to cocoa internal API required these fixes RN-side.

packages/react-native/ios/BugsnagReactNative/BugsnagEventDeserializer.m
```diff
@@ -156,9 +157,11 @@ - (BugsnagHandledState *)deserializeHandledState:(NSDictionary *)payload {
 
     BSGSeverity severity = BSGParseSeverity(payload[@"severity"]);
     BOOL unhandled = [payload[@"unhandled"] boolValue];
+    BOOL unhandledOverridden = [severityReason[@"unhandledOverridden"] boolValue];
```
bengourley, line 160: what is the behaviour of this when

a. `severityReason` is `nil`?
b. `severityReason.unhandledOverridden` doesn't exist, or isn't a boolean?

bengourley (approved):
Approving subject to satisfactory answers to my queries inline…

packages/react-native/ios/BugsnagReactNative/BugsnagEventDeserializer.m
kstenerud: Messages to nil objects in ObjC return 0, which is converted to boolean false in this case.
If `unhandledOverridden` is a string, number, boolean, or nil, it will return Objective-C's interpretation of that type as a boolean value. If it's any other type, it will throw an exception.

fractalwrench (approved):
These changes LGTM. There needs to be a separate PR that adapts the existing React Native E2E tests before we can release the unhandled override work.

PR #1199, opened 2020-12-15
RN CLI: support on premise URLs

imjoehaines:
## Goal

Adds prompts for custom URLs to `configure` (notify/session) and `automate-symbolication` (upload) to support Bugsnag on premise. These default to the `x.bugsnag.com` endpoints and if the default is used then no additional config is added


bengourley (approved):
Brilliant work. Tested out locally, working as expected.

PR #1202, opened 2020-12-15
e2e tests for running a CLI initialized RN app

twometresteve:
## Goal

Completes the end-to-end tests for dynamically initializing and building a React Native app, ensuring that JavaScript and Native errors are received when the app is run.

Also includes a small fix to the tool, ensuring that `bugsnag-react-native-xcode.sh` is included in the install.

## Design

There are a few TODOs that need to be revisited once other tools are released and the functionality of the tool is completed.

For now, source map uploads are disabled for both Android and iOS builds - checking of these will follow in a later PR.  Instructions for running locally will also be added later, as there is likely to be too much change to make it worthwhile doing now.

## Changeset

- New plumbing for dynamically initializing RN projects using the RN CLI.  Uses Expect to interact with the shell.
- New steps to run the e2e tests for each supported version of React Native, on the latest available iOS and Android versions.
- Test scenario that invokes the reporting of handled JS and Native errors.
- RN CLI test fixtures updated to support JS/Native handled errors off of button presses.
- New Cucumber steps to support button presses

## Testing

Covered by CI and run locally successfully prior to review.

### DIFF
--- a/test/react-native-cli/features/built-app.feature
+++ b/test/react-native-cli/features/built-app.feature
@@ -13,4 +13,4 @@ Scenario: Build app sends JavaScript and Native handled errors
   And the event "unhandled" is false
   And the event "exceptions.0.errorClass" equals the platform-dependent string:
     | android | java.lang.Exception |
-    | ios     | NSException         |
+    | ios     | NSError             |
--- a/test/react-native-cli/features/built-app.feature
+++ b/test/react-native-cli/features/built-app.feature
@@ -1,6 +1,16 @@
 Feature: Tests against a React Native app that was initialized using the Bugsnag React Native CLI
 
 Scenario: Build app sends JavaScript and Native handled errors
-  When I cause a handled JavaScript error
-  And I cause a handled native error
-  And I wait to receive 2 requests
+  When I notify a handled JavaScript error
+  And I wait to receive a request
+  Then the event "unhandled" is false
+  And the event "exceptions.0.errorClass" equals "ReferenceError"
+  And the exception "type" equals "reactnativejs"
+  And I discard the oldest request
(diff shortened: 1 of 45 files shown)


test/react-native-cli/features/built-app.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Tests against a React Native app that was initialized using the Bugsnag React Native CLI
+
+Scenario: Build app sends JavaScript and Native handled errors
+  When I notify a handled JavaScript error
```
bengourley, line 4: I can't see any configuration that prevents sessions from sending, so I would have expected a session to be the first payload that's delivered?

It seems like this is a potential avenue for flakes, seeing as the order of delivery might be non-deterministic. I think that checking for the delivery of a session is another useful data point in determining whether the app was configured correctly, so I'd lean heavily in favour of that over preventing sessions from being sent (which is the approach taken in many of our end-to-end tests).

bengourley (approved):

test/react-native-cli/features/built-app.feature
twometresteve: I completely agree and there's some going on here that I don't yet understand.  I'm not seeing a session request at all currently, even if I just sit and wait for two minutes.

twometresteve: This was due to MazeRunner clearing the set of received in the Cucumber `Before` hook - which happens after the RN CLI test fixture is first opened and the session request sent.  MazeRunner has now been updated to reset its list of received requests in the `After` hook, avoiding this situation whilst still achieving its core objective.

bengourley (approved):

PR #1204, opened 2020-12-15
Fix Node Maze Runner tests on Maze Runner v3.6.0

imjoehaines:
## Goal

The Node Maze Runner tests were failing due to the "I should receive no requests" step receiving a session payload. These tests now expect the session payload and are now passing on the latest v3 MR release


twometresteve (approved):
LGTM

PR #1205, opened 2020-12-16
[PLAT-5575]: Ensure init exits after any subcommand fails

bengourley:
## Goal

If any of the subcommands fails in a fatal way, it shouldn't carry on. e.g. when an npm install fails due to a non-existant version.

## Design

Despite detecting and logging the failure, the subcommand did not pass the information back up to the `init` command, so it carried on executing tasks.
I updated the return type of each of the commands to `Promise<boolean>` to signify sucess (`true`) or failure (`false`) so that `init` could know whether to continue or not.

Perhaps I could have used an enum to make it more clear what the boolean means?

## Changeset

- Updated the commands with the new return type
- Updated `init` to check the return value before running the next command

## Testing

Tested manually. An end to end test should be added when the tests for `init` land.

### DIFF
--- a/packages/react-native-cli/src/bin/cli.ts
+++ b/packages/react-native-cli/src/bin/cli.ts
@@ -49,10 +49,11 @@ export default async function run (argv: string[]): Promise<void> {
     switch (opts.command) {
       case 'init':
         await repoStatePreCommand(remainingOpts, projectRoot, opts)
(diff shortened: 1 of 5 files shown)


imjoehaines (approved):
> Perhaps I could have used an enum to make it more clear what the boolean means?

I think a boolean is fine for success/failure; we can always expand it in future if we need to

PR #1207, opened 2020-12-16
[PLAT-5618] RN-CLI: Let source map lib add the path

bengourley:
## Goal

One last change to this shell script – the source map lib [now determines the correct path](https://github.com/bugsnag/bugsnag-source-maps/pull/48).

## Testing

Tested manually. Will also be tested via e2e RN CLI tests when they land.


nickdowell (approved):
Nice; that simplifies things 👍

PR #1208, opened 2020-12-16
Avoid building unnecessary ABIs in test fixture

fractalwrench:
## Goal

Reduces the size of the mazerunner test fixture by excluding the `x86_64` and `armeabi_v7a` architectures. We don't actively test with these so excluding them makes sense to decrease the time spent waiting for APKs to upload.

This is based on the approach taken in the following Android PR: https://github.com/bugsnag/bugsnag-android/pull/1054


twometresteve (approved):
LGTM.  This will personally save me a fair amount of time waiting for app uploads to BrowserStack with my slow internet.

PR #1212, opened 2020-12-17
Ask for on-premise URLs up front, gated behind a confirmation prompt

imjoehaines:
## Goal

Changes the way we ask for on-premise URLs; there is now an initial "Are you using Bugsnag on-premise?" question and URLs are requested only if the user answers "yes"

We still only ask for the URLs that are required for the command that is running; `notify` & `session` for `configure`, `upload` & `build` for `automate-symbolication` and all four for `init`

### DIFF
--- a/packages/react-native-cli/src/commands/OnPremiseUrlsCommand.ts
+++ b/packages/react-native-cli/src/commands/OnPremiseUrlsCommand.ts
@@ -1,0 +1,48 @@
+import prompts, { PromptObject } from 'prompts'
+import { UrlType, OnPremiseUrls } from '../lib/OnPremise'
+import onCancel from '../lib/OnCancel'
+
+export default async function run (...requiredUrls: UrlType[]): Promise<OnPremiseUrls> {
+  const { isUsingOnPremise } = await prompts({
+    type: 'confirm',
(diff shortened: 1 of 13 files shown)


packages/react-native-cli/src/commands/OnPremiseUrlsCommand.ts
```diff
@@ -0,0 +1,48 @@
+import prompts, { PromptObject } from 'prompts'
+import { UrlType, OnPremiseUrls } from '../lib/OnPremise'
+import onCancel from '../lib/OnCancel'
+
+export default async function run (...requiredUrls: UrlType[]): Promise<OnPremiseUrls> {
```
bengourley, line 5: out of interest why did you go for a spread here rather than take a single array argument?

bengourley (approved):

packages/react-native-cli/src/commands/OnPremiseUrlsCommand.ts
imjoehaines: No real reason; I think it looks neater when it's the only parameter & saves 2 characters but I don't think there's an actual advantage

PR #1216, opened 2020-12-21
RN CLI: Move BAGP apply below react.gradle apply

imjoehaines:
## Goal

The previous order meant BAGP couldn't detect that it was running in a React Native project. It must be applied after react.gradle for this to work properly

This also allows us to enable source map uploads in the APK build on CI (which is part of this PR)


imjoehaines:
Merged in #1217

PR #1217, opened 2020-12-22
e2e tests and code fix for source maps upload during Android build

twometresteve:
## Goal

Adds an end-to-end test that expects a build request and source map upload during an Android build.  Includes a code fix for the insertion of Gradle script elements in the right location.

## Design

The existing Docker file used to do the Android build has been updated to also allow it to run Maze Runner as a Gem.  I've bumped it to Debian Buster (v10), which was the latest I could find within a reasonable amount of time.  It only has Ruby 2.5 on it, so I've needed to make a small tweak due to a 2.6 language feature that was being used (endless ranges).

## Changeset

1. Existing feature file renamed and contents reworded to try and distinguish between tests that apply during the build process and tests for when the app is run.
2. Existing Android build step updated to perform the build in the context of a Maze Runner test - new Cucumber feature/steps added to support.
3. Some changes made to support running tests locally, which will be properly rounded off and documented in a later PR.
4. Analogous changes for iOS laid out, but not active as yet.

## Testing

Additional test now covered by CI.  It is possible (and not difficult) to run the test locally, although full documentation for it will be provided later.


.buildkite/react-native-cli-pipeline.yml
```diff
@@ -24,7 +24,7 @@ steps:
       docker-compose#v3.7.0:
         run: react-native-cli-tool-maze-runner
         use-aliases: true
-        command: ["-e", "features/built-app.feature"]
+        command: ["-e", "build-android-app.feature", "-e", "features/run-and-notify.feature"]
```
twometresteve, line 27: I think it would be cleaner if I just moved the feature files into separate directories.  I'll do that as part of working on the iOS test.

dockerfiles/Dockerfile.android-builder-base
```diff
@@ -1,5 +1,5 @@
 # TODO: Candidate for moving to its own repo for sharing across notifier repositories
```
twometresteve, line 1: Probably not true anymore - I'll remove the TODO in the next change.

scripts/react-native-cli-helper.js
```diff
@@ -5,7 +5,10 @@ const common = require('./common')
 const fs = require('fs')
 
 module.exports = {
-  buildAndroid: function buildAndroid (sourceFixtures, destFixtures) {
+  buildAndroid: function buildAndroid (sourceFixturesIn, destFixturesIn) {
+    const baseDir = process.env.PWD
```
twometresteve, line 9: This made it a little easier to be agnostic of whether it was being run locally or in a Docker container on CI.

test/react-native-cli/features/fixtures/rn-cli-init-android.sh
```diff
@@ -33,9 +35,8 @@ send -- $notifierVersion\r
 expect "If you want the latest version of the Bugsnag Android Gradle plugin hit enter, otherwise type the version you want"
 send -- 5.5.0-alpha01\r
 
-# TODO: Use the usual 123123... test API key
 expect "What is your Bugsnag API key?"
-send -- b161f2dbabe3204527bbe5ed4b9334a4\r
+send -- 1234567890ABCDEF1234567890ABCDEF\r
```
twometresteve, line 39: There's a known issue on Android if you uses a purely numeric API key like 123123123123123..... - the entry in the manifest file gets interpreted as a Float and it falls over in a big heap.

test/react-native-cli/features/steps/steps.rb
```diff
@@ -100,14 +106,15 @@
...
+  stdout_lines = Runner.interactive_session.stdout_lines
+  length_before = stdout_lines.length
 
   steps %Q{
     When I input "cat package.json" interactively
     Then I wait for the shell to output '"dependencies": \{' to stdout
   }
 
-  after = Runner.interactive_session.stdout_lines[length_before..]
+  after = stdout_lines[length_before..stdout_lines.length]
```
imjoehaines, line 117: The endless range should do the same thing as this; was it broken?

test/react-native-cli/features/steps/steps.rb
```diff
@@ -69,6 +69,12 @@
   )
 end
 
+When('I build the Android app') do
+  # Handle both Dockerized and local Maze Runner executions
+  path = (File.exist? 'scripts/react-native-cli-helper.js') ? '.' : '../..'
+  `node -e 'require("#{path}/scripts/react-native-cli-helper").buildAndroid("./features/fixtures", "./local-build")'`
```
imjoehaines, line 75: Where is this relative to? `scripts` is four levels up locally so this is a bit confusing; I assume it works because the PWD would usually be `test/react-native-cli`?

I think this would be more resilient with an absolute path and check that the file exists in both cases so it's easier to debug if the file goes missing, e.g.

```suggestion
  # Handle both Dockerized and local Maze Runner executions

  possible_locations = [
    "#{__dir__}/../../scripts/react-native-cli-helper.js",
    "#{__dir__}/../../../../scripts/react-native-cli-helper.js",
  ]

  path = possible_locations.find { |path| File.exist?(path) }

  if path.nil?
    raise <<~ERROR
      The React Native helper script was not found in any of the expected locations:
        - #{possible_locations.join("\n  -")}
    ERROR
  end

  `node -e 'require("#{path}").buildAndroid("./features/fixtures", "./local-build")'` 
```

imjoehaines (approved):

test/react-native-cli/features/steps/steps.rb
twometresteve: Endless ranges were added in Ruby 2.6 and the Debian 10 image only has 2.5 on it.  I could probably have worked harder and found a newer image, but it felt like diminishing returns when I could just tweak the code.

imjoehaines: Yeah that makes sense, just checking I understood why

test/react-native-cli/features/steps/steps.rb
twometresteve: Yeah, it was working due to PWD being as you say.  I've updated in line with your suggestion on my next set of changes (primarily for iOS) so we can get the code fix in ASAP.

PR #1218, opened 2020-12-22
Source map upload test for iOS app build

twometresteve:
## Goal

Adds a test scenario to verify that a source map is uploaded as part of an RN CLI `init`ed iOS app. 

## Design

This change refactors the existing build step so that it if invoked as part of a Maze Runner scenario.  This means running Maze Runner as a Gem on macOS - until now we have always run it as a Docker container (in CI at least).

The export of the IPA, which is needed for the later e2e test, is purposefully performed outside of the MazeRunner process.  It initially landed inside it and the xcodebuild command failed for reasons I didn't understand.  Although the build command needs to run inside MazeRunner (to capture the source map upload), there is no need for the export to, so I chose to simply move it rather than spend time trying to fix a problem I had no clues on.

## Changeset

Excluding features files in build steps was getting quite messy, so I've filtered them into separate features and renamed a couple for greater clarity. 

## Testing

Covered by CI and also easy to run locally - instructions for doing that will be added shortly (the work is ticketed).


test/react-native-cli/scripts/init-and-build-test.sh
```diff
@@ -0,0 +1,18 @@
...
+REACT_NATIVE_VERSION=$VERSION bundle exec maze-runner features/build-app-tests/build-ios-app.feature
+
+# Export the IPA separately from MazeRunner (running it inside failed for an unknown reason)
+cd features/fixtures
+xcrun --log xcodebuild -exportArchive -archivePath "${VERSION}/${VERSION}.xcarchive" -exportPath output -verbose -exportOptionsPlist exportOptions.plist
+
+# Clear the archive away
+rm -rf ${VERSION}/${VERSION}.xcarchive
+
+# Copy file to build directory0
```
twometresteve, line 16:
```suggestion
# Copy file to build directory
```

bengourley (approved):
Looks good!

test/react-native-cli/scripts/init-and-build-test.sh
twometresteve: I'll correct this on the next PR to avoid suffering another CI run.

PR #1219, opened 2021-01-04
Expo SDK 40 support

imjoehaines:
- Bump versions of Expo packages we rely on to match the ones required by Expo 40
- Add another version check to expo-cli — assuming we'll release the RN CLI first, otherwise this will need to change slightly


bengourley (approved):
Changeset looks good, but let's hold until the promise rejection issue is resolved.

PR #1222, opened 2021-01-05
chore: configure jest coverage

djskinner:
## Goal

To monitor code coverage and ensure high / increasing code quality.

## Design

This is a fairly standard approach to generating code coverage reports with jest. jest supports many Istanbul reporters which are compatible with a range of viewers and CI setups.

This setup also includes files that do not have corresponding test files so provides a more accurate picture of the actual coverage of source.

This will also support (with additional work) danger bot to add code coverage comments on PRs.

## Changeset

- Jest configuration to get code coverage working
- Some tests were pointing to the package root, meaning the `package.json` `main` property was being used to resolve the import. This lead to tests being run on the `dist/` output. Whilst testing on the final output has its merits it is more common to test against the source. If we run the tests against the `dist/` output the coverage for the source file will be reported as zero.
- As a consequence of the above co-located TypeScript declaration files need to be created whereas previously they were able to be resolved using the `package.json` `types` field.

## Testing

automated build and tests passing


packages/browser/test/index.test.ts
```diff
@@ -1,5 +1,4 @@
-import BugsnagBrowserStatic from '..'
-import { Breadcrumb, Session } from '../types/bugsnag'
+import BugsnagBrowserStatic, { Breadcrumb, Session } from '../src/notifier'
```
djskinner, line 1: Change to test against src rather than dist or coverage doesn't report correctly. As a consequence we lose the types, hence the colocated definitions file above that re-exports the types.

packages/node/__mocks__/https.js
```diff
@@ -1,22 +0,0 @@
-/* global jest */
```
djskinner, line 1: This appears to be unused

jest.config.js
```diff
@@ -27,11 +41,12 @@ module.exports = {
       'plugin-window-onerror',
       'plugin-strip-query-string',
       'plugin-interaction-breadcrumbs',
+      'plugin-inline-script-content',
```
djskinner, line 44: This oversight was picked up by the coverage report

bengourley (approved):

PR #1224, opened 2021-01-06
Allow OnErrorCallbacks to return Promise<boolean>

imjoehaines:
## Goal

Currently an async `OnErrorCallback` can only return `Promise<void>` in the Typescript, which means they can't be used to prevent events from sending without causing a compile error. Returning/resolving a boolean value does work, so we can safely allow `Promise<boolean>` as a return type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Fixed
+
+- (core): the `OnErrorCallback` type definition now allows `Promise<boolean>` as a return type. [#1224](https://github.com/bugsnag/bugsnag-js/pull/1224)
+
 ## v7.5.6 (2021-01-11)
 
 ### Changed
@@ -21,7 +27,6 @@
   - Add persistenceDirectory config option for controlling event/session storage [bugsnag-android#998](https://github.com/bugsnag/bugsnag-android/pull/998)
   - Add configuration option to control maximum number of persisted events/sessions [bugsnag-android#980](https://github.com/bugsnag/bugsnag-android/pull/980)
   - Increase kotlin dependency version to 1.3.72 [bugsnag-android#1050](https://github.com/bugsnag/bugsnag-android/pull/1050)
-
 
 ## v7.5.5 (2020-12-14)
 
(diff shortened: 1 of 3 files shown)


bengourley (changes requested):
Changes look good, but since we're in this area I checked the node-style callback for correctness too and found it was also incorrect.

It's hard to get the typing strict enough that it doesn't let people write invalid sync/async without preventing people from writing valid node-style callbacks. I had a quick go and got close, but the `never` in this example makes it unusable for non-trivial node-style callbacks…

image

imjoehaines:
From some more testing, it looks like we can't make it perfectly strict because of how `void` works; I've pushed a fix for node style callbacks and split out each type so it's easier to see the three valid signatures — because of `void` you can still do 'invalid' things like returning an array, but we ignore anything that's not `false` so this shouldn't actually be a problem

bengourley (commented):
image

Node-style callback signature is still not good. Is this the best we can do given the limitations?

imjoehaines:
> Node-style callback signature is still not good. Is this the best we can do given the limitations?

The alternatives are to either make it a union type, which makes `event` `any`:

```ts
type OnErrorCallback
  = ((event: Event) => void | boolean)
  | ((event: Event) => Promise<void | boolean>)
  | ((event: Event, cb: (err: null | Error, shouldSend?: boolean) => void) => void)

notify(event => event.stuff) // error TS7006: Parameter 'event' implicitly has an 'any' type.
```

or mark `cb` as always present, which isn't true but does seem to fix the other problems:

```ts
type OnErrorCallback = (event: Event, cb: (err: null | Error, shouldSend?: boolean) => void) => void | boolean | Promise<void | boolean>

notify(event => event.stuff) // works
notify((event, cb) => { cb(null) }) // works

notify(() => 1) // error TS2322: Type 'number' is not assignable to type 'boolean | void | Promise<boolean | void>'.
```

bengourley:
> mark cb as always present, which isn't true but does seem to fix the other problems:

I'd go with this. Marking it as always present doesn't mean people have to define it.

bengourley (approved):

PR #1225, opened 2021-01-06
deps(react-native): Bump bugsnag-cocoa to v6.5.0

nickdowell:
* Updates bugsnag-cocoa to v6.5.0
* Updates BugsnagReactNative to use new private header files instead of redeclaring @interfaces - this has revealed some APIs that are not yet exposed in the private headers, which will be addressed in the future.


bengourley (changes requested):
Looks good, just needs a changelog entry.

bengourley (changes requested):
Looks like the changelog entry went under 7.5.5 which is a release that went out in December?

Can you add a new slot at the top of the file instead?

bengourley (approved):

PR #1226, opened 2021-01-06
chore: report code coverage with danger

djskinner:
## Goal

Configure danger to report code coverage on PRs

## Design

Standard off the shelf plugin

## Changeset

Configure danger to report code coverage on PRs


djskinner:
See https://github.com/bugsnag/bugsnag-js/pull/1249#issuecomment-762733100 for what the PR comment looks like when files are changed which increase, decrease and leave coverage unchanged.

bengourley (changes requested):
Looks great. Really good to have this info on PRs.

Can you update the name of the workflow action file `.github/workflows/pr-size.yml` and the "PR size bot" names inside it to make it more appropriate for the expanded use case?

bengourley (approved):

PR #1227, opened 2021-01-08
RN CLI: Fix flakes in Maze Runner tests

imjoehaines:
## Goal

The RN CLI tests sometimes seem to get stuck while checking the last command exit code. It seems that it was trying to run the command to check the exit code while the previous command was running. Usually this works but sometimes it seems to never run the exit code command, so I've added waits to ensure the prompt is ready before trying to check the exit code

We also now strip ANSI escape codes from output which seems to help, though I'm not entirely sure why

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gem 'cocoapods'
 
 # A reference to Maze Runner is only needed for running tests locally and if committed it must be
 # portable for CI, e.g. a specific release.  However, leaving it commented out would mean quicker CI.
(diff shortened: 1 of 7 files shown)


bengourley (approved):
Seems good so far

imjoehaines:
@bengourley annoyingly this just flaked in a way I've not seen before — this should still be a lot more reliable than it was so I think it's worth merging and raising a new ticket if the flakes are persistent

bengourley (approved):

PR #1230, opened 2021-01-08
RN CLI: Add a warning when the gradle file is applied

imjoehaines:
The BAGP automatically applies this repository, so this file is no longer necessary


fractalwrench (approved):
LGTM

PR #1240, opened 2021-01-13
Add tvOS to the @bugsnag/react-native podspec

marnusw:
## Goal

When building a React Native project for a tvOS target the error `Bugsnag/Bugsnag.h file not found` occurs.

## Design

The library seems to work fine with tvOS as it is, the podspec just needed to reflect this. It's possible version 9 can also be supported, but most React Native libraries seem to be targeting version 10.

## Changeset

Added the tvOS platform to the podspec.

## Testing

Successful build for a tvOS target and confirmed that handled and unhandled errors are submitted from JavaScript. I did not test native tvOS errors since I'm not a native developer.

### DIFF
--- a/packages/react-native/BugsnagReactNative.podspec
+++ b/packages/react-native/BugsnagReactNative.podspec
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
   s.homepage     = "https://github.com/bugsnag/"
   s.license      = "MIT"
   s.author       = { "Bugsnag" => "platforms@bugsnag.com" }
-  s.platform     = :ios, "7.0"
+  s.platforms    = { :ios => "7.0", :tvos => "10.0" }
   s.source       = { :git => "https://github.com/bugsnag/bugsnag-js.git", :tag => "v#{s.version}" }
   s.source_files = "ios/BugsnagReactNative/**/*.{h,m}",
                    "ios/vendor/bugsnag-cocoa/**/*.{h,mm,m,cpp,c}",


marnusw:
Thanks for the reply @mattdyoung.

marnusw:
Just a though: This change may be a "better than nothing" situation which other developers may also prefer. I understand there may be issues with native error reporting, but the JS reporting works 100% and that is the primary need in React Native apps. So perhaps merging this change with a disclaimer in the readme on exactly where tvOS support stands should be considered? If you disagree we can close this PR, I just thought it worth a mention.

xander-jones:
Hey @marnusw, as Matt mentioned above, whilst the codebase change is trivial, a huge part of modifying notifier SDKs is ensuring it works for all of our supported React Native versions, on all devices, with different option sets like Hermes and CodePush, and managing future versions too, alongside documentation and support. For this reason, I'm afraid we're not going to be merging the PR. However, this is not to say (for future travellers, too!) that you can't fork and modify the notifier, it's open-source after all! :), it's just something that we're not ready to officially support just yet. It is on our roadmap – but there are a few competing priorities.

marnusw:
For anyone else finding this, I figured adding this to a postinstall script in our project would be easier than maintaining a separate fork of this project:
```
sed -i.bak 's|s.platform *= :ios, "[0-9]*.[0-9]*"|s.platforms    = { :ios => "7.0", :tvos => "10.0" }|g' ./node_modules/@bugsnag/react-native/BugsnagReactNative.podspec
```

PR #1242, opened 2021-01-14
[PLAT-5726] fix(react-native): Ensure project root is set correctly in xcode build phase

bengourley:
## Goal

Ensure source map uploads that were automated via the RN CLI injecting the build phase have their project root stripped correctly.

## Design

The cwd of the script was `<project_root>/ios`. Since none of the JS on a project lives inside the `ios` directory, to correctly have the source map tool strip the project root it needs to be set to `<project_root>`.

## Changeset

Updated the bugsnag-react-native-xcode.sh to pass the desired `--project-root` option to @bugsnag/source-maps.

## Testing

Tested manually on RN 0.61, 0.62 and 0.63.

### DIFF
--- a/packages/react-native/bugsnag-react-native-xcode.sh
+++ b/packages/react-native/bugsnag-react-native-xcode.sh
@@ -30,6 +30,9 @@ if [ ! -f "$MAP_FILE" ]; then
   exit 1
 fi
 
+# This script gets executed in the <project_root>/ios directory
+PROJECT_ROOT=$(echo $PWD | sed -e 's/\ios$//')
+
 ARGS=(
     "--api-key" "$API_KEY"
     "--app-bundle-version" "$BUNDLE_VERSION"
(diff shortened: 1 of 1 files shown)


nickdowell (approved):
LGTM 👍 

You could alternatively take advantage of Bash's [shell parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) and do

```
PROJECT_ROOT=${PWD/\/ios}
```

Annoyingly the [Xcode documentation](https://help.apple.com/xcode/mac/11.4/index.html?localePath=en.lproj#/devc8c930575) does not say what `$PWD` will be when running a script phase, but the current behaviour is extremely unlikely to change.

bengourley:
@nickdowell updated this to use shell expansion as suggested. From checking the docs, I used `%` as otherwise it would replace potential occurrences mid-path, e.g.:

```diff
- /Users/bengourley/Development/iosprojects/ReactNativeSixtyOne/ios
+ /Users/bengourley/Developmentprojects/ReactNativeSixtyOne
```

nickdowell:
> @nickdowell updated this to use shell expansion as suggested. From checking the docs, I used `%` as otherwise it would replace potential occurrences mid-path, e.g.:
> 
> ```diff
> - /Users/bengourley/Development/iosprojects/ReactNativeSixtyOne/ios
> + /Users/bengourley/Developmentprojects/ReactNativeSixtyOne
> ```

Well spotted!

Issue #1143, opened 2020-11-25
[React-Native] React-Navigation Plugin will crash app


KDederichs:
### Describe the bug
When trying to execute react-native in remote debug mode the app will crash, since synchronous calls are apparently no-op.
Too bad `Bugsnag.getPlugin` seems to fall in that category as well...

### Steps to reproduce
1. Use bugsnag react navigation plugin
2. remote debug

### Environment
* Bugsnag version:
    "@bugsnag/plugin-react-navigation": "^7.5.0",
    "@bugsnag/react-native": "^7.3.5",
* Browser framework version (if any):
    * React: react-native 0.63.3

### Example code snippet 

At the application root level:

```
Bugsnag.start({
  plugins: [new BugsnagPluginReactNavigation()],
});
const { createNavigationContainer } = Bugsnag.getPlugin('reactNavigation');
```

<details><summary>Error messages :</summary>

`_Bugsnag$getPlugin is undefined`
</details>

mattdyoung:
Hi @KDederichs 

This is the same underlying issue as described on https://github.com/bugsnag/bugsnag-js/issues/1112 and we're considering how best we can fix this to cater for when a debugger is attached.

Until a fix is available you'd need to use a workaround similar to the one described on that issue:
https://github.com/bugsnag/bugsnag-js/issues/1112#issuecomment-722970857

yousif-bugsnag:
Fixed in [v7.6.1](https://github.com/bugsnag/bugsnag-js/releases/tag/v7.6.1)

Issue #1112, opened 2020-10-23
TypeError: Cannot read property 'createErrorBoundary' of undefined


duzliang:
Following the [Bugsnag Docs](https://docs.bugsnag.com/platforms/react-native/react-native/#capturing-react-render-errors) to config the project and run got this error from:
`const ErrorBoundary = Bugsnag.getPlugin('react').createErrorBoundary(React);`

- react 16.11.0
- react-native 0.62.2
- @bugsnag/react-native 7.5.0

johnkiely1:
Hi @duzliang ,

Could you confirm what error you are seeing? Also could you share you Bugsnag configuration code for both your JS and native layers. Feel free to email support@bugsnag.com if you don't want to share this information publicly. 

Thanks

duzliang:
@johnkiely1 Thank you for viewing this issue:

Both the Android and iOS configuration are set up step by step by the [Docs](https://docs.bugsnag.com/platforms/react-native/react-native/#basic-configuration)

Below is a simple Javascript code snippets:

1. App.js
```
import Bugsnag from '@bugsnag/react-native';

import ErrorView from './containers/ErrorView';

const ErrorBoundary = Bugsnag.getPlugin('react').createErrorBoundary(React);

export default class AppRepair extends React.Component {
  constructor(props) {
    Bugsnag.start();
  }

 render() {
    return (
      <ErrorBoundary FallbackComponent={ErrorView}>
        <Provider store={store}>
          <Routes handleExit={this.handleExit} />
        </Provider>
      </ErrorBoundary>
    );
  }
}
```

2. index.js
```
import { AppRegistry } from 'react-native';
import App from './src/App.js';
import { name as appName } from './app.json';

AppRegistry.registerComponent(appName, () => App);

```
when running the App, get the error:
```
TypeError: Cannot read property 'createErrorBoundary' of undefined 
```

kennethjiang:
I'm having the same problem here. 

* react 16.13.1
* react-native 0.63.2
* @bugsnag/react-native 7.5.1

mattdyoung:
@duzliang 
The issue is that you're calling `Bugsnag.getPlugin('react')` before you call `Bugsnag.start()`.

`Bugsnag.start()` must be called before any other method so you'll need to move it out of the constructor of your component.

duzliang:
Move it out of the constructor also get the Error while open the `Debug` mode of `React Native Debug Menu`

```
[...]

PR #1250, opened 2021-01-19
[PLAT-5357] fix(react-native): Ensure plugin usage does not crash debugger

bengourley:
## Goal

When run in the debugger, the React Native notifier starts asynchronously due to synchronous calls to native not being available in the debugging environment.

The existing workaround involved creating a no-op stubbed client interface, which covered most use cases, but not plugins where the return value of `Bugsnag.getPlugin('<name>')` was required, resulting in the following bugs: #1143 #1112.

## Design

The new solution is to create an actual `Client` instance so that plugins can be used in the short delay before the notifier starts properly.

## Changeset

- Updated the debugging-env only stub client to use an actual client object
- Added messaging to the debugging console warning

## Testing

The changes were manually tested in the remote debugger on an app using React Navigation.

### DIFF
--- a/packages/react-native/src/notifier.js
+++ b/packages/react-native/src/notifier.js
@@ -4,6 +4,7 @@ const NativeClient = NativeModules.BugsnagReactNative
 const REMOTE_DEBUGGING_WARNING = `Bugsnag cannot initialize synchronously when running in the remote debugger.
 
 Error reporting is still supported, but synchronous calls after Bugsnag.start() will no-op. This means Bugsnag.leaveBreadcrumb(), Bugsnag.setUser() and all other methods will only begin to work after a short delay.
+Plugins are also affected. Synchronous calls to Bugsnag.getPlugin() can be used as normal, but plugins may not report errors and set contextual data correctly in the debugger.
 
 This only affects the remote debugger. Execution of JS in the normal way (on the device) is not affected.`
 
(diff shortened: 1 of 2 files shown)


packages/react-native/src/notifier.js
```diff
@@ -97,32 +98,45 @@ const Bugsnag = {
...
+    const stubSchema = { ...schema }
+    delete stubSchema.apiKey
+    const stubClient = new Client({
+      ...opts,
+      autoTrackSessions: false,
+      autoDetectErrors: false,
+      enabledBreadcrumbTypes: []
+    }, stubSchema, internalPlugins, { name, version, url })
+
+    CLIENT_METHODS.map((m) => {
```
imjoehaines, line 120: This could be a forEach as we're not using the returned array

```suggestion
    CLIENT_METHODS.forEach((m) => {
```

packages/react-native/src/notifier.js
```diff
@@ -97,32 +98,45 @@ const Bugsnag = {
...
+      if (/^_/.test(m)) return
+      stubClient[m] = new Proxy(stubClient[m], {
+        apply: (target, thisArg, args) => {
+          if (!initialised) {
+            console.log(`Synchronous call to Bugsnag.${m}()`)
```
imjoehaines, line 125: I'm not sure I'd understand what this message is trying to say if I didn't know the context; how about something closer to the old message?

```suggestion
            console.log(`Synchronous call to Bugsnag.${m}() may not work because remote debugging is enabled and Bugsnag has not yet initialized`)
```

packages/react-native/src/notifier.js
```diff
@@ -97,32 +98,45 @@ const Bugsnag = {
...
-            if (!initialised) return Reflect.apply(target, thisArg, argumentsList)
-            return Reflect.apply(Bugsnag._client[m], Bugsnag._client, argumentsList)
+    }
+
+    console.warn(REMOTE_DEBUGGING_WARNING)
+
+    let initialised = false
+
+    const stubSchema = { ...schema }
+    delete stubSchema.apiKey
```
imjoehaines, line 112: It's not obvious to me why we need to remove the apiKey here; this could use a comment

packages/react-native/src/notifier.js
```diff
@@ -97,32 +98,45 @@ const Bugsnag = {
...
+          if (!initialised) {
+            console.log(`Synchronous call to Bugsnag.${m}()`)
           }
-        })
-      })
-      Bugsnag._client = stubClient
-      createClientAsync(opts).then(client => {
-        initialised = true
-        Bugsnag._client = client
+          return Reflect.apply(target, thisArg, args)
```
imjoehaines, line 127: Should this still be applying to `Bugsnag._client` if it has initialised?

packages/react-native/src/notifier.js
bengourley, line 125: I changed this to a log rather than a warn, so that it doesn't pop up in the app's yellowbox UI.

packages/react-native/src/notifier.js
bengourley: Yeah it has to carry being functional because plugins created and obtained synchronously still have a reference to that stubbed client. If we stopped `apply()`ing and no-op'ing, it could fail due to a plugin expecting it to return something.

packages/react-native/src/notifier.js
bengourley: There was another instance of `.map` being used in this way, so I changed both.

packages/react-native/src/notifier.js
bengourley: Comment added.

packages/react-native/src/notifier.js
bengourley: I think it's pretty clear in context, because you get it right after the big block of warning text:

image

imjoehaines: Ah yep, that looks good

imjoehaines (approved):

PR #1254, opened 2021-01-21
test(react-native-cli): Source map upload for RN now defaults to overwrite=true

bengourley:
`@bugsnag/source-maps` v2 now defaults to `overwrite=true` for RN uploads.

The RN CLI tests in this repo grab the latest source maps CLI and have an assertion against the overwrite value in the upload, so tests are incorrect and failing from this point on.

This updates the assertion in the tests and gets CI passing again.


twometresteve (approved):
LGTM

twometresteve:
About to force merge as the build succeeded and the status checks just got confused.  I branched off of this and pushed immediately before cancelling elements of the build I was not interested in - this caused a "fail" status to be sent from Buildkite for the same commit id.

PR #1256, opened 2021-01-21
Add mixed stacktrace support for Android

bengourley:
This PR incorporates updates to the Android notifier which are not yet published. The current head of the [`integration/capture-ndk-stack-traces-for-anrs`](https://github.com/bugsnag/bugsnag-android/pull/1003) branch has been vendored in, but this includes other unrelated changes. Before release this will need to be updated to a released version of the Android notifier.

https://github.com/bugsnag/bugsnag-js/pull/1256/commits/3adc2a12af04ec21cae380c96c5d78bd4c8fb80b is unrelated to the changeset but was required to get the notifier built for running the tests locally.

This just needed a small tweak to the Kotlin code in this codebase, and the tests reworked from the previous implementation.


test/react-native/features/native-stack.feature
```diff
@@ -1,66 +1,66 @@
...
+  And the payload field "events.0.exceptions.0.stacktrace.2.method" equals one of:
+    | com.reactnative.module.BugsnagModule.runScenario |
+    | runScenario |
 
-#   # PLAT-5117 addresses float serialization
-#   And the payload field "events.0.exceptions.1.stacktrace.0.lineNumber" equals 1
-#   And the payload field "events.0.exceptions.1.stacktrace.1.lineNumber" equals 1
-#   And the payload field "events.0.exceptions.1.stacktrace.2.lineNumber" equals 2
+  # the javascript part follows
+  And the payload field "events.0.exceptions.0.stacktrace.10.columnNumber" is not null
```
fractalwrench, line 28: This seems to fail on RN 0.63 as there are 13 native stackframes. Changing the frame index to 14 might work in the short term, or it might be possible to create some step that looks at the last frame in a stacktrace?

fractalwrench (approved):
LGTM barring one minor tweak to the E2E test assertions

test/react-native/features/native-stack.feature
bengourley: Ahh that's frustrating! The last two stackframes don't have much info so I think I'll bump this to the 14th item as you suggest.

PR #1278, opened 2021-02-02
Tests: Update react-native e2e tests to use maze-runner v4

Cawllec:
## Goal

Allows the maze-runner tests to use new v4 features, including endpoint separation and other improvements in most recent release.

Running the tests locally haven't changed, and the documentation remains the same.


test/react-native/features/override_unhandled.feature
```diff
@@ -2,9 +2,10 @@ Feature: Overriding unhandled state
 
 Scenario: Non-fatal error overridden to unhandled
     When I run "HandledOverrideJsErrorScenario"
-    Then I wait to receive 2 requests
-    And I discard the oldest request
-    And the exception "message" equals "HandledOverrideJsErrorScenario"
+    Then I wait to receive an error
+    And I wait to receive a session
```
twometresteve, line 6: Is it important to these two scenarios that a session is received - it was simply discarded before, presumably because it was just the first out and needed to be got past.  I suspect we can simply remove the steps to receive a Session here.

test/react-native/features/session.feature
```diff
@@ -25,142 +25,143 @@ Scenario: Automatic sessions disabled
...
+  And I discard the oldest error
 
-  And the payload has a valid sessions array
-  And the payload field "sessions.0.id" does not equal the stored value "initial_session_id"
-  And the payload field "sessions.0.id" is stored as the value "second_session_id"
-
-  And I discard the oldest request
+  And the session payload has a valid sessions array
+  And the session payload field "sessions.0.id" does not equal the stored value "initial_session_id"
+  And the session payload field "sessions.0.id" is stored as the value "second_session_id"
```
twometresteve, line 72: I think it would be cleaner to move these expectations on the second session to immediately follow the first. Then we're checking all sessions, then checking all errors - without saying anything about the order that we might expect between them.

test/react-native/features/session.feature
```diff
@@ -25,142 +25,143 @@ Scenario: Automatic sessions disabled
...
   And the exception "errorClass" equals "Error"
   And the exception "message" equals "SessionJsControlledManualJsScenarioC"
   And the event "unhandled" is false
   And the event "session" is not null
-  And the payload field "events.0.session.id" equals the stored value "initial_session_id"
+  And the error payload field "events.0.session.id" equals the stored value "initial_session_id"
   And the event "session.events.handled" equals 2
 
-  And I discard the oldest request
+  And I discard the oldest error
```
twometresteve, line 68: This can be removed.

test/react-native/features/session.feature
```diff
@@ -25,142 +25,143 @@ Scenario: Automatic sessions disabled
...
-
-  And the payload has a valid sessions array
-  And the payload field "sessions.0.id" does not equal the stored value "initial_session_id"
-  And the payload field "sessions.0.id" is stored as the value "second_session_id"
+  And I discard the oldest error
 
-  And I discard the oldest request
+  And the session payload has a valid sessions array
+  And the session payload field "sessions.0.id" does not equal the stored value "initial_session_id"
+  And the session payload field "sessions.0.id" is stored as the value "second_session_id"
```
twometresteve, line 128: Same here, move the session checks up and remove line 124.

test/react-native/features/session.feature
```diff
@@ -169,30 +170,28 @@ Scenario: Manual Native sessions (JS Controls)
...
```
twometresteve, line 188: Same

twometresteve (approved):
Looks good - just made a few suggestions for little improvements.

bengourley (approved):

test/react-native/features/override_unhandled.feature
Cawllec: I think it's reasonable to expect it, as we're verifying the session elements in the error payload and if that request _isn't_ received it's indicative that those elements won't be present because the session wasn't delivered.

test/react-native/features/session.feature
Cawllec: I did that initially, but thought it would be more readable to follow the order in which sessions/errors were delivered:
- First session received
- Error matches first session
- Error matches no session (session is paused)
- Error matches first session (session is resumed)
- Second session received
- Error matches second session

twometresteve: Yeah, that's fair - I take the point that there is a logical ordering there and it's not about the order in which requests are received as such.

test/react-native/features/session.feature
twometresteve: Self-resolved based on previous comment.

test/react-native/features/session.feature
twometresteve: Self-resolved based on previous comment.

test/react-native/features/session.feature
Cawllec: I've added some comments to make this ordering clearer

PR #1279, opened 2021-02-02
Use separate branch for RN promise rejection stacktraces

fractalwrench:
## Goal

Updates the RN mixed stacktrace integration branch to use an Android dep generated from this branch: https://github.com/bugsnag/bugsnag-android/pull/1103


packages/react-native/update-android.sh
```diff
@@ -14,7 +14,7 @@ echo "Clearing local maven repo"
 rm -rf $MAVEN_REPO_DIR
 
 echo "Building local copy of bugsnag-android $AMENDED_VERSION"
-./$ANDROID_REPO_DIR/gradlew -p $ANDROID_REPO_DIR assembleRelease publishToMavenLocal -PVERSION_NAME=$AMENDED_VERSION
+./$ANDROID_REPO_DIR/gradlew -p $ANDROID_REPO_DIR clean assembleRelease publishToMavenLocal -PVERSION_NAME=$AMENDED_VERSION
```
bengourley, line 17: Does this fix the vendoring issue you were having?

bengourley (approved):

packages/react-native/update-android.sh
fractalwrench: Yep - this should prevent the vendoring issue by ensuring that the build cache isn't used.

PR #1281, opened 2021-02-03
deps(react-native): Update bugsnag-cocoa to v6.6.1

nickdowell:


nickdowell:
Branch name containing `.` appears to be causing the build script to fail 🤨

PR #1283, opened 2021-02-04
[PLAT-5850] Add end-to-end tests for Vue+TypeScript

bengourley:
Add end to end test coverage for TypeScript + Vue 2-3. The types are written to support both versions so it's important we test against both.

### DIFF
--- a/test/browser/features/fixtures/plugin_vue/typescript_vue2/package.json
+++ b/test/browser/features/fixtures/plugin_vue/typescript_vue2/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "bugsnag-js-fixtures-vue3-typescript",
+  "private": true,
+  "scripts": {
+    "build": "NODE_ENV=production browserify src/index.ts -p [ tsify ] -g envify -o dist/bundle.js"
+  },
+  "dependencies": {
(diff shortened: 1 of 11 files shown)


test/browser/features/fixtures/plugin_vue/typescript_vue2/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "bugsnag-js-fixtures-vue3-typescript",
```
imjoehaines, line 2:
```suggestion
  "name": "bugsnag-js-fixtures-vue2-typescript",
```

imjoehaines (approved):
Looks like Firefox 30 doesn't like the Vue3 fixture even with the Proxy check, otherwise LGTM

bengourley:
> Looks like Firefox 30 doesn't like the Vue3 fixture even with the Proxy check, otherwise LGTM

Vue uses the `class` keyword a lot in its bundle, which is not supported in Firefox 30. Rather than transpile to ES3 (which I think setting `target: "es3"` in the `tsconfig.json` would have done) I opted to skip test if the use of `class` throws.

PR #1286, opened 2021-02-09
[PLAT-5651] Update react-native-cli tests to use Maze-runner v4

Cawllec:
## Goal

Separate endpoints introduced in Maze-runner v4 should reduce the number of flakes involving payload ordering, and add several other improvements to how the end-to-end tests run.

At the moment there are several steps that are ready to be moved into the main maze-runner repo, which will be done in a follow-up PR.


test/react-native-cli/features/steps/steps.rb
```diff
@@ -509,22 +509,8 @@ def parse_xml_file(path)
...
-  assert_not_nil(read_key_path(request_body, 'metadata.java_version'))
-  assert_not_nil(read_key_path(request_body, 'metadata.gradle_version'))
-  assert_not_nil(read_key_path(request_body, 'metadata.git_version'))
-end
-
 Then('the Content-Type header is valid multipart form-data') do
   expected = /^multipart\/form-data; boundary=--------------------------\d+$/
-  actual = Server.current_request[:request]["content-type"]
-
+  actual = Maze::Server.list_for('build').current[:request]['content-type']
```
imjoehaines, line 514: I think this can be simplified:

```suggestion
  actual = Maze::Server.builds.current[:request]['content-type']
```

imjoehaines (approved):

TESTING.md
```diff
@@ -248,3 +248,103 @@ particular, these commands need the `BrowserStackLocal` binary (available
...
+Before running these tests the previous tests, `Build Tests` must be run for the test fixtures to be present.
+
+1. Change directory into `test/react-native-cli`
+1. Check the contents of `Gemfile` to select the version of `maze-runner` to use
+1. Install maze-runner with `bundle install`
+
+##### Running
+
+Ensure that the following environment variables are set:
+- `BROWSER_STACK_USERNAME` - Your BrowserStack App Automate Username
```
twometresteve, line 322: The use of these variables is now deprecated.  If you update these line to say `MAZE_DEVICE_FARM_USERNAME` and `MAZE_DEVICE_FARM_ACCESS_KEY` then you can remove the `--username` and `--access-key` CL options in the examples below.

Also, if you add `MAZE_BS_LOCAL` ("location of the BrowserStack local testing binary") to the list then you can remove use of the `--bs-local` option.

Gemfile
```diff
@@ -4,7 +4,7 @@ gem 'cocoapods'
 
 # A reference to Maze Runner is only needed for running tests locally and if committed it must be
 # portable for CI, e.g. a specific release.  However, leaving it commented out would mean quicker CI.
-gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v3.7.0'
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v4.6.0'
```
twometresteve, line 7: I don't think we want to change this yet as it will break local running of the Expo and Node tests, which have not been moved to MR v4 yet and do not have their own Gemfile(.lock)s.  

For now I think I'd simply revert this file and `Gemfile.lock`, but once the Expo and Node scenarios have been updated to v4 we can lose them completely.

twometresteve (approved):

PR #1140, opened 2020-11-23
[PLAT-4935] Add automate-symbolication command (iOS)

bengourley:
Add a command to automate source map upload when a React Native build happens.

Unit tests included for relevant parts. Interactive CLI aspects will be test via maze runner at a later point.

To test this out, use the following steps:

- `git fetch && git checkout feature/rn-cli-automate-symbolication-cmd-ios`
- `npm i`
- `npm run bootstrap`
- `npm run build`
- `npx lerna publish v99.99.99-canary.`git rev-parse HEAD` --no-push --exact --no-git-tag-version --registry http://localhost:4873/` (ensure verdaccio is running, per the react-native/CONTRIBUTING.md guide)
- `cd packages/react-native-cli`
- `npm link` (this symlinks the local cli package to your global npm modules)

On a React Native project:

- check that the command is linked up correctly `bugsnag-react-native-cli --help` 
- install the notifier you published to your local npm `npm i @bugsnag/react-native@v99.99.99-canary.{gitsha} --registry http://localhost:4873/` or `yarn add @bugsnag/react-native@v99.99.99-canary.{gitsha} --registry http://localhost:4873/`
- run `bugsnag-react-native-cli automate-symbolication`

PR #1289, opened 2021-02-11
[PLAT-5950] Store source maps in $TMPDIR during Xcode build phases

nickdowell:
## Goal

Fix building projects whose build directory path contains whitespace.

The underlying issue is that react native's script cannot accept any paths containing spaces in `$EXTRA_PACKAGER_ARGS`

## Design

A new location has been designated for the source map: `$TMPDIR/$UID-main.jsbundle.map`

This path is guaranteed to not contain any whitespace, which will fix compatibility with `react-native-xcode.sh`

Use `md5` to generate a unique `$UID` from the build path; unique for that project, build configuration and target.

The script will fall back to the previous location to remain compatible with the old source map location, for Xcode projects that have not been updated.

## Testing

Manually tested using sample RN projects.

Tested by following instructions from https://github.com/bugsnag/bugsnag-js/pull/1140 against a freshly created react native app.


packages/react-native/bugsnag-react-native-xcode.sh
```diff
@@ -23,7 +23,14 @@ if [ ! -f "$BUNDLE_FILE" ]; then
...
 
-MAP_FILE="$BUNDLE_FILE.map"
+# Generate a stable unique path that does not contain spaces, for compatibility with react-native-xcode.sh
+UID=$(md5 -qs "$CONFIGURATION_BUILD_DIR")
+MAP_FILE="$TMPDIR/$UID-main.jsbundle.map"
+if [ ! -f "$MAP_FILE" ]; then
+  echo "Warning: $MAP_FILE could not be found, falling back to $BUNDLE_FILE.map"
+  # Older versions of the Xcode build phase used this location
+  MAP_FILE="$BUNDLE_FILE.map"
+fi
```
tomlongridge, line 33: Can we instead take this as a parameter so that the same value can be passed in as was supplied to `react-native-xcode.sh`? That way, if the user is setting this up manually they can set the directory to whatever they like.

nickdowell: The build phase now looks like

Screenshot 2021-02-16 at 11 30 06

bengourley (approved):
Looks good. I tested with a project set up with the old version of the cli, using the new version of the notifier and it falls back nicely.

Just needs a changelog entry.

PR #1290, opened 2021-02-11
[PLAT-5980] Ensure inline <script> detection only includes script content when relevant

bengourley:
## Goal

In certain edge cases it is possible for the inline script detection plugin to add `metadata.script.content` when the source of the error is not an inline script.

## Design

~The first problem is that errors without stacktraces were able to get past this line which exits early if the topmost stackframe `file` does not match that of the current page:~

https://github.com/bugsnag/bugsnag-js/blob/a53eaccefe343548ecb925d1f3d8b3aac2a8f6c3/packages/plugin-inline-script-content/inline-script-content.js#L66

~It stands to reason that if we have no stacktrace, we don't know whether the error was from an inline script or not, so attaching the inline script content seems like a bad idea.~

The second problem is that when Bugsnag is loaded _after_ the document `readyState=interactive` event has fired – note this is **not** recommended – it remains in a constant state believing that the document is still loading, and therefore running synchronous inline scripts.

https://github.com/bugsnag/bugsnag-js/blob/a53eaccefe343548ecb925d1f3d8b3aac2a8f6c3/packages/plugin-inline-script-content/inline-script-content.js#L14

When an error happens in this instance (but only when there are no stackframes), it grabs the last `<script>` tag on the page, believing that that is was the cause of the error.

The solution is simply to check the `document.readyState` before deciding whether the DOM content has loaded or not.

---

📢 **Update**: in old IE (8, 9 and 10), sometimes errors that we can get the inline script for don't have a stacktrace. We do this by determining if the document is "ready" – if it's not, and we're handling an error, we can assume it came from inline in the page, in the last script that exists.

Fixing the readyState logic above works for the modern browsers, because `getCurrentScript()` will now alway be `null` when an inline script is not running. However, it broke the use case for old IEs where the page is still loading due to [a bug where "interactive" event fires too early](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState#browser_compatibility). A special case has been made for those browsers to wait for the "complete" state, which is a reasonable workaround.

---

## Changeset

- Updated the inline script plugin based on the reasoning above.
- Updated unit tests
- Added some end to end tests for inline script hand

## Testing

Tested manually as well as the aforementioned automated tests.

### DIFF
--- a/packages/plugin-inline-script-content/inline-script-content.js
+++ b/packages/plugin-inline-script-content/inline-script-content.js
@@ -11,7 +11,7 @@ module.exports = (doc = document, win = window) => ({
 
     const originalLocation = win.location.href
     let html = ''
-    let DOMContentLoaded = false
+    let DOMContentLoaded = doc.readyState !== 'loading'
     const getHtml = () => doc.documentElement.outerHTML
 
     // get whatever HTML exists at this point in time
@@ -66,7 +66,7 @@ module.exports = (doc = document, win = window) => ({
       if (!frame) return
 
       // if frame.file exists and is not the original location of the page, this can't be an inline script
-      if ((frame.file && frame.file.replace(/#.*$/, '')) !== originalLocation.replace(/#.*$/, '')) return
+      if (frame && (frame.file && frame.file.replace(/#.*$/, '')) !== originalLocation.replace(/#.*$/, '')) return
 
       // grab the last script known to have run
       const currentScript = getCurrentScript()
--- a/packages/plugin-inline-script-content/inline-script-content.js
+++ b/packages/plugin-inline-script-content/inline-script-content.js
@@ -62,8 +62,11 @@ module.exports = (doc = document, win = window) => ({
 
       const frame = event.errors[0].stacktrace[0]
 
+      // if there are no frames, we can't tell whether this is an inline script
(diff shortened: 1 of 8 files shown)


packages/plugin-inline-script-content/inline-script-content.js
```diff
@@ -62,8 +62,11 @@ module.exports = (doc = document, win = window) => ({
 
       const frame = event.errors[0].stacktrace[0]
 
+      // if there are no frames, we can't tell whether this is an inline script
+      if (!frame) return
+
       // if frame.file exists and is not the original location of the page, this can't be an inline script
-      if (frame && frame.file && frame.file.replace(/#.*$/, '') !== originalLocation.replace(/#.*$/, '')) return
+      if (frame && (frame.file && frame.file.replace(/#.*$/, '')) !== originalLocation.replace(/#.*$/, '')) return
```
imjoehaines, line 69: Does this still need the first frame check now?

```suggestion
      if (frame.file && frame.file.replace(/#.*$/, '') !== originalLocation.replace(/#.*$/, '')) return
```

imjoehaines (approved):

imjoehaines (approved):

PR #1292, opened 2021-02-12
fix(plugin-koa) add request body if present

### LINKED ISSUE #1183: plugin-koa: body not showing up in request tab
Same as https://github.com/bugsnag/bugsnag-js/issues/620 but for plugin-koa

Looks like what was fixed in https://github.com/bugsnag/bugsnag-js/pull/872 still exists in plugin-koa

https://github.com/bugsnag/bugsnag-js/blob/55443da3eeed541adb585cc33d78e738e3d8aef9/packages/plugin-koa/src/koa.js#L24-L28

djskinner:
## Goal

Include the request body when `ctx.request.body` is present, for example, when using the `koa-bodyparser` middleware.

Closes #1183 

## Design

Even if the request body was being included in `request` and `metadata` it wouldn't have been included because those values were being determined when the `requestHandler` middleware was run, which would be before any body parsing middleware is added, as per our integration instructions, which suggest adding the Bugsnag middleware first.

The changes here are somewhat in line with the way the express plugin works, see https://github.com/bugsnag/bugsnag-js/pull/872 with the exception being the body is only added to `event.request`, rather than metadata, which is the preferred approach going forwards in alignment with the notifier spec.

`httpVersion` is also added for the same reason. `params` and `cookies` were not added as part of this change as it was not clear which middleware we should support (whereas `koa-bodyparser` seems to be the de-facto standard for koa body parsing)

## Changeset

- Add `body: ctx.request.body` to `extractRequestInfo`
- Include `body` in `request` in `getRequestAndMetadataFromCtx`
- Include `httpVersion` in `request` in `getRequestAndMetadataFromCtx`. This is an unrelated change but further brings the plugin into compliance with the [spec](https://github.com/bugsnag/platforms-notifier-spec/blob/master/request.md).
- Call `getRequestAndMetadataFromCtx(ctx)` in the onError handler instead of when the middleware is first executed. Add the metadata to the event, not the client as any metadata added to the client during onError is not included in the event.

## Testing

The changes are covered by an E2E test.


test/node/features/koa.feature
```diff
@@ -90,3 +90,16 @@ Scenario: A handled error with ctx.bugsnag.notify()
...
+  And I wait to receive a request
+  Then the request is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
+  And the event "unhandled" is true
+  And the event "severity" equals "error"
+  And the exception "errorClass" equals "Error"
+  And the exception "message" equals "request body"
+  And the exception "type" equals "nodejs"
+  And the "file" of stack frame 0 equals "scenarios/app.js"
+  And the event "request.body.data" equals "in_request_body"
+  And the event "request.httpMethod" equals "POST"
```
imjoehaines, line 105: Could also check the `request.httpVersion` here ([should be 1.1](https://github.com/ruby/ruby/blob/b51e5c2370e1579f1f225ea4cfc9e834b7575944/lib/net/http.rb#L393))

packages/plugin-koa/src/koa.js
```diff
@@ -90,15 +90,19 @@ module.exports = {
 }
 
 const getRequestAndMetadataFromCtx = ctx => {
-  const requestInfo = extractRequestInfo(ctx)
+  // Exclude new mappings from metaData but keep existing ones to preserve backwards compatibility
+  const { body, ...requestInfo } = extractRequestInfo(ctx)
```
imjoehaines, line 94: If we're trying to prevent the new mappings from being in metadata, should this also exclude `httpVersion`?

imjoehaines (approved):

packages/plugin-koa/src/koa.js
imjoehaines: Ah no, that was an existing thing in metadata so we can't remove it

bengourley (approved):

test/node/features/koa.feature
```diff
@@ -90,3 +90,17 @@ Scenario: A handled error with ctx.bugsnag.notify()
   And the event "request.url" equals "http://koa/handled"
   And the event "request.httpMethod" equals "GET"
   And the event "request.clientIp" is not null
+
+Scenario: adding body to request metadata
+  When I POST the data "data=in_request_body" to the URL "http://koa/bodytest"
+  And I wait to receive a request
```
bengourley, line 96:
```suggestion
  And I wait to receive an error request
```

PR #1297, opened 2021-02-16
Add Maze Runner scaffolding for Lambda tests

imjoehaines:
## Goal

This PR adds the scaffolding necessary to be able to run Maze Runner tests against Lambda functions. The fixture and tests are by no means complete, but will be expanded on as further work. This PR is purely to prove that we can build, and run tests against, a Lambda fixture

Each commit is standalone so can be reviewed separately if that's easier

## Testing

As we can't run tests using SAM on Buildkite yet, this has to be run locally for now:

1. Install Maze Runner

    ```sh
    $ bundle install
    ```

2. Run the tests

    ```sh
    $ bundle exec maze-runner
    ```

Note: the tests don't assert against Bugsnag payloads as we don't guarantee delivery in Lambda functions (yet!) so they assert against the Lambda responses instead

### DIFF
--- a/test/aws-lambda/.gitignore
+++ b/test/aws-lambda/.gitignore
@@ -1,0 +1,3 @@
+.aws-sam
+node_modules/
+.tgz
--- a/test/aws-lambda/Gemfile
+++ b/test/aws-lambda/Gemfile
@@ -1,0 +1,6 @@
+source 'https://rubygems.org'
+
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v4.9.1'
+
+# Locally, you can run against Maze Runner branches and uncommitted changes:
+#gem 'bugsnag-maze-runner', path: '../../../maze-runner'
--- a/test/aws-lambda/Gemfile.lock
+++ b/test/aws-lambda/Gemfile.lock
@@ -1,0 +1,85 @@
+GIT
+  remote: https://github.com/bugsnag/maze-runner
+  revision: 8677bcf4c1ab27422084b19e7da8149ede954ded
+  tag: v4.9.1
+  specs:
+    bugsnag-maze-runner (4.9.1)
+      appium_lib (~> 11.2.0)
+      boring (~> 0.1.0)
+      cucumber (~> 3.1.2)
+      cucumber-expressions (~> 6.0.0)
+      curb (~> 0.9.6)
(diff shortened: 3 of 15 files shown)


bengourley (approved):
Looks good.

I have a small concern regarding similar work being done in slightly different ways in different languages. The packing of fixtures is now done in Ruby and JS for these tests and the browser/node end to end tests. Maybe the context is too different that we can't consolidate the work, but I think it's worth a look.

twometresteve (approved):
LGTM.

imjoehaines:
> I have a small concern regarding similar work being done in slightly different ways in different languages. The packing of fixtures is now done in Ruby and JS for these tests and the browser/node end to end tests. Maybe the context is too different that we can't consolidate the work, but I think it's worth a look.

I didn't realise we had packing in JS land too; I thought that was done by the dockerfiles and so had to be done differently for these tests as they can't run in docker. It's still a bit different as this is run by MR automatically rather than being a manual script

I think there's still an advantage to this being in Ruby; the buildkite boxes that will run these tests won't need Node installed as the JS execution is done in docker by SAM, but they do need Ruby to run Maze Runner so it cuts out a dependency

We could rewrite it in JS but I think it'd be more code & dependencies (e.g. for the YAML parsing) and be less robust, e.g. `ensure` blocks always run in Ruby even after `exit` or cancelling with <kbd>Ctrl</kbd>+<kbd>c</kbd> but `finally` blocks don't run after `process.exit` or cancelling in Node

bengourley:
> We could rewrite it in JS but I think it'd be more code & dependencies (e.g. for the YAML parsing) and be less robust, e.g. ensure blocks always run in Ruby even after exit or cancelling with Ctrl+c but finally blocks don't run after process.exit or cancelling in Node

Maybe we rewrite the JS version in ruby then?

This is the script I'm talking about https://github.com/bugsnag/bugsnag-js/blob/next/bin/local-test-util

It builds the browser notifier(s) and test fixtures outside of docker for quicker iteration in development (I was mistaken when I said node too).

twometresteve:
> Maybe we rewrite the JS version in ruby then?


I would back that - it could move us a step closer to a "Maze Builder".

imjoehaines:
Created https://bugsnag.atlassian.net/browse/PLAT-6086 for rewriting `local-test-util`

PR #1304, opened 2021-02-24
Create @bugsnag/in-flight internal package

imjoehaines:
## Goal

This PR adds a new `@bugsnag/in-flight` package, to be used internally

This package exports two functions: `trackInFlight` and `flush`. `trackInFlight` will monkey-patch the Client passed to it in order to track all calls to `_notify` and it's Delivery's `sendSession`. `flush` uses this tracking to determine if there are any outstanding requests — it will resolve if there are no requests, retrying repeatedly until all requests settle or it times out

### DIFF
--- a/packages/in-flight/src/in-flight.js
+++ b/packages/in-flight/src/in-flight.js
@@ -1,0 +1,67 @@
+const cuid = require('@bugsnag/cuid')
+
+const FLUSH_POLL_INTERVAL_MS = 50
+const inFlightRequests = new Map()
+
+const noop = () => {}
+
+module.exports = {
+  trackInFlight (client) {
+    const originalNotify = client._notify
+
+    client._notify = function (event, onError, callback = noop) {
+      const id = cuid()
+      inFlightRequests.set(id, true)
+
+      const _callback = function () {
+        inFlightRequests.delete(id)
+        callback.apply(null, arguments)
+      }
+
+      client._depth += 1
+
+      try {
+        originalNotify.call(client, event, onError, _callback)
+      } finally {
+        client._depth -= 1
+      }
+    }
+
+    const delivery = client._delivery
+    const originalSendSession = delivery.sendSession
+
+    delivery.sendSession = function (session, callback = noop) {
+      const id = cuid()
+      inFlightRequests.set(id, true)
+
+      const _callback = function () {
+        inFlightRequests.delete(id)
+        callback.apply(null, arguments)
+      }
+
+      originalSendSession.call(delivery, session, _callback)
+    }
+  },
+
+  flush (timeoutMs) {
+    return new Promise(function (resolve, reject) {
+      const timeout = setTimeout(
+        () => { reject(new Error(`flush timed out after ${timeoutMs}ms`)) },
+        timeoutMs
+      )
+
+      const resolveIfNoRequests = function () {
+        if (inFlightRequests.size === 0) {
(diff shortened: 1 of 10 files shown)


packages/in-flight/src/in-flight.js
```diff
@@ -0,0 +1,67 @@
...
+
+      const resolveIfNoRequests = function () {
+        if (inFlightRequests.size === 0) {
+          clearTimeout(timeout)
+          resolve()
+
+          return
+        }
+
+        setTimeout(resolveIfNoRequests, FLUSH_POLL_INTERVAL_MS)
```
bengourley, line 61: I'm not sure how important this is, because I'm pretty sure `Promise` internally guarantees that it will only call either a resolve or reject callback once…

However, for clean/completeness' sake I think you might want to hold a reference to the poll timeout, and clear it in the in the callback for the timeout on line 49.

The hypothetical situation goes like so:

- at `t=timeoutMs` there are still requests in flight
- the timeout callback fires and `reject()` is called
- at `t>timeoutMs && t<timeoutMs + FLUSH_POLL_INTERVAL_MS` before the process manages to shutdown, the in-flight requests finish
- the `resolveIfNoRequests` callback calls `resolve()` even though `reject()` was already called

bengourley (commented):
Looks good. One nit-picky comment left inline.

My only outstanding concern is around the stacktrace generation logic that uses the _depth parameter.

Can you do a comparison of this kind of thing with and without the flush module in use, to make sure the generated stacktraces are the same?

```js
const Bugsnag = require('@bugsnag/node')

// start the facade client
Bugsnag.start('API_KEY')

// start a non-facade client
const client = Bugsnag.createClient('API_KEY')

// 1. notify with a non-error, Bugsnag creates one internally
// and then strips `_depth` frames from the stacktrace
Bugsnag.notify('string not error')
client.notify('string not error')

// 2. error with no stacktrace, Bugsnag walks the callstack
//  to generate a stacktace, then removes `_depth` frames
const err = new Error('no trace')
err.stack = ''
Bugsnag.notify(err)
client.notify(err)
```

packages/in-flight/src/in-flight.js
imjoehaines: The ES spec does guarantee that promises can only be resolved/rejected once, but I think it makes sense to do just in case a JS engine isn't totally spec compliant

> When a promise reject function F is called with argument reason, the following steps are taken:
> 
> 1. Assert: F has a [[Promise]] internal slot whose value is an Object.
> 1. Let promise be the value of F's [[Promise]] internal slot.
> 1. Let alreadyResolved be the value of F's [[AlreadyResolved]] internal slot.
> 1. If alreadyResolved.[[value]] is true, return undefined.
> 1. Set alreadyResolved.[[value]] to true.
>
> https://262.ecma-international.org/6.0/#sec-promise-reject-functions

> When a promise resolve function F is called with argument resolution, the following steps are taken:
> 
> 1. Assert: F has a [[Promise]] internal slot whose value is an Object.
> 1. Let promise be the value of F's [[Promise]] internal slot.
> 1. Let alreadyResolved be the value of F's [[AlreadyResolved]] internal slot.
> 1. If alreadyResolved.[[value]] is true, return undefined.
>
> https://262.ecma-international.org/6.0/#sec-promise-resolve-functions

bengourley (approved):

PR #1307, opened 2021-02-25
Add new AWS Lambda plugin

imjoehaines:
## Goal

This PR adds a plugin to allow Bugsnag to work in an AWS Lambda function. There is more work to do before release, such as session support, but the plugin is functional in this state

This plugin works by using the new `@bugsnag/in-flight` package so that it can wait for all requests to finish before allowing the Lambda to terminate. There is a default 2 second timeout, but this can be configured by the user to whatever makes sense for their app

## Basic usage

```js
const Bugsnag = require('@bugsnag/js')
const BugsnagPluginAwsLambda = require('@bugsnag/plugin-aws-lambda')

Bugsnag.start({
  apiKey: 'api key here',
  plugins: [BugsnagPluginAwsLambda]
})

const bugsnagHandler = Bugsnag.getPlugin('awsLambda').createHandler()

const handler = async (event, context) => // code goes here

module.exports.lambdaHandler = bugsnagHandler(handler)
```

`createHandler` takes an options object, with the only current option being `flushTimeoutMs`. This is the length of time (in milliseconds) that we will wait for requests to finish. For example:

```js
createHandler({ flushTimeoutMs: 5000 }) // 5 seconds
```

## Testing

Tested with unit tests and the MR tests have been updated to assert events are reported when using the new plugin

### DIFF
--- a/packages/plugin-aws-lambda/src/index.js
+++ b/packages/plugin-aws-lambda/src/index.js
@@ -36,7 +36,7 @@ function wrapHandler (client, flushTimeoutMs, handler) {
           severityReason: { type: 'unhandledException' }
         }
 
-        const event = client.Event.create(err, true, handledState, 1)
+        const event = client.Event.create(err, true, handledState, 'aws lambda plugin', 1)
 
         client._notify(event)
       }
--- a/packages/plugin-aws-lambda/src/index.js
+++ b/packages/plugin-aws-lambda/src/index.js
@@ -1,0 +1,82 @@
+const bugsnagInFlight = require('@bugsnag/in-flight')
+
+const BugsnagPluginAwsLambda = {
+  name: 'awsLambda',
+
+  load (client) {
+    bugsnagInFlight.trackInFlight(client)
+
+    return {
+      createHandler ({ flushTimeoutMs = 2000 } = {}) {
+        return wrapHandler.bind(null, client, flushTimeoutMs)
+      }
+    }
+  }
+}
+
+function wrapHandler (client, flushTimeoutMs, handler) {
+  let _handler = handler
+
+  if (handler.length > 2) {
+    // This is a handler expecting a 'callback' argument, so we convert
+    // it to return a Promise so '_handler' always has the same API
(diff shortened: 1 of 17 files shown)


packages/plugin-aws-lambda/src/index.js
```diff
@@ -0,0 +1,84 @@
...
+    _handler = promisifyHandler(handler)
+  }
+
+  return async function (event, context) {
+    client.addMetadata('AWS Lambda context', context)
+
+    try {
+      return await _handler(event, context)
+    } catch (err) {
+      if (client._config.autoDetectErrors && client._config.enabledErrorTypes.unhandledExceptions) {
```
bengourley, line 32: Since we're async-ly `try`ing user code, doesn't this mean we'll catch promise rejections too, preventing them from bubbling up to the `onunhandledrejection` handler?

In that case we'll fail to honour the `enabledErrorTypes.unhandledRejections` configuration. And it also means the `severityReason.type` will always be `unhandledException`.

I'm not quite sure what we should do to "fix" this, because really with the `async` keyword, the concept of unhandled exceptions/rejections are somewhat muddied.

packages/plugin-aws-lambda/src/index.js
```diff
@@ -0,0 +1,84 @@
...
+        const handledState = {
+          severity: 'error',
+          unhandled: true,
+          severityReason: { type: 'unhandledException' }
+        }
+
+        const event = client.Event.create(err, true, handledState, 1)
```
bengourley, line 39: `Event.create` has the signature `(maybeError, tolerateNonErrors, handledState, component, errorFramesToSkip = 0, logger)` – is `1` meant to be the value for `framesToSkip`?

imjoehaines: Yeah, I think this was a mistake [copied from the Vue plugin](https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-vue/src/vue.js#L6) — I'll make a ticket to fix it there too

bengourley (approved):

PR #1309, opened 2021-02-26
Session support for AWS Lambda functions

imjoehaines:
## Goal

The existing Node session plugin uses a model that works well for long running processes (like traditional webservers) but won't work for short lived Lambda executions. To fix this we can use the browser session plugin, which isn't actually browser specific. This sends a session every time `startSession` is called, which is exactly what we want in a Lambda function

As always, the `autoTrackSessions` option can be disabled to prevent the automatic `startSession` call

## Testing

Unit tests have been updated and the MazeRunner tests also check for sessions

### DIFF
--- a/packages/plugin-aws-lambda/package.json
+++ b/packages/plugin-aws-lambda/package.json
@@ -24,7 +24,8 @@
   "author": "Bugsnag",
   "license": "MIT",
   "dependencies": {
-    "@bugsnag/in-flight": "^7.7.0"
+    "@bugsnag/in-flight": "^7.7.0",
+    "@bugsnag/plugin-browser-session": "^7.7.0"
   },
   "devDependencies": {
     "@bugsnag/core": "^7.7.0"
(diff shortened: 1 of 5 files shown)


bengourley (commented):
Looks good.

The maze runner tests could do with some assertions on the `event.session` part of the payload – currently the tests verify that sessions are sent, but an equally important piece of information is the handled/unhandled counts in the event payload.

imjoehaines:
> Looks good.
> 
> The maze runner tests could do with some assertions on the `event.session` part of the payload – currently the tests verify that sessions are sent, but an equally important piece of information is the handled/unhandled counts in the event payload.

Session tests have been improved in a follow up PR to avoid having to deal with conflicts from changing them here: https://github.com/bugsnag/bugsnag-js/pull/1312/commits/976ff62a090fe902d31c8cd6ec56f5b51e97dd3e

bengourley (approved):

PR #1310, opened 2021-02-26
Calculate the duration for Lambda functions

imjoehaines:
## Goal

Our existing plugin-app-duration doesn't report the correct duration for Lambda functions because its 'appStart' is cached between invocations. This means the duration would essentially be the Lambda uptime, rather than the invocation time

We now reset the duration in the AWS plugin's `load` method so it starts fresh for each invocation

### DIFF
--- a/packages/plugin-aws-lambda/src/index.js
+++ b/packages/plugin-aws-lambda/src/index.js
@@ -8,6 +8,13 @@ const BugsnagPluginAwsLambda = {
     bugsnagInFlight.trackInFlight(client)
     client._loadPlugin(BugsnagPluginBrowserSession)
 
+    // Reset the app duration between invocations, if the plugin is loaded
+    const appDurationPlugin = client.getPlugin('appDuration')
+
+    if (appDurationPlugin) {
+      appDurationPlugin.reset()
+    }
+
     return {
       createHandler ({ flushTimeoutMs = 2000 } = {}) {
         return wrapHandler.bind(null, client, flushTimeoutMs)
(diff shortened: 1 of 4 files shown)


packages/plugin-aws-lambda/src/index.js
```diff
@@ -26,6 +26,14 @@ function wrapHandler (client, flushTimeoutMs, handler) {
   }
 
   return async function (event, context) {
+    const startTime = new Date()
+
+    client.addOnError(event => {
+      const endTime = new Date()
+
+      event.app.duration = endTime - startTime
```
bengourley, line 34: Presumably setting this here means that it overwrites the value set by the `plugin-app-duration` plugin? In that case all other `onError` callbacks supplied by the user will see the yet-to-be-overwritten value.

That might be a tolerable side-effect of keeping this implementation simple.

The "in pursuit of full correctness" solution, for comparison might look something like this:

`plugin-app-duration/app.js`
```js
let appStart
const reset = () => { appStart = new Date() }
reset()

module.exports = {
  name: 'appDuration',
  load: client => {
    client.addOnError(event => {
      const now = new Date()

      event.app.duration = now - appStart
    }, true)

    return { reset }
  }
}
```

`plugin-aws-lambda/src/index.js`
```js
// ...
  return async function (event, context) {
    client.getPlugin('appDuration').reset()
    // ...
  })
```

bengourley (commented):
This looks reasonable. I just want to check you consciously ruled out the alternative solution I've outlined in favour of pragmatism/simplicity.

packages/plugin-aws-lambda/src/index.js
imjoehaines: Yeah, I think your way makes more sense. It's probably also more accurate to call `reset` in the `load` method itself rather than the lambda wrapper

bengourley (approved):

PR #1316, opened 2021-03-02
Handle promise rejections in AWS Lambda functions

imjoehaines:
## Goal

AWS have a promise rejection handler that forcefully exits the process. This means we can't send reports for promise rejections because 1) our handler is never called as it is always added after AWS' and 2) there's no time to deliver the event anyway

We can work around this by removing AWS' handler and calling it after delivery of the promise rejection

## Testing

MazeRunner tests added to cover this — they pass with & without the new handler, the only difference being the stacktrace

### DIFF
--- a/packages/plugin-aws-lambda/src/index.js
+++ b/packages/plugin-aws-lambda/src/index.js
@@ -22,14 +22,13 @@ const BugsnagPluginAwsLambda = {
       const listeners = process.listeners('unhandledRejection')
       process.removeAllListeners('unhandledRejection')
 
-      const BugsnagPluginUnhandledRejection = require('@bugsnag/plugin-node-unhandled-rejection')
-      client._loadPlugin(BugsnagPluginUnhandledRejection)
-
-      const originalOnUnhandledRejection = client._config.onUnhandledRejection
-      client._config.onUnhandledRejection = (reason, event, logger) => {
-        originalOnUnhandledRejection(reason, event, logger)
-        listeners.forEach(listener => { listener(reason) })
-      }
+      // This relies on our unhandled rejection plugin adding its listener first
+      // using process.prependListener, so we can call it first instead of AWS'
+      process.on('unhandledRejection', async (reason, promise) => {
+        for (const listener of listeners) {
+          await listener.call(process, reason, promise)
+        }
+      })
     }
 
     return {
--- a/packages/plugin-aws-lambda/src/index.js
+++ b/packages/plugin-aws-lambda/src/index.js
@@ -15,6 +15,23 @@ const BugsnagPluginAwsLambda = {
       appDurationPlugin.reset()
     }
 
+    // AWS add a default unhandledRejection listener that forcefully exits the
+    // process. This breaks reporting of unhandled rejections, so we have to
+    // remove all existing listeners and call them after we handle the rejection
+    if (client._config.autoDetectErrors && client._config.enabledErrorTypes.unhandledRejections) {
+      const listeners = process.listeners('unhandledRejection')
+      process.removeAllListeners('unhandledRejection')
+
+      const BugsnagPluginUnhandledRejection = require('@bugsnag/plugin-node-unhandled-rejection')
+      client._loadPlugin(BugsnagPluginUnhandledRejection)
+
+      const originalOnUnhandledRejection = client._config.onUnhandledRejection
+      client._config.onUnhandledRejection = (reason, event, logger) => {
+        originalOnUnhandledRejection(reason, event, logger)
+        listeners.forEach(listener => { listener(reason) })
+      }
+    }
+
     return {
       createHandler ({ flushTimeoutMs = 2000 } = {}) {
         return wrapHandler.bind(null, client, flushTimeoutMs)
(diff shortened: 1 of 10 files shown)


packages/plugin-aws-lambda/src/index.js
```diff
@@ -15,6 +15,23 @@ const BugsnagPluginAwsLambda = {
...
     }
 
+    // AWS add a default unhandledRejection listener that forcefully exits the
+    // process. This breaks reporting of unhandled rejections, so we have to
+    // remove all existing listeners and call them after we handle the rejection
+    if (client._config.autoDetectErrors && client._config.enabledErrorTypes.unhandledRejections) {
+      const listeners = process.listeners('unhandledRejection')
+      process.removeAllListeners('unhandledRejection')
+
+      const BugsnagPluginUnhandledRejection = require('@bugsnag/plugin-node-unhandled-rejection')
```
bengourley, line 25: Couple of issues here.

Modules/duplication:
- `require`s should really be at the top of the file
- Since this is bundled, the fact it is low in the file doesn't defer it being loaded (though it does defer the execution of its module scope) and it will result in it being present twice in the app – once bundled in `@bugsnag/node` and once in `@bugsnag/plugin-aws-lambda`

Double execution
- I'm pretty sure the logic here means that the plugin gets executed twice? The second time it won't be around for long enough to actually send the report due to the AWS callback forcefully executing, but there might be duplicate logs?
- The plugin actually has a (non-standard) `destroy()` method that you could call before taking a snapshot of the process `unhandledError` listeners, however, you can't get a reference to the plugin due to the module duplication mentioned above.

packages/plugin-aws-lambda/src/index.js
```diff
@@ -15,6 +15,23 @@ const BugsnagPluginAwsLambda = {
...
+      client._loadPlugin(BugsnagPluginUnhandledRejection)
+
+      const originalOnUnhandledRejection = client._config.onUnhandledRejection
+      client._config.onUnhandledRejection = (reason, event, logger) => {
+        originalOnUnhandledRejection(reason, event, logger)
+        listeners.forEach(listener => { listener(reason) })
```
bengourley, line 31: listeners also expect the offending `promise` as the second param

bengourley (commented):
I have an idea for a possible improvement to this:

1. Update `@bugsnag/plugin-node-unhandled-rejection` to use [`prependListener()`](https://nodejs.org/dist/latest-v14.x/docs/api/events.html#events_emitter_prependlistener_eventname_listener) [here](https://github.com/bugsnag/bugsnag-js/blob/ae4a01755ca12d70adc2e8a7a4d8a4d6c6278449/packages/plugin-node-unhandled-rejection/unhandled-rejection.js#L16)
2. That means that all of the `unhandledRejection` callbacks are now in the desired order, but since they will execute in parallel, the process will still exit before the event has been delivered, remove the listeners like you did before in `plugin-aws-lambda/src/index.js`, but this time add one new listener which executes the existing ones in serial:
    ```js
    const listeners = process.listeners('unhandledRejection')
    process.removeAllListeners('unhandledRejection')
    process.on('unhandledRejection', async (reason, promise) => {
      // iterate through each listener, waiting for each to
      // resolve before executing the next
      for (const listener of listeners) {
        await listener.call(process, reason, promise)
      }
    })
    ```
3. Now all that's left is for the unhandled rejection plugin to actually return a promise, signalling when it is actually done:
    ```js
    _handler = err => {
      const event = client.Event.create(err, false, {
        severity: 'error',
        unhandled: true,
        severityReason: { type: 'unhandledPromiseRejection' }
      }, 'unhandledRejection handler', 1)
      return new Promise((resolve) => client._notify(event, () => {}, (e, event) => {
        if (e) client._logger.error('Failed to send event to Bugsnag')
        client._config.onUnhandledRejection(err, event, client._logger)
        resolve()
      })
    }
    ```

What do you think?

packages/plugin-aws-lambda/src/index.js
imjoehaines: They do, but we can't get a reference to it here

packages/plugin-aws-lambda/src/index.js
imjoehaines: How can we avoid having it present twice?

imjoehaines:
>     1. Update `@bugsnag/plugin-node-unhandled-rejection` to use [`prependListener()`](https://nodejs.org/dist/latest-v14.x/docs/api/events.html#events_emitter_prependlistener_eventname_listener) [here](https://github.com/bugsnag/bugsnag-js/blob/ae4a01755ca12d70adc2e8a7a4d8a4d6c6278449/packages/plugin-node-unhandled-rejection/unhandled-rejection.js#L16)

Are we OK to mess with the order for non-Lambda apps? I don't think it's necessarily a realistic problem but someone could be relying on their callback firing first

It also only exists on Node 6+ so we'd have to fallback to not prepending on older versions, which is a bit weird but not the end of the world

bengourley:
> Are we OK to mess with the order for non-Lambda apps? I don't think it's necessarily a realistic problem but someone could be relying on their callback firing first

I don't think it should matter – we do very little synchronous work. I'm happy to insert ourselves before any other listeners as I'd be surprised if anybody relies on the order of `unhandledRejection` events.

bengourley (approved):

PR #1317, opened 2021-03-02
Fix app duration test flakes

imjoehaines:
## Goal

On GH actions sometimes the timeout can be invoked slightly too soon in the new `reset` test, e.g. after 24ms instead of 25

This PR builds in 5ms of leeway to each assertion, which should hopefully be enough to ensure it doesn't flake


bengourley (approved):
Weird!

PR #1319, opened 2021-03-02
fix(expo): Ensure Expo plugins depend on same version of NetInfo package

bengourley:
Ensures the plugins used by `@bugsnag/expo` all depend on the same version of NetInfo. [Expo 40 depends on v5.9.7](https://github.com/expo/expo/blob/45dd06df4c3d844e8adf774207d64015e76ec902/packages/expo/bundledNativeModules.json#L3).

I also updated the CONTRIBUTING.md guide because it was missing the connectivity breadcrumbs package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+## Fixed
+
+- (expo): Ensure Expo plugins depend on same version of NetInfo package [#1319](https://github.com/bugsnag/bugsnag-js/pull/1319)
+
 ## v7.8.1 (2021-02-25)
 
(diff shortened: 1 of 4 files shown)


imjoehaines (commented):
I think `5.9.7` is right as it's what Expo SDK 40 uses — https://github.com/expo/expo/blob/45dd06df4c3d844e8adf774207d64015e76ec902/packages/expo/bundledNativeModules.json#L3

imjoehaines (approved):

bengourley:
My bad – I was trawling through the Expo repo since they don't tag their releases, and missed that 40 was already out. Changed and rebased.

PR #1324, opened 2021-03-04
Automatically notify when a lambda may timeout

imjoehaines:
## Goal

We can never be 100% sure that a timeout will happen because we need to notify with enough time for the event to be delivered. Therefore this is a best effort with a generous default value (1000ms) that errs on the side of delivering the warning at the cost of possible false positives. The default value can be changed by providing the new `lambdaTimeoutNotifyMs` parameter to `createHandler`, e.g.:

```js
const bugsnagHandler = Bugsnag.getPlugin('awsLambda').createHandler({
  lambdaTimeoutNotifyMs: 500
})
```

If set to "0" we don't bother adding the warning at all. This is largely arbitrary as a value of 0 would never be triggered (because the lambda would timeout at the same time so we would never be able to deliver the event) but it allows users to turn this feature off if they don't want it

## Testing

This can only be covered by unit tests because of a bug in AWS' SAM CLI: https://github.com/aws/aws-sam-cli/issues/2519

In short, SAM returns a Unix timestamp from `getRemainingTimeInMillis` so the timeout could never be triggered with `sam local invoke` (short of mocking the function, which isn't really better than a unit test)

### DIFF
--- a/packages/plugin-aws-lambda/src/lambda-timeout-approaching.js
+++ b/packages/plugin-aws-lambda/src/lambda-timeout-approaching.js
@@ -1,0 +1,11 @@
+module.exports = class LambdaTimeoutApproaching extends Error {
+  constructor (remainingMs) {
+    const message = `Lambda will timeout in ${remainingMs}ms`
+
+    super(message)
+
+    this.name = 'LambdaTimeoutApproaching'
+    this.message = message
+    this.stack = []
+  }
+}
--- a/packages/plugin-aws-lambda/src/lambda-timeout-approaching.js
+++ b/packages/plugin-aws-lambda/src/lambda-timeout-approaching.js
@@ -5,7 +5,6 @@ module.exports = class LambdaTimeoutApproaching extends Error {
     super(message)
 
     this.name = 'LambdaTimeoutApproaching'
-    this.message = message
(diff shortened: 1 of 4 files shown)


packages/plugin-aws-lambda/src/lambda-timeout-approaching.js
```diff
@@ -0,0 +1,11 @@
+module.exports = class LambdaTimeoutApproaching extends Error {
+  constructor (remainingMs) {
+    const message = `Lambda will timeout in ${remainingMs}ms`
+
+    super(message)
+
+    this.name = 'LambdaTimeoutApproaching'
+    this.message = message
```
bengourley, line 8: is this line necessary? Calling `super(message)` should take care of setting it

bengourley (approved):
One minor comment inline.

packages/plugin-aws-lambda/src/lambda-timeout-approaching.js
imjoehaines: Yeah it should be unnecessary — it's here because originally I wasn't calling `super` but I forgot you *have* to call `super` in JS to use `this`

PR #1331, opened 2021-03-09
Add AWS Lambda tests to CI

imjoehaines:
## Goal

Adds the AWS Lambda tests to CI. This was failing to install the top-level dependencies because NPM resolved the dependencies differently to Artifactory (which is the default on CI) — I'm not sure why this happens though!

NPM has resolved both [`@babel/helper-define-map`](https://github.com/bugsnag/bugsnag-js/blob/next/package-lock.json#L454-L457) and [`@babel/helper-explode-assignable-expression`](https://github.com/bugsnag/bugsnag-js/blob/next/package-lock.json#L484-L487) to the same package (`@babel/helper-define-map`). When Artifactory is used as the registry, it doesn't do this so the checksums don't match — because it's downloaded them as two different packages

This is weird but doesn't seem to break anything and NPM does the exact same thing if you recreate the lockfile so I don't think it's a bug per-se. To fix it, we can tell NPM to use its registry instead of Artifactory when installing the top-level dependencies

The CI boxes also have an older version of Bundler which refused to install from our Gemfile.lock — we don't really need the lockfile because we pin our single dependency (MazeRunner) to a specific tag, so I opted to delete the lockfile instead of changing the CI environment

The 35 minute timeout is ~10 minutes longer than it currently takes to run, so it has a bit of room


twometresteve (approved):
LGTM

PR #1334, opened 2021-03-11
[ROAD-756] AWS Lambda support

bengourley:
This is the integration branch for all of the work that has gone into the support for Node.js on AWS Lambda.

I'm creating this PR to cut a prerelease and collect any UAT feedback.


bengourley (commented):
Done with UAT. All working as expected 👍

imjoehaines (approved):

PR #1338, opened 2021-03-16
Ensure step fails if IPA export does

twometresteve:
## Goal

Ensure the Buildkite step fails if exporting the IPA does.

## Design

If the IPA is not created for any reason, `cp` will return a non-zero code.  With this passed back to the called (Buildkite agent), the step will fail and a human can examine the log further to understand why, before potentially hitting retry.

## Testing

I've performed an initial cut-down CI run using both good and known-broken build servers.  The steps passed and failed accordingly.


bengourley (approved):
Approved but please hold the merge until the current release process is complete (it makes for a non-linear history on next when anything lands while the release is made, which it's nice to avoid).

PR #1341, opened 2021-03-24
Remove android vendoring from the repo and use a script to pull on demand instead

kstenerud:
## Changeset

Updated build scripts to pull in android vendoring on demand, and removed existing vendoring from the repo.

## Testing

Re-ran build scripts


packages/react-native/prepare-android-vendor.sh
```diff
@@ -0,0 +1,86 @@
...
+trap cleanup EXIT
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+SCRIPT_NAME="${0##*/}"
+CONFIG_FILE="$SCRIPT_DIR/prepare-android-vendor.config"
+
+MODE=$(head -1 "$CONFIG_FILE")
+ARG=$(tail -1 "$CONFIG_FILE")
+
+MAVEN_REPO_DIR="$HOME/.m2/repository/com/bugsnag"
```
fractalwrench, line 20: The intent behind this change is that the artefacts are downloaded from [maven central](https://search.maven.org/) so that it's not necessary to build them twice, which could potentially result in different binaries if done on different machines. It doesn't look like this would achieve that requirement as it's building the artefact locally.

fractalwrench (changes requested):
This removes a large number of Android files from VCS which makes it impossible to build the React Native package.

The `android/com/bugsnag` directory need removing from VCS but the rest of the files (such as `BugsnagReactNative) need to stay in place. Getting CI passing or testing locally with Verdaccio would be enough to confirm that the changeset works.

It's also worth noting that the intent of the change was to allow downloading the artefacts from mavenCentral directly - it doesn't look like that's currently an option in the script.

PR #1343, opened 2021-03-25
Remove android vendoring from the repo and use a script to pull on demand instead

kstenerud:
## Changeset

Removed android vendoring from the repo and use a script to pull on demand instead

## Testing

reran test scripts


packages/react-native/.gitignore
```diff
@@ -31,3 +31,4 @@ DerivedData
 *.ipa
 *.xcuserstate
 project.xcworkspace
+android
```
fractalwrench, line 34: This is a bit too broad - it currently ignores _all_ files in the `android` directory, whereas it's only necessary to ignore the directory containing the artefacts.

packages/react-native/update-ios.sh
```diff
@@ -115,9 +119,7 @@ revendor_from_clean_repo() {
 	echo "Checking out https://github.com/bugsnag/bugsnag-cocoa.git with tag $tag"
 	TEMP_DIR="$(mktemp -d)"
 	pushd "$TEMP_DIR" >/dev/null
-	git clone https://github.com/bugsnag/bugsnag-cocoa.git
-	cd bugsnag-cocoa
-	git checkout $tag || exit 1
+	git clone https://github.com/bugsnag/bugsnag-cocoa.git --depth 1 --branch $tag
```
fractalwrench, line 122: I tried this out for all the options, and it seems to work for everything apart from checking out a SHA, which fails with the following error message:

```
./update-ios.sh --sha 5a00246
Checking out https://github.com/bugsnag/bugsnag-cocoa.git with tag 5a00246
Cloning into 'bugsnag-cocoa'...
warning: Could not find remote branch 5a00246 to clone.
fatal: Remote branch 5a00246 not found in upstream origin
```

The behaviour is the same on Android/Cocoa

fractalwrench (commented):
This is looking good - I spotted a couple of things and left some inline comments. It also looks like the contributing guide for iOS/Android could use a small update detailing what the new instructions are.

packages/react-native/.gitignore
kstenerud: Changed to android/com/bugsnag

packages/react-native/update-ios.sh
kstenerud: Updated to use clone depth 1 only for tags and branches. SHAs now checkout the entire repo and checkout that sha.

kstenerud:
Updated instructions as well.

fractalwrench (approved):
LGTM. Happy for this to be merged on the assumption that the `--sha` option has been manually retested.

kstenerud:
Yup, I did manual retesting

bengourley (changes requested):
Looks good and mostly works well.

One piece of logic needs moving from the prepare to the update script (commented inline).

I would also like to query if both `packages/react-native/prepare-android-vendor.config` and `packages/react-native/android/.bugsnag-android-version` are required? They seem to do very similar jobs? The latter only appearing to be updated if vendoring from a directory and so presumably out of date for any other kind of update?

packages/react-native/prepare-android-vendor.sh
```diff
@@ -0,0 +1,132 @@
...
+        local dstdir="$bugsnag_dir/$name"
+        mkdir -p "$dstdir"
+        pushd "$dstdir"
+        for variant in ${REPO_FILE_VARIANTS[@]}; do
+            curl https://repo1.maven.org/maven2/com/bugsnag/$name/$version/$name-$version$variant >>$name-$version$variant
+        done
+    done
+
+    sed_in_place "s/api \"com.bugsnag:bugsnag-android:.*/api \"com.bugsnag:bugsnag-android:$version\"/" "$dst_dir/build.gradle"
+    sed_in_place "s/api \"com.bugsnag:bugsnag-plugin-react-native:.*/api \"com.bugsnag:bugsnag-plugin-react-native:$version\"/" "$dst_dir/build.gradle"
```
bengourley, line 102: This logic (and the other instances of `sed_in_place` in this file) makes the git tree dirty again after running `prepare-android-vendor.sh`. It should be moved in to the `update-android.sh` script because its output needs to be checked in. The stage at which the prepare script is run during a release is too late to make changes to the checked in files.

kstenerud:
I don't actually know what packages/react-native/android/.bugsnag-android-version is for. It's pre-existing, and nothing in the repository refers to it, other than the script which creates it. Maybe one of the CI tools needs it?

bengourley (approved):
Looks good.

I manually tested a locally published version of this using the `--version` and `--sha` variants of the command and it worked as expected.

👍

Issue #1344, opened 2021-03-26
plugin-aws-lambda has imcompatible typing with aws-lambda in TypeScript 4.0.2


yskkin:
### Describe the bug

`createHandler` in plugin-aws-lambda has imcompatible typing with aws-lambda

### Steps to reproduce
Compile following code.

```ts
import Bugsnag from "@bugsnag/js";
import BugsnagPluginAwsLambda from "@bugsnag/plugin-aws-lambda";
import { APIGatewayProxyHandler } from "aws-lambda";

export const wrapHandler = (func: APIGatewayProxyHandler): APIGatewayProxyHandler => {
  Bugsnag.start({
    apiKey: process.env.BUGSNAG_API_KEY as string,
    plugins: [BugsnagPluginAwsLambda],
  });
  const bugsnagHandler = Bugsnag.getPlugin("awsLambda")?.createHandler();
  return bugsnagHandler ? bugsnagHandler(func) : func;
};
```
I expected this compiles fine.
Actual output:
```
error TS2345: Argument of type 'Handler<APIGatewayProxyEvent, APIGatewayProxyResult>' is not assignable to parameter of type 'AsyncHandler | CallbackHandler'.
  Type 'Handler<APIGatewayProxyEvent, APIGatewayProxyResult>' is not assignable to type 'AsyncHandler'.

28   return bugsnagHandler ? bugsnagHandler(func) : func;
```

### Environment
aws-lambda: 8.10.72
@bugsnag/js: 7.9.0
@bugsnag/plugin-aws-lambda: 7.9.0
typescript: 4.0.2

johnkiely1:
Hi @yskkin, 

Thanks for raising, we will look to get this resolved as soon as we can.

PR #1353, opened 2021-04-06
Align Callback with @types/aws-lambda

h-kishi:
## Goal

The goal is to align the callback types of [@bugsnag/plugin-aws-lambda](https://www.npmjs.com/package/@bugsnag/plugin-aws-lambda) and [@types/aws-lambda](https://www.npmjs.com/package/@types/aws-lambda).
`@types/aws-lambda` is often used to develop lambda functions with TypeScript.
The type of lambda callback is defined in `@types/aws-lambda`, but the type is different from the callback of CallbackHandler in `@bugsnag/plugin-aws-lambda`.

https://github.com/bugsnag/bugsnag-js/blob/51a4422e2a84b2c9732272a9aacf39c34a6e0db3/packages/plugin-aws-lambda/types/bugsnag-plugin-aws-lambda.d.ts#L7

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b9b6b0e9bd5de6ed18be0a3e07816f60fe6f2cfa/types/aws-lambda/handler.d.ts#L171
```typescript
export type Callback<TResult = any> = (error?: Error | string | null, result?: TResult) => void;
```

(I guess that this is the cause of #1344)

## Testing

I added a test for the pattern of passing a string to a callback.


johnkiely1:
Hey @h-kishi, 

Thanks for this, we will look to get this reviewed as soon as priorities allow.

imjoehaines (approved):
Thanks @h-kishi!

Can confirm this fixes #1344

PR #1363, opened 2021-04-16
update Next.js example

djskinner:
Work in progress revised next.js example. With inspiration from https://github.com/vercel/next.js/tree/canary/examples/with-sentry

NB: This is currently added as a new example side-by-side with the old one for easier reviewing whilst WIP. Eventually it will replace the existing example.

# Known issues

## Session started twice on page load

This seems to be because nextjs triggers a router (history) change following load and so two requests to sessions.bugsnag.com can be seen on page load. Not currently sure how to prevent this.

## Source maps

Uploaded source maps do appear to work but there are some issues that need ironing out:

- No source maps for server-side code. It doesn't look like next.js generates these anyway.
- Many events have source map warnings even though mapping seems to have been applied.

## react-plugin

Need to add the bugsnag react plugin

# To Do

- [ ] More SSR test pages
- [ ] Make styling etc consistent with existing bugsnag examples
- [ ] Add error handling support for API routes
- [ ] Add support for vercel / serverless


examples/js/nextjs-new/pages/api/handled3.js
```diff
@@ -0,0 +1,12 @@
+import Bugsnag from '@bugsnag/js';
+import { start } from '../../lib/bugsnag'
+
+start();
+
+const handler = async (req, res) => {
+  Bugsnag.notify(new Error('a handled error occurred'))
+  await require('@bugsnag/in-flight').flush(2000)
```
djskinner, line 8: This only notifies Bugsnag intermittently when deployed to Vercel.

examples/js/nextjs-new/pages/api/handled4.js
```diff
@@ -0,0 +1,13 @@
+import Bugsnag from '@bugsnag/js';
+import { start } from '../../lib/bugsnag'
+
+start();
+
+const handler = async (req, res) => {
+  await new Promise(resolve =>
+    Bugsnag.notify(new Error('a handled error occurred'), undefined, () => resolve())
+  )
+  res.status(200).json({ name: 'John Doe' })
```
djskinner, line 10: This approach seems to notify consistently when deployed to vercel

shaohaolin:
Can you provide an example for error boundary being use for this case?

examples/js/nextjs-new/lib/bugsnag.js
```diff
@@ -0,0 +1,21 @@
+import Bugsnag from '@bugsnag/js'
+import BugsnagPluginAwsLambda from '@bugsnag/plugin-aws-lambda'
+
+export function start() {
+  // TODO NEXT_PHASE seems to be undocumented. We're doing this because we don't want
+  // Bugsnag to run during the build phase
```
djskinner, line 6: Looks like phases are slightly more documented now. See https://nextjs.org/docs/api-reference/next.config.js/introduction and https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/constants.ts#L1-L5

PR #1364, opened 2021-04-20
License auditing script

twometresteve:
## Goal

Enforce license checks in CI.

## Design

The approach here needed to be slightly more sophisticated than our other repos due to our use of Lerna.  It performs a `lerna bootstrap` before running the license_finder tool over all of the `packages/`.

## Changeset

New CI step with supporting scripts.  The github status check for the new step is now required.

## Testing

Covered by CI and licenses not covered by the approval decision files were observed to fail the step before new decisions were added and approved.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,7 +4,7 @@ steps:
   # License audit
   #
   - label: ':copyright: License Audit'
-    timeout_in_minutes: 10
+    timeout_in_minutes: 20
     agents:
       queue: opensource-mac-cocoa-11
     command: scripts/license_finder.sh
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,144 +1,153 @@
 steps:
 
   #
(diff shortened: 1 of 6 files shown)


Cawllec (approved):
LGTM 👍

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,14 @@
 steps:
 
+  #
+  # License audit
+  #
+  - label: ':copyright: License Audit'
+    timeout_in_minutes: 20
+    agents:
+      queue: opensource-mac-cocoa-11
```
bengourley, line 9: why does this need to run on a mac?

bengourley (approved):

.buildkite/pipeline.yml
twometresteve: Not as such, I don't think - but I had problems running the license_finder tool in its Docker image, either on CI or locally.  It yielded different results, so I chose to go for one of our Macs as I had the most confidence in it working.

PR #1365, opened 2021-04-21
[PLAT-6128] RN-CLI: Use a more exhaustive search for MainApplication.java

bengourley:
Ejected Expo projects have a different directory structure to RN projects generated with the RN CLI. This update uses the glob package to determine the location of MainApplication.java no matter where it exists within the app directory.

## Goal

Ensure the CLI works with the default directory structure of an ejected Expo app.

## Design

The `glob` module was added to recursively search the app directory for `MainApplication.java`. This means the CLI can cope with various directory structures.

## Changeset

- Added `glob` dependency
- Updated unit test mocking

## Testing

Relying on updated unit tests and existing automated tests.
As well as catching regressions on the existing fixture apps, I'm currently adding a new ejected Expo fixture to the automated tests to determine that the updated logic does works as intended.


test/react-native-cli/features/fixtures/Dockerfile
```diff
@@ -10,5 +10,6 @@ COPY . /app
 WORKDIR /app
 
 RUN npm i -g bugsnag-react-native-cli-*.tgz
+RUN npm i
```
bengourley, line 13: Without this, the project dependencies get installed as a side effect of installing `@bugsnag/source-maps` or `@bugsnag/react-native` during the tests. Since the new fixture has more dependencies, it caused the tests to time out. This means the project deps are installed at build time rather than during the tests, which should also speed up the other tests due to the build being cache-able.

test/react-native-cli/features/steps/steps.rb
```diff
@@ -2,18 +2,18 @@
 require 'securerandom'
 
 fixtures = Dir["#{__dir__}/../fixtures/rn0_*"].map { |dir| File.basename(dir) }.sort
-current_fixture = ENV['REACT_NATIVE_VERSION']
+$current_fixture = ENV['REACT_NATIVE_VERSION']
```
bengourley, line 5: The Ruby-n00bie in me struggled to access this variable inside of the new `get_android_main_application_path` function I wrote without making it global. I'm sure there's a nicer way to do it.

imjoehaines: `get_android_main_application_path` seems to do the same thing regardless of `current_fixture` — is it no longer useful?

I think the best & easiest way to make it not-global would be to pass it as a parameter

bengourley: It was doing the same thing due to a copy pasta error :) I fixed the logic in c3a630d. I'll unglobalise it and pass `current_fixture` as a parameter.

test/react-native-cli/features/steps/steps.rb
```diff
@@ -85,6 +85,8 @@ def parse_package_json
   # once and sometimes appear twice, depending on if another command is running
   # when it's input
   json = after.drop_while { |line| line.include?('cat package.json') }
+  # Drop lines until we get to the start of the JSON
+  json = after.drop_while { |line| line != '{' }
```
imjoehaines, line 89: We don't actually need the other `drop_while` since this line will also remove the "cat package.json" lines too — sorry if that wasn't clear

test/react-native-cli/features/steps/steps.rb
```diff
@@ -210,9 +212,16 @@ def parse_package_json
...
 end
 
-Then("the Android app contains the bugsnag initialisation code") do
-  filename = "android/app/src/main/java/com/#{current_fixture}/MainApplication.java"
+def get_android_main_application_path(current_fixture)
+  if current_fixture.include? 'expo_ejected'
+     "android/app/src/main/java/com/bugsnag/#{current_fixture}/MainApplication.java"
+  else
+    "android/app/src/main/java/com/#{current_fixture}/MainApplication.java"
+  end
```
imjoehaines, line 220: There's an extra space here 😱

```suggestion
  if current_fixture.include? 'expo_ejected'
    "android/app/src/main/java/com/bugsnag/#{current_fixture}/MainApplication.java"
  else
    "android/app/src/main/java/com/#{current_fixture}/MainApplication.java"
  end
```

imjoehaines (approved):
Changes look good — approved assuming fixing the timeout won't change much of the main focus of this PR

twometresteve (commented):
No comments from me.

PR #1367, opened 2021-04-22
React Native tests with pre-release native notifiers

twometresteve:
## Goal

Provides a mechanism for running integration test of the React Native notifier, using the pre-release Android and Cocoa notifiers from their respective `next` branches.

## Design

The `RN_INTEGRATION` build environment variable can be set (to "true") either on a Buildkite schedule, trigger step, or when creating a build manually.  If set, the pipeline will pull in and build the latest `next` version of bugsnag-android and bugsnag-cocoa before publishing the notifier.  It will then run only the React Native e2e tests.

If run in this mode, the version for the published notifier is based on the current timestamp, rather than the current commit Id of bugsnag-js.  Publishing is skipped if the same version is used as a previous build and even though bugsnag-js may not have changed, the native notifier may have.

An improvement to the current would be to somehow build in a check that the latest native notifiers really have been build in, but I couldn't think of a way to achieve that.

## Testing

Covered by CI and inspection of the logs.  I've also run it three times via a schedule.


nickdowell:
The logic LGTM, but I find `RN_INTEGRATION` to be quite a vague name, whereas something like `BUILD_WITH_LATEST_NATIVE_NOTIFIERS` would be more obvious 🙃

fractalwrench (approved):

scripts/common.js
```diff
@@ -30,12 +30,19 @@ module.exports = {
...
+    // Base version for RN integration builds on the current time
+    if (process.env.RN_INTEGRATION) {
+      const timestamp = Math.round(Date.now() / 1000)
+
+      return `${lernaVersion}-rn-integration.${timestamp}`
+    } else {
+      // Distinguish from local use
+      const ciIndicator = (process.env.BUILDKITE ? 'ci-' : '')
+
+      return `${lernaVersion}-${ciIndicator}${branchName}.${commitId}`
```
bengourley, line 45: if you base the version off of:

`checksum(bugsnagJsSha + bugsnagAndroidSha + bugsnagCocoaSha)`

do we get caching for free?

bengourley (approved):
Agree with Nick's comment about the env var name. Also a question about repeated builds/caching to avoid building exactly the same release more than once.

Otherwise LGTM.

twometresteve:
> The logic LGTM, but I find `RN_INTEGRATION` to be quite a vague name, whereas something like `BUILD_WITH_LATEST_NATIVE_NOTIFIERS` would be more obvious 🙃

Changing to `BUILD_RN_WITH_LATEST_NATIVES` - slightly shorter whilst also mentioning React Native.

scripts/common.js
twometresteve: Discussed IRL and decided we're happy with how it is.

PR #1373, opened 2021-04-28
ci: run electron runner tests on GH actions

kattrali:
The other unit test configuration is alpine-based, which uses musl libc and is currently incompatible with running electron

## Testing

* If the unit tests pass, we win


kattrali:
oh shucks. I broke coverage bot because the `test:coverage` task doesn't exist in the "before" changeset.

bengourley (approved):

PR #1375, opened 2021-04-29
Fix node build

imjoehaines:
## Goal

The `process` require is breaking the node build script as browserify can't find this module

I'm still not sure why this started breaking, but the require should be unnecessary anyway as `process` is a global

### DIFF
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -1,7 +1,6 @@
 const { schema } = require('@bugsnag/core/config')
 const stringWithLength = require('@bugsnag/core/lib/validators/string-with-length')
 const os = require('os')
-const process = require('process')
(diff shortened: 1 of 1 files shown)


bengourley (approved):
Slightly disconcerting. Would be ideal if we could understand what's going wrong here… maybe the browser-compatible `process` from npm was installed in the root `node_modules` before, but isn't now due to a dependency change?

imjoehaines:
> Slightly disconcerting. Would be ideal if we could understand what's going wrong here… maybe the browser-compatible `process` from npm was installed in the root `node_modules` before, but isn't now due to a dependency change?

There's a `process` in node_modules on next and this branch, so I don't think that's it 😕

PR #1376, opened 2021-04-29
feat(electron): Default user.id to device.id

bengourley:
## Goal

When `user.id` is not set it should default to `device.id`.

## Design

It required more complexity than seems desirable to ensure the `device.id` was populated before sending a session. Any other ideas for workable solutions welcome!

## Changeset

- Updated `electron-session` to receive the filestore param so it can wait for the id to be loaded
- Added `setDefaultUserId()` function in Electron's device plugin which runs during onError and onSession to set `user.id` if it's not already set.

## Testing

Tested manually, with unit tests and end to end tests.

### DIFF
--- a/packages/plugin-electron-session/session.js
+++ b/packages/plugin-electron-session/session.js
@@ -12,9 +12,9 @@ module.exports = (app, BrowserWindow, filestore) => ({
     }
 
     // jump through hoops to ensure the session gets started _after_ we have a device id
-    filestore.getDeviceInfo().then(() => {
-      app.whenReady().then(() => { setTimeout(() => client.startSession(), 10) })
-    }).catch(() => {})
+    Promise.all([filestore.getDeviceInfo(), app.whenReady()])
+      .then(() => { setTimeout(() => client.startSession(), 10) })
+      .catch(() => {})
 
     let lastInBackground = null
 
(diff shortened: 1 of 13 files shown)


packages/plugin-electron-session/session.js
```diff
@@ -11,7 +11,10 @@ module.exports = (app, BrowserWindow) => ({
       return
     }
 
-    app.whenReady().then(() => { client.startSession() })
+    // jump through hoops to ensure the session gets started _after_ we have a device id
+    filestore.getDeviceInfo().then(() => {
+      app.whenReady().then(() => { setTimeout(() => client.startSession(), 10) })
+    }).catch(() => {})
```
imjoehaines, line 17: This might be a bit cleaner looking with `Promise.all`; it might change the timing though

```suggestion
    Promise.all([filestore.getDeviceInfo(), app.whenReady()])
      .then(() => { setTimeout(() => client.startSession(), 10) })
      .catch(() => {})
```

imjoehaines (approved):

packages/plugin-electron-session/session.js
bengourley: Nice, let's give it a go!

PR #1378, opened 2021-04-29
Allow setting appType in renderer config

imjoehaines:
## Goal

Adds `appType` to the `ALLOWED_IN_RENDERER` list and the related Typescript definition


bengourley (commented):
Looks good. Would be nice to see an end to end test, but there's currently no logic for specifying variable renderer config.

Would be good to test that:
1. events sent from main after renderer has loaded have main config `appType`
2. events sent from renderer have `appType` set in renderer config

I think having a precedent/mechanism will be benefitial in testing the other renderer-only config options.

bengourley (approved):

PR #1382, opened 2021-05-04
Fix project root stripping in renderers

imjoehaines:
## Goal

The current project root stripping doesn't quite work in renderers because it leaves a leading path separator e.g. `/root/a/b/c` is stripped to `/a/b/c` when it should be `a/b/c`

I've made two changes in this PR:

1. pre-normalise the default project root path
    This fixes the issue as it allows the renderer to use the normalised path. Renderers wouldn't normally be able to do this as the normalisation requires the `path` module
1. add a path separator to the project root when stripping paths
    This fixes the issue if a non-default project root is used. I think this is always safe unless the project root is a filename, but that doesn't seem like a reasonable use case as it would only work for one path and would leave an empty filename in the stacktrace

### DIFF
--- a/packages/electron/src/client/main.js
+++ b/packages/electron/src/client/main.js
@@ -51,6 +51,13 @@ module.exports = (opts) => {
 
   bugsnag._logger.debug('Loaded! In main process.')
 
+  // Normalise the project root upfront so renderers have a fully resolved path
+  // The renderers can't do this themselves as they cannot access the 'path' module
+  if (bugsnag._config.projectRoot) {
+    const normalizePath = require('@bugsnag/core/lib/path-normalizer')
+    bugsnag._config.projectRoot = normalizePath(bugsnag._config.projectRoot)
+  }
+
   return bugsnag
 }
 


packages/electron/src/config/main.js
```diff
@@ -30,7 +31,7 @@ module.exports.schema = {
     validate: value => typeof value === 'function'
   },
   projectRoot: {
-    defaultValue: () => app.getAppPath(),
+    defaultValue: () => normalizePath(app.getAppPath()),
```
imjoehaines, line 34: I'm not totally sure if this is a good idea — it does fix the problem when testing locally with macOS but maybe doesn't work as well on Windows?

The integration tests still pass so I think it's ok, but the other change does fix the issue so this might just be unnecessary

bengourley: yeah I'm not sure about this for two reasons:

1. it seems unnecessary, seeing as you duplicate the logic to add a trailing slash in the renderer anyway
2. it doesn't normalize the path it it is set in configuration (only the default value)

the only thing you'd miss in the renderer would be resolving any `..`s.

An alternate suggestion: you could call out to `normalizePath` in the `plugin-electron-ipc/lib/config-serializer.js` module?

kattrali: It seems fine if the integration tests are fine, though after a quick check, I think the only context assertions are on custom context. So adding assertions for default context for renderer events would cover everything.

imjoehaines: Is this not covered by the stacktrace assertions?

kattrali: oh yeah! that should cover it.

kattrali (approved):

bengourley (approved):

PR #1389, opened 2021-05-06
Electron: refactor app version fields

bengourley:
- native package version is now the default for `appVersion`
- `codeBundleId` can now be set in renderer config and is included in the payload for renderer events


test/electron/features/renderer-config.feature
```diff
@@ -15,3 +15,19 @@ Feature: Setting config options in renderers
...
+            | events  | 1        |
+        Then the headers of every event request contains:
+            | Bugsnag-API-Key   | 6425093c6530f554a9897d2d7d38e248 |
+            | Content-Type      | application/json                 |
+            | Bugsnag-Integrity | {BODY_SHA1}                      |
+        Then the contents of an event request matches "renderer/config/<property>.json"
+
+        Examples:
+            | property | config                      |
+            | codeBundleId  | { "codeBundleId": "1.0.0-r0123" } |
```
imjoehaines, line 33: I think this can be in the table of the other scenario, rather than duplicating it, i.e.

```cucumber
    Scenario Outline: Setting config options in renderers
        Given I launch an app with configuration:
            | renderer_config | <config> |
        And I click "renderer-notify"
        Then the total requests received by the server matches:
            | events  | 1        |
        Then the headers of every event request contains:
            | Bugsnag-API-Key   | 6425093c6530f554a9897d2d7d38e248 |
            | Content-Type      | application/json                 |
            | Bugsnag-Integrity | {BODY_SHA1}                      |
        Then the contents of an event request matches "renderer/config/<property>.json"
        Examples:
            | property     | config                            |
            | appType      | { "appType": "real great" }       |
            | codeBundleId | { "codeBundleId": "1.0.0-r0123" } |
```

imjoehaines (approved):

tomlongridge:
As I understand it, we are also dropping `bundleVersion` and reporting new a new `metadata.app.CFBundleVersion` instead. Is that planned under a separate PR?

bengourley:
> As I understand it, we are also dropping `bundleVersion`

`bundleVersion` was removed here https://github.com/bugsnag/bugsnag-js/pull/1389/files#diff-b9305f224b4d7de52d1f1b05b2740104d49f53cf977f72be6d57df5c7b37eceeL61

> reporting new a new metadata.app.CFBundleVersion instead. Is that planned under a separate PR

`CFBundleVersion` is what the native_package_version logic uses.

imjoehaines (approved):

PR #1394, opened 2021-05-07
Add CFBundleVersion to app metadata

imjoehaines:
## Goal

This PR adds `CFBundleVersion` to app metadata on macOS. The implementation exists for other platforms in case we want to add similar fields in the future, but they will currently always return `null`

## Testing

I've added support for platform specific matchers so this field can be checked on macOS and should be missing on Windows/Linux

e.g. `{PLATFORM_MACOS:123|PLATFORM_WINDOWS:456|PLATFORM_LINUX:789}` would expect '123' on macOS, '456' on Windows and '789' on Linux

A missing platform indicates that the field should not be present for that platform, e.g. `{PLATFORM_MACOS:123}` would expect '123' on macOS and no value on Windows/Linux

I also used this for `app.type` instead of the existing regex check as it's a bit stricter (previously macOS could have reported "Linux" and the tests would have passed)

### DIFF
--- a/packages/plugin-electron-app/app.js
+++ b/packages/plugin-electron-app/app.js
@@ -62,7 +62,8 @@ module.exports = (NativeClient, process, electronApp, BrowserWindow, NativeApp =
 
     client.addMetadata('app', {
       installedFromStore: getInstalledFromStore(process),
-      name: electronApp.getName()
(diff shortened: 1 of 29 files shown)


kattrali (approved):
nicely done

PR #1399, opened 2021-05-11
Electron: reduce filestore startup latency

bengourley:
Because the async file read to pick up device id happens during the app start up, it loses out to other blocking io and takes almost half a second to return device id. By switching to a sync fs read, it takes 1ms.

Clearly this blocks the process itself, but the file will always be tiny, and Bugsnag can't send any events/sessions until it has generated/retrieved a device id, so is deemed an acceptable trade off.

### DIFF
--- a/packages/electron-filestore/filestore.js
+++ b/packages/electron-filestore/filestore.js
@@ -1,6 +1,6 @@
 const fs = require('fs')
-const { readFileSync } = fs
-const { unlink, readdir, access, mkdir, writeFile } = fs.promises
+const { readFileSync, writeFileSync, mkdirSync } = fs
+const { unlink, readdir, access, mkdir } = fs.promises
 const { F_OK } = fs.constants
 const { dirname, join } = require('path')
 const { getIdentifier, createIdentifier, identifierKey } = require('./lib/minidump-io')
@@ -16,6 +16,7 @@ class FileStore {
       device: join(base, 'device.json'),
       minidumps: join(crashDir, isMac ? 'completed' : 'reports')
     }
+    this._deviceInfoPromise = null
   }
 
   // Create directory layout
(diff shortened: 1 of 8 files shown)


packages/electron-filestore/filestore.js
```diff
@@ -82,24 +84,37 @@ class FileStore {
...
+        }
+        // this intentionally resolves the device field before async'ly persisting device info
+        // – it means the getDeviceInfo() function returns device info as soon as it's available,
+        // rather than having to wait until it's been persisted to disk
+        resolve(device)
+        try {
+          await this.setDeviceInfo(device)
+        } catch (e) {
+          // it's too late to do anything with this error, so just swallow it
+        }
```
bengourley, line 101: Having this logic after the `resolve()` means that the `{ id }` value can be returned/resolved asap. However, if the device repeatedly fails to write to this storage, a different ID will be generated every app run. Is this an acceptable tradeoff?

kattrali: It makes sense. If it can't be written immediately, it might never be written ever, though that computer will always send fully formed events/sessions. There's also nothing we can do about it if it fails to write in any case.

packages/plugin-electron-session/session.js
```diff
@@ -11,9 +11,8 @@ module.exports = (app, BrowserWindow, filestore) => ({
       return
     }
 
-    // jump through hoops to ensure the session gets started _after_ we have a device id
-    Promise.all([filestore.getDeviceInfo(), app.whenReady()])
-      .then(() => { setTimeout(() => client.startSession(), 10) })
+    app.whenReady()
+      .then(() => client.startSession())
```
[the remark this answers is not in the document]
bengourley: Calling `getDeviceInfo()` would incur another unnecessary sync fs call, so I want to avoid that if possible. Since `app.whenReady()` is async it's (virtually) guaranteed to happen after the sync file read.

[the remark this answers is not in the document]
imjoehaines: The "virtually" is making me nervous 😀

bengourley:
Ok, I've pushed a significant update to this.

First off, to address @imjoehaines's well-placed nervousness around the virtual guarantee that the id would be resolved before the app is ready, I refactored `getDeviceInfo()` to return the same promise if called multiple times before the first call had resolved. However, in testing this, it turned out that because the persistence of the ID was done _after_ the promise resolves, two subsequent calls would generate two different IDs.

I measured the impact of making the entire function synchronous for the worst case (read, mkdir, write) which is **6ms**. For the best case (when an app ID has already been generated, i.e. the second launch after Bugsnag has been installed) it is **0ms**. I think this is a very reasonable tradeoff.

It also has the beneficial side-effect of devices where we repeatedly fail to persist an ID now don't include an ID in the payload, meaning the user-stability score can't be skewed by 1 bad device.

packages/electron-filestore/filestore.js
```diff
@@ -16,6 +16,7 @@ class FileStore {
       device: join(base, 'device.json'),
       minidumps: join(crashDir, isMac ? 'completed' : 'reports')
     }
+    this._deviceInfoPromise = null
```
imjoehaines, line 19: This looks unused, I guess it's leftover from when `getDeviceInfo` was async

packages/electron-filestore/filestore.js
```diff
@@ -82,47 +83,43 @@ class FileStore {
...
+    } catch (e) {
+      // either the file could not be read or wasn't valid JSON
+    }
+
+    try {
+      // attempt to create or update the device.json file with
+      // a) the device data we retrieved or
+      // b) a new auto generated id
+      device = this.setDeviceInfo(device)
+      return device
```
imjoehaines:
```suggestion
      return this.setDeviceInfo(device)
```

imjoehaines (approved):

kattrali (approved):

PR #1403, opened 2021-05-12
Add inline script plugin to renderer notifier

imjoehaines:
## Goal

Adds `plugin-inline-script-content` to the Electron renderer notifier. This allows reporting the code of script tags that throw errors

This required a couple of changes to the merging of renderer options:

1. falsy values can now be set, which allows disabling the `trackInlineScripts` option
2. plugins can now define schema in renderers. Previously only the main renderer schema was used when determining which properties to merge from the given config

## Testing

Integration tests have been added that test this works

I also added the ability to assert object keys are missing by using `{TYPE:undefined}`, which allows testing when `trackInlineScripts` is `false`


imjoehaines:
This isn't needed — with some tweaks the surrounding code plugin also works for errors thrown from HTML files

PR #1404, opened 2021-05-13
feat(electron): Set process.type=preload for errors in preload scripts

bengourley:
## Goal

Set `process.type = "preload"` for handled and unhandled errors in preload scripts.

## Design

Handled and unhandled errors are surfaced differently in preload scripts.

#### Unhandled errors
Unhandled errors are surfaced to the main process and are dealt with using the `plugin-electron-preload-error` module. The solution for that, though a little untidy, is pretty straightforward – annotate the Bugsnag `event` object and detect that annotation in the `procinfo` `onerror` callback. Before the the changes in this PR, these events had `process.type = "browser"` because they were captured in the main process.

#### Handled errors
For handled errors, these are dealt with in process and are routed through a client instance with the preload script context. It is hard to reliably determine whether the notifier is executing in the context of a preload script or not. Before this PR these events had `process.type = "renderer"` because they were handled by the renderer notifier. To determine whether the notifier is running in a preload script, the best test I could think of was to detect the presence a global variable that is present in Node, but not in the browser. This works when the script is not bundled, however, when bundled by webpack (or similar), attempting to use the `process` or `require` variables, results in a false positive in the browser because it defines them. To prevent the false positives, I attempted to `eval()` these checks, which works (because webpack doesn't parse the code in eval), but it is prevented from running when a sensible CSP is used. This is why the automated tests fail.

### DIFF
--- a/packages/plugin-electron-preload-error/preload-error.js
+++ b/packages/plugin-electron-preload-error/preload-error.js
@@ -27,6 +27,8 @@ module.exports = app => ({
 
         event.context = context
 
+        event._isPreloadError = true
+
         client._notify(event)
       })
     })
--- a/packages/plugin-electron-preload-error/preload-error.js
+++ b/packages/plugin-electron-preload-error/preload-error.js
@@ -27,7 +27,7 @@ module.exports = app => ({
 
         event.context = context
 
-        event._isPreloadError = true
+        event.addMetadata('process', 'type', 'preload')
 
         client._notify(event)
       })
--- a/packages/plugin-electron-process-info/procinfo.js
+++ b/packages/plugin-electron-process-info/procinfo.js
@@ -14,9 +14,9 @@ module.exports = (source = process) => ({
         info.heapStatistics = source.getHeapStatistics()
       }
 
-      if (isPreload || event._isPreloadError) {
+      if (isPreload) {
         info.type = 'preload'
-      } else if (typeof source.type === 'string') {
+      } else if (event.getMetadata('process', 'type') !== 'preload' && typeof source.type === 'string') {
         // type should always be available
         info.type = source.type
       }
--- a/packages/plugin-electron-process-info/procinfo.js
+++ b/packages/plugin-electron-process-info/procinfo.js
@@ -1,5 +1,9 @@
-// eslint-disable-next-line no-eval
-const isPreload = !!eval('typeof global !== "undefined"') && !!eval('typeof window !== "undefined"')
+const isPreload =
+  // is the process actually node-like (webpack defines a "global" variable when bundling, but node's
+  // global contains a circular reference to itself global.global)
+  typeof global !== 'undefined' && typeof global.global !== 'undefined' && global.global === global &&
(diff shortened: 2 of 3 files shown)


packages/plugin-electron-preload-error/preload-error.js
```diff
@@ -27,6 +27,8 @@ module.exports = app => ({
 
         event.context = context
 
+        event._isPreloadError = true
```
bengourley, line 30: This is untidy. An alternative would be to add an `onerror` callback to the `client._notify()` call below, but this would run _after_ user defined callbacks, so the value wouldn't be correct for those callbacks.

packages/plugin-electron-process-info/procinfo.js
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line no-eval
+const isPreload = !!eval('typeof global !== "undefined"') && !!eval('typeof window !== "undefined"')
```
bengourley, line 2: This works for preload scripts because `global` is not defined in the browser but it is in a Node-like environment. However, when you bundle with webpack, it defines `global` for you, so it also returns `true` in a renderer bundle. The `eval()` solution is a workaround for this, but then it's not compatible with sensible CSP rules.

packages/plugin-electron-process-info/procinfo.js
```diff
@@ -7,7 +10,9 @@ module.exports = (source = process) => ({
         info.heapStatistics = source.getHeapStatistics()
       }
 
-      if (typeof source.type === 'string') {
+      if (isPreload || event._isPreloadError) {
```
bengourley, line 13: Checking the untidy, arbitrary and ad-hoc'ly defined event flag here.

packages/plugin-electron-preload-error/preload-error.js
imjoehaines: Maybe it'd be a bit cleaner if this set the process type directly and the process info plugin avoided overwriting it, e.g.

```js
// plugin-preload-error
event.addMetadata('process', { type: 'preload' })

// plugin-process-info
info.type = event.getMetadata('process', 'type') || 'something else'
```

bengourley: I did something similar to this. It actually runs twice (once in main and once in the renderer) so it needs to be a bit smarter about what it overwrites, but I got it working.

packages/plugin-electron-process-info/procinfo.js
```diff
@@ -1,3 +1,10 @@
+const isPreload =
+  // is the process actually node-like (webpack defines a "global" variable when bundling, but node's
+  // global contains a circular reference to itself global.global)
+  typeof global !== 'undefined' && typeof global.global !== 'undefined' && global.global === global &&
+  // AND is the process browser-like (does the process have a window and a document?)
+  typeof window !== 'undefined' && typeof document !== 'undefined'
+
```
bengourley, line 7: Found a "nice" way of detecting a preload context without eval 😄

imjoehaines (approved):

PR #1405, opened 2021-05-13
Load surrounding code from paths relative the project root

imjoehaines:
## Goal

This allows the plugin to work if the project root has already been stripped before it runs. If there is no project root, or the file path is absolute, there should be no change in behaviour


bengourley (commented):
Looks good. Are you planning to add it to Electron in a separate PR?

imjoehaines:
> Looks good. Are you planning to add it to Electron in a separate PR?

Yes; I'll add some tests for it too

PR #1406, opened 2021-05-17
feat(electron): Add 'Bugsnag loaded' breadcrumb

bengourley:
## Goal

Add a "Bugsnag loaded" breadcrumb to the main process.

### DIFF
--- a/packages/electron/src/client/main.js
+++ b/packages/electron/src/client/main.js
@@ -63,8 +63,8 @@ module.exports = (opts) => {
   bugsnag.markLaunchComplete = markLaunchComplete
 
   bugsnag._logger.debug('Loaded! In main process.')
-  if (bugsnag._config.enabledBreadcrumbTypes && bugsnag._config.enabledBreadcrumbTypes.includes('navigation')) {
-    bugsnag.leaveBreadcrumb('Bugsnag loaded', {}, 'navigation')
+  if (bugsnag._config.enabledBreadcrumbTypes && bugsnag._config.enabledBreadcrumbTypes.includes('state')) {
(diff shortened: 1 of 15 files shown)


packages/electron/src/client/main.js
```diff
@@ -63,6 +63,9 @@ module.exports = (opts) => {
   bugsnag.markLaunchComplete = markLaunchComplete
 
   bugsnag._logger.debug('Loaded! In main process.')
+  if (bugsnag._config.enabledBreadcrumbTypes && bugsnag._config.enabledBreadcrumbTypes.includes('navigation')) {
```
imjoehaines, line 66: Is there value in a loaded breadcrumb in renderers too?

Mostly thinking of cases where a renderer is started while using an app, which could be a long time after the app is opened

packages/electron/src/client/main.js
```diff
@@ -63,6 +63,9 @@ module.exports = (opts) => {
...
+    bugsnag.leaveBreadcrumb('Bugsnag loaded', {}, 'navigation')
```
imjoehaines, line 67: This should have the type `state`

packages/electron/src/client/main.js
bengourley: It was agreed with product that it would be just in the main process.

There is the breadcrumb on the `browser-window-created` event which would give you the same info though?

packages/electron/src/client/main.js
bengourley: Ah ok, in which case the browser notifier needs updating!

imjoehaines (approved):

PR #1413, opened 2021-05-19
feat(react-native): Always upload JS version of bundle file

bengourley:
When Hermes is used, the final main.jsbundle is a precompiled binary format. This means that the Bugsnag backend can't
parse it as JS to find the original (precompiled) method name.

The JS bundle (which is fed into the Hermes compiler) always exists in this location, whether Hermes is in use or not.

### DIFF
--- a/packages/react-native/bugsnag-react-native-xcode.sh
+++ b/packages/react-native/bugsnag-react-native-xcode.sh
@@ -17,7 +17,7 @@ fi
 
 DEST="$CONFIGURATION_BUILD_DIR/$UNLOCALIZED_RESOURCES_FOLDER_PATH"
 
-BUNDLE_FILE="$DEST/main.jsbundle"
+BUNDLE_FILE="$CONFIGURATION_BUILD_DIR/main.jsbundle"
 if [ ! -f "$BUNDLE_FILE" ]; then
   echo "Skipping source map upload because app has not been bundled."
   exit 0
--- a/packages/react-native/bugsnag-react-native-xcode.sh
+++ b/packages/react-native/bugsnag-react-native-xcode.sh
@@ -15,8 +15,6 @@ if [ -z "$API_KEY" ]; then
   exit 1
 fi
 
-DEST="$CONFIGURATION_BUILD_DIR/$UNLOCALIZED_RESOURCES_FOLDER_PATH"
-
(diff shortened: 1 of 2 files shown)


packages/react-native/bugsnag-react-native-xcode.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 DEST="$CONFIGURATION_BUILD_DIR/$UNLOCALIZED_RESOURCES_FOLDER_PATH"
```
nickdowell, line 18: DEST doesn't appear to be used any more

packages/react-native/bugsnag-react-native-xcode.sh
```diff
@@ -17,7 +17,7 @@ fi
...
 
-BUNDLE_FILE="$DEST/main.jsbundle"
+BUNDLE_FILE="$CONFIGURATION_BUILD_DIR/main.jsbundle"
```
nickdowell, line 20: So this path works for non-Hermes React Native too?

bengourley: Yeah, I checked it.

For both Hermes and non-Hermes, the RN Xcode build script sets its own `BUNDLE_FILE` variable [here](https://github.com/facebook/react-native/blob/dc80b2dcb52fadec6a573a9dd1824393f8c29fdc/scripts/react-native-xcode.sh#L118).

For non-Hermes, it then [copies the `BUNDLE_FILE` into `$DEST/main.jsbundle`](https://github.com/facebook/react-native/blob/dc80b2dcb52fadec6a573a9dd1824393f8c29fdc/scripts/react-native-xcode.sh#L163-L164).

For Hermes, it [compiles the JS bundle into an .hbc (Hermes byte-code) bundle directly in the `$DEST/main.jsbundle` location](https://github.com/facebook/react-native/blob/dc80b2dcb52fadec6a573a9dd1824393f8c29fdc/scripts/react-native-xcode.sh#L175).

Then it [bundles the JS](https://github.com/facebook/react-native/blob/dc80b2dcb52fadec6a573a9dd1824393f8c29fdc/scripts/react-native-xcode.sh#L157) into that location.

Hope that makes sense!

nickdowell (approved):

PR #1415, opened 2021-05-20
[PLAT-5961] feat(browser,node): Add default appType config

bengourley:
## Goal

App type is given visual prominence in the dashboard, so populate it with reasonable default values for browser and Node.js events.

### DIFF
--- a/packages/node/src/config.js
+++ b/packages/node/src/config.js
@@ -4,6 +4,10 @@ const os = require('os')
 const { inspect } = require('util')
 
 module.exports = {
+  appType: {
+    ...schema.appType,
+    defaultValue: () => 'node'
+  },
   projectRoot: {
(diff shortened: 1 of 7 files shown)


packages/node/src/config.js
```diff
@@ -4,6 +4,10 @@ const os = require('os')
 const { inspect } = require('util')
 
 module.exports = {
+  appType: {
+    ...schema.appType,
+    defaultValue: () => 'node'
```
imjoehaines, line 9: Should we update the Node MazeRunner tests to check for this too?

imjoehaines (approved):

packages/node/src/config.js
bengourley: I did… but it turns out I didn't add it to the commit :)

PR #1418, opened 2021-05-25
Change console-breadcrumbs.js - support  lower case and other dev env…

idanlevi1:
Without this change, we get `console-breadcrumbs.js` in the source (when I in DEVELOPMENT or 'test' environment)

## Goal

- Support lower case (for example 'DEVELOPMENT')
- Support other dev environments (for example - 'stage', 'test'...)

## Changeset

Check `isDev` with not production environment 

## Testing

manual with react native debugger

@imjoehaines @bengourley

### DIFF
--- a/packages/plugin-console-breadcrumbs/console-breadcrumbs.js
+++ b/packages/plugin-console-breadcrumbs/console-breadcrumbs.js
@@ -7,7 +7,7 @@ const includes = require('@bugsnag/core/lib/es-utils/includes')
  * Leaves breadcrumbs when console log methods are called
  */
 exports.load = (client) => {
-  const isDev = /^dev(elopment)?$/.test(client._config.releaseStage)
+  const isDev = !/^prod(uction)?$/.test(client._config.releaseStage.toLowerCase())
 
   if (!client._config.enabledBreadcrumbTypes || !includes(client._config.enabledBreadcrumbTypes, 'log') || isDev) return
 


idanlevi1 (commented):
done

mattdyoung:
Hi @idanlevi1 

If you want to disable log breadcrumbs for console messages on other release stages this can be done by excluding `log` from `enabledBreadcrumbTypes` in your Bugsnag configuration, which gives you full control over the console breadcrumbs without this PR:
https://docs.bugsnag.com/platforms/javascript/configuration-options/#enabledbreadcrumbtypes

The reason for us disabling these by default in development environments is that wrapping console methods to leave breadcrumbs has the side effect of messing with line numbers in log messages which can make debugging trickier. In general console breadcrumbs are useful in other non-production environments so we wouldn't want to disable them for other release stage names e.g. `staging`, `testflight` etc.

idanlevi1:
> Hi @idanlevi1
> 
> If you want to disable log breadcrumbs for console messages on other release stages this can be done by excluding `log` from `enabledBreadcrumbTypes` in your Bugsnag configuration, which gives you full control over the console breadcrumbs without this PR:
> https://docs.bugsnag.com/platforms/javascript/configuration-options/#enabledbreadcrumbtypes
> 
> The reason for us disabling these by default in development environments is that wrapping console methods to leave breadcrumbs has the side effect of messing with line numbers in log messages which can make debugging trickier. In general console breadcrumbs are useful in other non-production environments so we wouldn't want to disable them for other release stage names e.g. `staging`, `testflight` etc.

I don't want to disabled console logs, I want to see the source of logs in the debugger, without my change (in other stages/environment except for development - 'testing' for example) i see like that:

image

mattdyoung:
@idanlevi1 
Excluding `log` from `enabledBreadcrumbTypes` in your Bugsnag configuration doesn't disable console logs, it disables capturing those logs as breadcrumbs shown on errors in the Bugsnag dashboard.

In order to capture those breadcrumbs Bugsnag needs to wrap console methods which can interfere with the console log output.

Can you try setting `enabledBreadcrumbTypes` to exclude `logs` in your Bugsnag configuration for your test environment and confirm whether that solves the issue e.g.
```
Bugsnag.start({ 
  enabledBreadcrumbTypes: ['error', 'navigation', 'request', 'user']
})
```

yousif-bugsnag:
Closing this as we have not had a response, but if the previous advice does not resolve the issue let us know and we can take another look.

Issue #1311, opened 2021-02-26
React Native async error messages are swallowed during development


AndrewMorsillo:
### Describe the bug

During development async error messages don't get output to the react native console. Only a warning about an unhandled promise rejection is output with a bugsnag internal stack trace. 

The actual error text is sent to bugsnag but not output to the console which makes development difficult. 

There was another issue about this which got closed but unfortunately the bug wasn't fixed https://github.com/bugsnag/bugsnag-js/issues/1135

### Steps to reproduce
1. Throw an async error in a react native app in development mode
2. Observe that the error is swallowed. It is hidden from the log output and is only viewable on bugsnag.

### Environment
* Bugsnag version: "@bugsnag/react-native": "^7.7.0"
* React native version: 0.63.4
* Device (e.g. iphonex): any

johnkiely1:
Hi @AndrewMorsillo, 

In the situation of an async error, even without Bugsnag enabled, the console would only display a warning. The fix for #1135 allows that warning to now be displayed in console as, with Bugsnag enabled, it was hiding it previously. So to me this would sound as expected behaviour. If I've misunderstood the problem you are suggesting do let me know and perhaps you could add some screenshots to demonstrate what you are seeing vs what you expect, thanks.

AndrewMorsillo:
Hi @johnkiely1, here's an example. When an unhandled async exception is grabbed by bugsnag in dev only the warning and stack is printed in the console like so:

![image](https://user-images.githubusercontent.com/858965/109703666-bc88e900-7b63-11eb-8138-e3e9d558d320.png)

but on Bugsnag we get the useful part, the message of the exception that was thrown:

![image](https://user-images.githubusercontent.com/858965/109703847-f2c66880-7b63-11eb-9731-c646fac0c967.png)

Bugsnag should not hide the exception message from the react native console.

AndrewMorsillo:
Another less important issue is that each of those warning lines that are logged by bugsnag appear as a react-native "yellowbox" development error popup. This means you have to click through to close ~20+ bugsnag warning lines to get to any other errors that might have been shown by react native's yellowbox before the bugsnag warnings were logged.

yousif-bugsnag:
Hey @AndrewMorsillo, I'm still looking into the separate warning messages, but I haven't been able to reproduce your original issue.

Using React native 0.63.4 and @bugsnag/react-native 7.7.0, the error message is output to the console as expected: 
[...]

PR #1419, opened 2021-05-25
fix(react-native): Copy RN's internal logging for unhandled rejections

bengourley:
## Goal

Prevent duplicate log lines for promise rejection warnings in development (as observed in https://github.com/bugsnag/bugsnag-js/issues/1311#issuecomment-796917603).

## Design

The previous implementation copied logic from the `promise` module – however, in development, RN used a custom `onUnhandled` callback which logged out once, rather than once per stackframe. This solution copies the RN internal logic, instead of the `promise` library default.


lemnik (approved):
LGTM

PR #1420, opened 2021-05-26
deps(react-native): Update bugsnag-cocoa to v6.9.3

nickdowell:


nickdowell:
Closing because v6.9.4 is ready for integration now

PR #1422, opened 2021-05-29
perf(angular): do not run change detections when notifying the event

arturovt:
Angular runs 10 change detection cycles if the `notify()` method is called within the Angular zone. This drastically affects the performance and testing (since Angular is waiting for all tasks to be completed when running tests).

This PR adds an ability to avoid running unwanted change detections when notifying the event.


xander-jones:
Hey @arturovt thanks for the PR, we'll review as priorities allow :)

PR #1425, opened 2021-06-02
Fix RN 0.63 navigation CI

bengourley:
A recent version of `react-native-navigation` introduced some Kotlin code that wasn't compatible with the version of the Kotlin compiler we were using. This bumps the version of the compiler on the fixture app and moves along the minimum version of RNN required (to ensure a cached version of e.g. 7.0.0 was installed, giving me a false positive).


twometresteve (approved):
LGTM

PR #1426, opened 2021-06-02
fix(react-native): Only include codeBundleId in payload when it has a value

bengourley:
Makes the behaviour of this field consistent with Electron in that it only appears in the payload if it has a value.


lemnik (approved):
LGTM

PR #1435, opened 2021-06-10
Update client.d.ts

theahmadzai:
Add optional _client instance type

## Goal

We would do something like this before 

```javascript 
type WithInstance = Client & { _client?: Client }

if (!(Bugsnag as WithInstance)?._client) {
  Bugsnag.start({
    apiKey: BUGSNAG_API_KEY,
  })
}
```

## Design

nil

## Changeset

added missing _client optional property to type Client to check if the instance is already initialized or not

## Testing

nil

### DIFF
--- a/packages/core/types/client.d.ts
+++ b/packages/core/types/client.d.ts
@@ -74,4 +74,4 @@ declare class Client {
   public Session: typeof Session;
 }
 
-export default Client
+export default Client & { _client?: Client }


yousif-bugsnag:
Hi @theahmadzai 

Unfortunately we wouldn't accept this PR as we don't recommend using Bugsnag in this way.

Our recommendation would always be to load Bugsnag once as early in the app lifecycle as possible, and ideally before any other scripts in your app, so that you don't lose visibility of errors from your other scripts or those that occur during startup. 

This kind of lazy loading also means you'll potentially be missing the breadcrumbs that Bugsnag logs leading up to the error event, as well as missing session information that is used to calculate stability scores.

theahmadzai:
I'm using this with nextjs and it shows annoying logs again and gain that bugsnag start has already been called.

xander-jones:
How are you instantiating Bugsnag in your application? Is `Bugsnag.start()` being called more than once? Even with Next.js, it should be possible to instantiate Bugsnag server-side for the render (as an server-side Node client), and then instantiate browser Bugsnag on the client-side page without the two interfering.

theahmadzai:
I'm using file-based API routes so I'm instantiating it in a file then using the notify method on the on-error middleware for API and I'm using it with react-plugin for client-side, I guess it just says this in nextjs development.

Thanks for your replies.

PR #1437, opened 2021-06-10
Resolve failures in e2e tests for ejected expo apps

twometresteve:
## Goal

Resolves a few issues that were preventing the e2e tests from running properly.

NOTE: Although this is ready for review as-is, we will wait for PLAT-6764 to be fixed before merging it to avoid still having an incomplete test run.  

## Changeset

- Android fixture - allow HTTP traffic
- iOS fixed - include crashy native modules in the build

I have also upgraded the RN tests to use MazeRunner v5 to allow newer features to be used - but this is an incidental change.

## Testing

Will be fully covered by CI once complete.


Cawllec (approved):
LGTM, and tests are responding correctly

PR #1438, opened 2021-06-14
fix(react-native): Ensure source maps are uploaded on iOS RN<0.64

bengourley:
An update that fixed RN source map uploads for Hermes on iOS (v0.64+) inadvertently broke uploads for RN <0.64 on iOS due to the combination of the JS bundle location changing, and some false +ves in the end to end tests.

This fix looks in both possible locations before skipping the upload.


twometresteve (approved):
LTGM - just a question over testing for RN 0.64, but we'll add further checks to CI for future.

PR #1440, opened 2021-06-15
refactor(plugin-react): Make plugin Yarn PnP compatible

bengourley:
Currently, Yarn PnP mode is not supported by `@bugsnag/plugin-react`'s types.

This change is a draft for PnP + TypeScript support by breaking the hard dependency on `@bugsnag/core`'s types, and constructing the required types explicitly within the plugin. This dependency was previously expressed as a "peerDependency" because it was not a requirement for non-TypeScript users. However, with the Yarn PnP system, despite `@bugsnag/core` being installed elsewhere – it is not visible from `@bugsnag/plugin-react`.

### DIFF
--- a/packages/plugin-react/types/bugsnag-plugin-react.d.ts
+++ b/packages/plugin-react/types/bugsnag-plugin-react.d.ts
@@ -1,11 +1,22 @@
-import { Plugin, OnErrorCallback } from '@bugsnag/core'
 import React from 'react'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-interface BugsnagPluginReact extends Plugin { }
-// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+type OnErrorCallback = (event: any, cb: (err: null | Error, shouldSend?: boolean) => void) => void | boolean | Promise<void | boolean>
+type NotifiableError = Error
+| { errorClass: string, errorMessage: string }
+| { name: string, message: string }
+| string
+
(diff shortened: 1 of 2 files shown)


packages/plugin-react/types/bugsnag-plugin-react.d.ts
```diff
@@ -1,11 +1,22 @@
-import { Plugin, OnErrorCallback } from '@bugsnag/core'
 import React from 'react'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-interface BugsnagPluginReact extends Plugin { }
-// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+type OnErrorCallback = (event: any, cb: (err: null | Error, shouldSend?: boolean) => void) => void | boolean | Promise<void | boolean>
```
bengourley, line 3: Bit of a rabbit hole here – `OnErrorCallback` needs `@bugsnag/core/event.d.ts` which pulls in virtually everything else from core. Typed as any, but this means `onError` callbacks added on the ErrorBoundary are awkward to write.

imjoehaines: Yeah, this isn't great 😕 Maybe this approach isn't worth doing then? I imagine we'll hit similar issues in other plugins as well

The one saving grace is `noImplicitAny` seems to paper over this problem because you have to type the `event` parameter manually, both before & after this change

PR #1443, opened 2021-06-16
fix(plugin-network-breadcrumbs): Strip query string before checking if url should be ignored

bengourley:
The `NetInfo` package inside Expo makes some requests when it checks for connectivity which should be ignored. However, the module adds a cachebusting querysting value to the URL which means it doesn't exactly match the entry in the list of ignored URLs.

This update strips any querystring before checking whether the URL should be ignored.


lemnik (approved):
LGTM

PR #1444, opened 2021-06-22
feat(examples): Ensure all example apps build and run

bengourley:
- Some examples had lockfiles which were generated against a private repo, so all were regenerated
- The Expo example used the incorrect API for getting the current screen name
- The Nuxt example has some code that didn't work in the config file, plus the node version listed in the Dockerfile was too old to work with some of Nuxt's syntax


imjoehaines (commented):
Should we include the lock files at all? It might be easier to add them to gitignore instead

bengourley:
> Should we include the lock files at all? It might be easier to add them to gitignore instead

I think it helps for when dependencies that say they do semver don't actually follow it. For example I'm pretty sure the changes I made to Expo were probably because I got a newer version of the navigation module that satisfied the `package.json` range.

The lockfiles mean that the examples are more likely to work. If you wanted to remove them I'd say the `package.json`s should all be updated to be more strict.

imjoehaines (approved):

PR #1445, opened 2021-06-22
Suggest Bugsnag plugin v7 if AGPv7 is used

lemnik:
## Goal
When adding Bugsnag to an existing ReactNative project using the `react-native-cli` we should suggest `bugsnag-android:7+` if the ReactNative project is using the Android Gradle Plugin 7.

## Design
Before prompting the user to enter a version:
 - Read the project-level `build.gradle`
 - Attempt to detect the Android Build Plugin used
 - If successfully found - and the version number starts with `7.` suggest `7+`
 - Otherwise fallback to suggesting `5+`

## Testing
Included test coverage for the AGP version-number detection. These changes can also be manually tested by using the `cli` with ReactNative projects and adjusting the AGP version number.

### DIFF
--- a/packages/react-native-cli/src/commands/InstallCommand.ts
+++ b/packages/react-native-cli/src/commands/InstallCommand.ts
@@ -3,7 +3,7 @@ import logger from '../Logger'
 import { install as npmInstall, detectInstalled, guessPackageManager } from '../lib/Npm'
 import { install as podInstall } from '../lib/Pod'
 import onCancel from '../lib/OnCancel'
-import { findAndroidPluginVersion, modifyRootBuildGradle, modifyAppBuildGradle } from '../lib/Gradle'
+import { getSuggestedBugsnagGradleVersion, modifyRootBuildGradle, modifyAppBuildGradle } from '../lib/Gradle'
 
 export default async function run (argv: string[], projectRoot: string, opts: Record<string, unknown>): Promise<boolean> {
   try {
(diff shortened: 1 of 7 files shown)


packages/react-native-cli/src/commands/InstallCommand.ts
```diff
@@ -44,11 +44,14 @@ async function installJavaScriptPackage (projectRoot: string): Promise<void> {
 async function addGradlePluginDependency (projectRoot: string): Promise<void> {
   logger.info('Adding the Bugsnag Android Gradle Plugin')
 
+  const androidPluginVersion = await findAndroidPluginVersion(projectRoot)
+  const suggestedPluginVersion = androidPluginVersion?.startsWith('7.') ? '7+' : '5+'
```
bengourley, line 48: what happens when AGP gets to 8? I wonder if we can future proof this a little?

lemnik: We could emit a suggestion for `8+` in that case, but then we could (in theory) also start suggesting versions before they are released.

fractalwrench (approved):

packages/react-native-cli/src/commands/InstallCommand.ts
tomlongridge: How about going the other way and only suggesting a version if we positively match it? i.e. `5+` if the version is <7, `7+` if the version =7, and nothing otherwise. We could print a suitable warning to say that we couldn't find an AGP version that we support and link to the docs?

(The message below needs updating either way.)

bengourley (approved):

PR #1450, opened 2021-06-28
Dockerise release process

bengourley:
## Goal

The current release process happens on developer machines, which while usually are in a "good state" to publish a release, not guaranteed to be so. The goal of dockerising the release process is to provide a clean machine state to provide greater confidence in the process.

## Design

The Dockerfile builds a machine with all the required system level dependencies and system level state with which to deploy. Some things to note here are:

- `git` – we need git to a) clone the repo and b) push the commit/tag created by the lerna release process
- `curl` – required by the android vendoring script for react native
- a non-root user – npm won't run post install scripts for dependencies as root

Then at runtime in the container, the script in `./bin/release.sh` executes the process:
- cloning and checking out the repo with the desired branch
- installing and linking depedencies
- running the lerna release process
- uploading assets to the cdn

The container mounts the `.npmrc` and `.gitconfig` from the host's home directory.

### Usage

Build the release container: `docker-compose build release`.

Make a release:

```sh
GITHUB_USER=youruser \
GITHUB_ACCESS_TOKEN=<generate a personal access token> \
AWS_ACCESS_KEY_ID=123 \
AWS_SECRET_ACCESS_KEY=123 \
RELEASE_BRANCH=master \
VERSION=patch \
  docker-compose run release
```

Make a prerelease

```sh
GITHUB_USER=youruser \
GITHUB_ACCESS_TOKEN=<generate a personal access token> \
AWS_ACCESS_KEY_ID=123 \
AWS_SECRET_ACCESS_KEY=123 \
RELEASE_BRANCH=myfeaturebranch \
VERSION=prepatch \
  docker-compose run release
```

## Testing

I tested this manually with a prerelease.


bin/release.sh
```diff
@@ -0,0 +1,45 @@
...
+npm ci
+npm run bootstrap -- --ci
+
+prerelease () {
+  npx lerna publish $VERSION --dist-tag next
+  npm run cdn-upload
+}
+
+release () {
+  echo "npx lerna publish $VERSION"
```
bengourley, line 33: todo: to prevent doing a release by accident while testing, this command is printed to the console rather than executed

docker-compose.yml
```diff
@@ -208,6 +208,21 @@ services:
...
+      dockerfile: dockerfiles/Dockerfile.release
+    environment:
+      GITHUB_USER:
+      GITHUB_ACCESS_TOKEN:
+      RELEASE_BRANCH:
+      AWS_ACCESS_KEY_ID:
+      AWS_SECRET_ACCESS_KEY:
+      VERSION:
+    volumes:
+      - ~/.gitconfig:/home/releaser/.gitconfig
```
imjoehaines, line 223: I don't think this will work for me as my `.gitconfig` looks like this:

```
[include]
    path = ~/dotfiles/.gitconfig
```

Can we use the output of `git config user.name` and `git config user.email` instead (somehow)? I'd be OK with extra environment variables for `GIT_USER`/`GIT_EMAIL` as well

imjoehaines (commented):
Some initial comments but I'll take a better look tomorrow

docker-compose.yml
imjoehaines: Actually it seems like Git is OK with paths that don't exist and duplicate config sections, so I can just put my user info in `~/.gitignore`

bin/release.sh
```diff
@@ -0,0 +1,45 @@
...
+  echo "npm run cdn-upload"
```
imjoehaines, line 34: Is there any danger in running this unnecessarily? e.g. if there are only React Native changes in the release

Maybe we can get lerna to tell us if the browser package changed before running this?

docker-compose.yml
bengourley: Damn, I took a chance on that as it simplifies things a lot. It will have to be env variables, as you can't use the output of a command on the host machine as input for docker/compose.

bin/release.sh
bengourley: A little danger I guess.

https://github.com/lerna/lerna/tree/main/commands/list#--parseable

docker-compose.yml
imjoehaines: I tested this morning and I think it's OK as-is — I can just put my user info in `~/.gitconfig` and git is happy even when the included file doesn't exist

imjoehaines (commented):
Looks good! Left a couple more comments from testing it out (without publishing) but nothing major

The release steps in the contributing guide could also use an update

bin/release.sh
```diff
@@ -0,0 +1,45 @@
...
+  exit 1
+}
+
+# Ensure all required variables are set before doing any work
+if [[ -z $GITHUB_USER ]]; then error_missing_field "GITHUB_USER"; fi
+if [[ -z $GITHUB_ACCESS_TOKEN ]]; then error_missing_field "GITHUB_ACCESS_TOKEN"; fi
+if [[ -z $RELEASE_BRANCH ]]; then error_missing_field "RELEASE_BRANCH"; fi
+if [[ -z $VERSION ]]; then error_missing_field "VERSION"; fi
+if [[ -z $AWS_ACCESS_KEY_ID ]]; then error_missing_field "AWS_ACCESS_KEY_ID"; fi
+if [[ -z $AWS_SECRET_ACCESS_KEY ]]; then error_missing_field "AWS_SECRET_ACCESS_KEY"; fi
```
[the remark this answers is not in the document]
bengourley: ahh yeah I was using `-v` which is compatible with `set -u` – but I think is only supported on newer versions of bash.

CONTRIBUTING.md
```diff
@@ -115,28 +123,30 @@ Finally:
...
 
 ```
-lerna publish from-git --dist-tag next
+GITHUB_USER=<your github username> \
+GITHUB_ACCESS_TOKEN=<generate a personal access token> \
+AWS_ACCESS_KEY_ID=xxx \
+AWS_SECRET_ACCESS_KEY=xxx \
+RELEASE_BRANCH=master \
+VERSION=patch \
+  docker-compose run preminor
```
imjoehaines, line 147:
```suggestion
VERSION=preminor \
  docker-compose run
```

imjoehaines (approved):

bin/release.sh
```diff
@@ -0,0 +1,43 @@
...
+  https://"$GITHUB_USER":"$GITHUB_ACCESS_TOKEN"@github.com/bugsnag/bugsnag-js.git
+
+cd /app/bugsnag-js
+
+# "ci" rather than "install" ensures the process doesn't make the work tree dirty by modifying lockfiles
+npm ci
+npm run bootstrap -- --ci
+
+# check if the browser package changed – if it didn't we don't need to upload to the CDN
+BROWSER_PACKAGE_CHANGED=`npx lerna changed --parseable | grep -c packages/js$`
```
imjoehaines, line 29:
```suggestion
BROWSER_PACKAGE_CHANGED=$(npx lerna changed --parseable | grep -c packages/js$)
```

bin/release.sh
```diff
@@ -0,0 +1,43 @@
...
+
+case $VERSION in
+  "prerelease" | "prepatch" | "preminor" | "premajor")
+    npx lerna publish "$VERSION" --dist-tag next
+    ;;
+
+  *)
+    npx lerna publish $VERSION
```
imjoehaines, line 37:
```suggestion
    npx lerna publish "$VERSION"
```

bin/release.sh
```diff
@@ -0,0 +1,43 @@
...
+    ;;
+esac
+
+if [ $BROWSER_PACKAGE_CHANGED -eq 1 ]; then
```
imjoehaines, line 41:
```suggestion
if [ "$BROWSER_PACKAGE_CHANGED" -eq 1 ]; then
```

imjoehaines (commented):
Some more shellcheck errors from the new changes

djskinner (approved):

PR #1464, opened 2021-07-07
[PLAT-6174] feat: Ensure server plugins honour autoDetectErrors option

bengourley:
The server plugins did not honour the `autoDetectErrors` configuration option.


test/node/features/express_disabled.feature
```diff
@@ -0,0 +1,62 @@
...
+Background:
+  Given I store the api key in the environment variable "BUGSNAG_API_KEY"
+  And I store the notify endpoint in the environment variable "BUGSNAG_NOTIFY_ENDPOINT"
+  And I store the sessions endpoint in the environment variable "BUGSNAG_SESSIONS_ENDPOINT"
+  And I start the service "express-disabled"
+  And I wait for the host "express-disabled" to open port "80"
+
+Scenario: a synchronous thrown error in a route
+  Then I open the URL "http://express-disabled/sync"
+  And I wait for 5 seconds
```
imjoehaines, line 12: I don't think you need to wait explicitly anymore; MR has a 30 second wait to check it doesn't get any requests already — https://github.com/bugsnag/maze-runner/blob/v4.11.3/lib/maze/configuration.rb#L9

test/node/features/express_disabled.feature
```diff
@@ -0,0 +1,62 @@
...
+  And I should receive no errors
+
+Scenario: an asynchronous thrown error in a route
+  Then I open the URL "http://express-disabled/sync"
```
imjoehaines, line 16: Should this be `async`?

```suggestion
  Then I open the URL "http://express-disabled/async"
```

test/node/features/express_disabled.feature
bengourley: ah cool, I thought they were a bit slow!

imjoehaines (approved):

PR #1465, opened 2021-07-08
Deliver minidumps to configured endpoint

lemnik:
## Goal
Upload minidump files (produced during native crashes) to the new `minidump` endpoint.

## Design
The primary minidump delivery function has been added to the existing Electron delivery module, while a new `plugin-electron-deliver-minidumps` plugin configures the native crashReporter and schedules delivery of the minidumps.

## Testing
Basic unit tests added, with MazeRunner tests to be added in PLAT-6117

### DIFF
--- a/packages/delivery-electron/delivery.js
+++ b/packages/delivery-electron/delivery.js
@@ -20,14 +20,16 @@ const delivery = (client, filestore, net, app) => {
 
     try {
       req.write(body)
-      req.end()
     } catch (err) {
       // if we can't write this body to the request, it's likely impossible to
       // ever send it successfully
       err.isRetryable = false
 
       cb(err)
+      return
     }
+
+    req.end()
   }
 
   const logError = e => client._logger.error('Error delivering payload', e)
--- a/packages/delivery-electron/delivery.js
+++ b/packages/delivery-electron/delivery.js
@@ -1,8 +1,12 @@
 const { createHash } = require('crypto')
+const { createReadStream } = require('fs')
+const { basename } = require('path')
+const { createGzip } = require('zlib')
 const payload = require('@bugsnag/core/lib/json-payload')
 const PayloadQueue = require('./queue')
 const PayloadDeliveryLoop = require('./payload-loop')
 const NetworkStatus = require('./network-status')
+const FormData = require('form-data')
 
 const delivery = (client, filestore, net, app) => {
   const send = (opts, body, cb) => {
@@ -19,17 +23,19 @@ const delivery = (client, filestore, net, app) => {
     req.on('error', cb)
 
     try {
-      req.write(body)
+      if (typeof body === 'function') {
+        body(req)
+      } else {
+        req.write(body)
+        req.end()
+      }
     } catch (err) {
       // if we can't write this body to the request, it's likely impossible to
       // ever send it successfully
       err.isRetryable = false
 
       cb(err)
-      return
     }
-
-    req.end()
   }
 
   const logError = e => client._logger.error('Error delivering payload', e)
@@ -122,6 +128,39 @@ const delivery = (client, filestore, net, app) => {
         })
       } catch (e) {
         onerror(e, { opts, body }, 'session', cb)
+      }
+    },
+
+    sendMinidump: (minidumpPath, event, cb = () => {}) => {
+      try {
+        const url = new URL(client._config.endpoints.minidumps)
+        url.searchParams.set('api_key', client._config.apiKey)
+
+        const minidumpStream = createReadStream(minidumpPath).pipe(createGzip())
+
+        const formData = new FormData()
+        formData.append('upload_file_minidump', minidumpStream, {
+          filename: basename(minidumpPath)
+        })
+
+        if (event) {
+          const eventBody = payload.event(event, client._config.redactedKeys)
+          formData.append('event', eventBody)
+        }
+
+        const opts = {
+          url: url.toString(),
+          method: 'POST',
+          headers: {
+            'Bugsnag-Sent-At': (new Date()).toISOString(),
+            ...formData.getHeaders()
+          }
+        }
+
+        send(opts, req => formData.pipe(req), cb)
+      } catch (err) {
+        // ensure cb is always called asynchronously to avoid side-effect surprises
+        process.nextTick(() => cb(err))
       }
     }
   }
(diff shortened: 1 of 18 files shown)


packages/delivery-electron/delivery.js
```diff
@@ -1,8 +1,12 @@
 const { createHash } = require('crypto')
+const { createReadStream } = require('fs')
+const { basename } = require('path')
+const { createGzip } = require('zlib')
 const payload = require('@bugsnag/core/lib/json-payload')
 const PayloadQueue = require('./queue')
 const PayloadDeliveryLoop = require('./payload-loop')
 const NetworkStatus = require('./network-status')
+const FormData = require('form-data')
```
imjoehaines, line 9: I guess this is available because Electron uses it internally?

Should we add this as a dependency of this module so we're not necessarily reliant on the version Electron uses? e.g. if a major version comes out with a breaking change and Electron updates to use it

packages/plugin-electron-deliver-minidumps/send-minidump.js
```diff
@@ -0,0 +1,77 @@
...
+        // ever send it successfully
+        err.isRetryable = false
+        reject(err)
+      }
+    })
+  }
+
+  const sendMinidump = async (minidumpPath, event) => {
+    const url = new URL(client._config.endpoints.minidumps)
+    url.searchParams.set('api_key', client._config.apiKey)
```
imjoehaines, line 35: I don't see anything mentioning this in the design but an event can have a different API key to the config's API key. This lets you send different events to different Bugsnag projects, e.g. if parts of the code are owned by different teams

Should we be using the event API key instead of the config API key if there is an event?

imjoehaines: Raised this in Slack so don't worry about changing anything yet

packages/plugin-electron-deliver-minidumps/minidump-loop.js
```diff
@@ -0,0 +1,86 @@
...
+    const shouldSendMinidump = event && await this._onSend(event)
+
+    if (shouldSendMinidump === false) {
+      this._minidumpQueue.remove(minidump)
+      this._scheduleSelf()
+    } else {
+      try {
+        await this._sendMinidump(minidump.minidumpPath, event)
+      } catch (e) {
+        this._onerror(e, minidump)
```
imjoehaines, line 44: If `e.isRetryable === true` should we be retrying the delivery here? IIUC we're only ever attempting to deliver a minidump once

packages/plugin-electron-deliver-minidumps/test/send-minidump.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(net.request).toBeCalledTimes(1)
+
+    const { url, method, headers } = net.request.mock.calls[0][0]
+    expect(url).toMatch(/\?api_key=test-api-key/)
+    expect(method).toBe('POST')
+    expect(headers['content-type']).toMatch(/^multipart\/form-data/)
+    expect(headers['Bugsnag-Sent-At']).toMatch(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/)
+  })
+
+  it('marks server error as retry', async () => {
```
imjoehaines, line 48: It'd be good to add a test where `isRetryable` is false too

packages/plugin-electron-deliver-minidumps/send-minidump.js
imjoehaines: > I believe that the API key in the query parameter would take precedence over the one in the payload so they could probably be different (if it is a pain to get them the same) but ideally both would match

```suggestion
    url.searchParams.set('api_key', event.apiKey || this._config.apiKey)
```

Would be good to add a test for this case as well

imjoehaines: Will be done separately

packages/plugin-electron-deliver-minidumps/minidump-loop.js
```diff
@@ -0,0 +1,93 @@
...
+
+        // if we had a successful delivery - remove the minidump from the queue, and schedule the next
+        this._minidumpQueue.remove(minidump)
+      } catch (e) {
+        // if the error is not retryable: remove the minidump from the queue and signal the error
+        if (!e.isRetryable) {
+          this._minidumpQueue.remove(minidump)
+          this._onerror(e, minidump)
+        }
+        // if the error is retryable: leave things as-is and we will attempt a redelivery
```
imjoehaines, line 52: Is there value in always calling `_onerror`? It could help surface issues assuming we aren't expecting it to fail

`_onerror` also already calls `_minidumpQueue.remove` so this doesn't need to be in both places

```suggestion
        this._onerror(e, minidump)
```

packages/plugin-electron-deliver-minidumps/test/send-minidump.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    const minidumpFile = join(minidumpsPath, 'test-minidump.dmp')
+    await writeFile(minidumpFile, '{}', 'utf8')
+
+    const { sendMinidump } = sendMinidumpFactory(net, client)
+    await expect(sendMinidump(minidumpFile, null)).rejects.toHaveProperty('isRetryable', true)
+  })
+
+  it('marks server error as no-retry', async () => {
```
imjoehaines, line 62:
```suggestion
  it('marks client error as no-retry', async () => {
```

packages/plugin-electron-deliver-minidumps/deliver-minidumps.js
```diff
@@ -0,0 +1,38 @@
...
+    // make sure that the Electron CrashReporter is configured
+    const metadata = filestore.createAppRunMetadata()
+    crashReporter.start({
+      submitURL: '',
+      uploadToServer: false,
+      extra: metadata
+    })
+
+    nativeClient.install(
+      join(filestore.getPaths().runinfo, metadata.bugsnag_crash_id),
```
imjoehaines, line 25: This could use `Filestore.getEventInfoPath` so it doesn't need to know where the file actually lives:

```suggestion
      filestore.getEventInfoPath(metadata.bugsnag_crash_id),
```

imjoehaines (approved):

imjoehaines (approved):

PR #1467, opened 2021-07-08
Fix crashes when `enabledBreadcrumbTypes` is null

imjoehaines:
## Goal

Following on from https://github.com/bugsnag/bugsnag-js/pull/1466, this PR fixes crashes in various plugins that assumed `enabledBreadcrumbTypes` must always be an array. They now allow for the case where it may be null and handle it accordingly

## Testing

Existing tests & new unit tests. Manually tested in browser, ~~electron~~ and ~~react native~~ projects

### DIFF
--- a/test/react-native/features/fixtures/ios-module/BugsnagModule.m
+++ b/test/react-native/features/fixtures/ios-module/BugsnagModule.m
@@ -61,7 +61,7 @@ BugsnagConfiguration *createConfiguration(NSDictionary * options) {
   if (options[@"enabledReleaseStages"] != nil) {
     [config setEnabledReleaseStages:[NSSet setWithArray:options[@"enabledReleaseStages"]]];
   }
-  if (options[@"enabledBreadcrumbTypes"] != nil) {
+  if (options[@"enabledBreadcrumbTypes"] && ![options[@"enabledBreadcrumbTypes"] isEqual:[NSNull null]]) {
     [config setEnabledBreadcrumbTypes:[NSSet setWithArray:options[@"enabledBreadcrumbTypes"]]];
   }
   if (options[@"configMetaData"] != nil) {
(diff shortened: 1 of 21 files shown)


test/react-native/features/fixtures/ios-module/BugsnagModule.m
```diff
@@ -62,7 +62,11 @@ @implementation BugsnagModule
     [config setEnabledReleaseStages:[NSSet setWithArray:options[@"enabledReleaseStages"]]];
   }
   if (options[@"enabledBreadcrumbTypes"] != nil) {
-    [config setEnabledBreadcrumbTypes:[NSSet setWithArray:options[@"enabledBreadcrumbTypes"]]];
+    if ([options[@"enabledBreadcrumbTypes"] isEqual:[NSNull null]]) {
```
kattrali, line 65:
```suggestion
if (options[@"enabledBreadcrumbTypes"] && ![options[@"enabledBreadcrumbTypes"] isEqual:[NSNull null]]) {
```

kattrali (approved):
Is it possible to either:

* Coerce `enabledBreadcrumbTypes` to always be an array during config validation (where null -> `all`)? or
* Add a helper on config to handle the null check? Most of the validation checks now are of the form `if (config.breadcrumbTypes !== null && config.breadcrumbTypes.contain(type))` where maybe it could be something like `if (config.isBreadcrumbTypeEnabled(type))`

Trying to think of a way to prevent this kind of issue going forward since every future plugin/consumer of this config option basically needs to have this check. Stronger type validation on our part? Additional linting?

test/react-native/features/fixtures/ios-module/BugsnagModule.m
kattrali: I meant to leave it as the default, treating `nil` and `NSNull` the same

imjoehaines:
> Trying to think of a way to prevent this kind of issue going forward since every future plugin/consumer of this config option basically needs to have this check. Stronger type validation on our part? Additional linting?

I avoided doing anything extra in core to try to avoid increasing the browser bundle size, but it turns out that a helper method actually saves space because it removes duplication from plugins

I've opened #1468 for this as it affects files in this PR, #1466 and files that weren't touched by either PR

PR #1468, opened 2021-07-12
Add `_isBreadcrumbTypeEnabled` helper to core

imjoehaines:
## Goal

Building on #1467, this PR adds a new helper method to the core Client: `_isBreadcrumbTypeEnabled`

This can be used by plugins to avoid having to check for null everywhere, which saves duplication and should make this kind of error much less likely to happen again

The de-duplication also saves a few bytes in the browser bundle

## Testing

Tests added for the new helper and existing tests cover the plugins still work

### DIFF
--- a/packages/core/test/client.test.ts
+++ b/packages/core/test/client.test.ts
@@ -1,6 +1,8 @@
 import Client from '../client'
 import Event from '../event'
 import Session from '../session'
+import breadcrumbTypes from '../lib/breadcrumb-types'
+import { BreadcrumbType } from '../types/common'
 
 describe('@bugsnag/core/client', () => {
   describe('constructor', () => {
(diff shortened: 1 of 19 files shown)


packages/core/test/client.test.ts
```diff
@@ -627,6 +629,43 @@ describe('@bugsnag/core/client', () => {
     })
   })
 
+  describe('_isBreadcrumbTypeEnabled()', () => {
+    it.each(breadcrumbTypes)('returns true for "%s" when enabledBreadcrumbTypes is not configured', (type) => {
+      const client = new Client({ apiKey: 'API_KEY_YEAH', enabledBreadcrumbTypes: null })
```
kattrali, line 634: Was this intending `enabledBreadcrumbTypes` to be undefined? This test case and the following one seem to be identical.
```suggestion
      const client = new Client({ apiKey: 'API_KEY_YEAH' })
```

kattrali (approved):

packages/core/test/client.test.ts
imjoehaines: It was indeed, good spot!

PR #1471, opened 2021-07-14
Network aware minidump delivery

lemnik:
## Goal
Minidumps should only be delivered if there is network connectivity, and delivery should be aware of changes to the network connectivity (turning off and on as required). By using the NetworkStatus module, the minidump-loop starts and stops when a network connectivity change is detected.

## Design
Extracted `NetworkStatus` from the electron-delivery module into a new electron-network-status module, so that it can be accessed by both electron-delivery and the plugin-electron-deliver-minidump.

## Testing
Added unit tests to cover the network-aware minidump logic.


packages/electron-network-status/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "7.10.0",
+  "main": "network-status.js",
+  "description": "@bugsnag/electron package for watching network status",
+  "author": "Bugsnag",
+  "license": "MIT",
+  "homepage": "https://www.bugsnag.com/",
+  "devDependencies": {
+    "@bugsnag/core": "^7.10.0",
+    "@bugsnag/plugin-electron-client-state-manager": "^7.10.0"
+  },
```
kattrali, line 12: The package manifest should define which files are included.
```suggestion
  },
  "files": [
    "*.js"
  ],
```

packages/plugin-electron-deliver-minidumps/minidump-loop.js
```diff
@@ -44,7 +44,12 @@ module.exports = class MinidumpDeliveryLoop {
         // if we had a successful delivery - remove the minidump from the queue, and schedule the next
         this._minidumpQueue.remove(minidump)
       } catch (e) {
-        this._onerror(e, minidump)
+        // if the error is not retryable: remove the minidump from the queue and signal the error
+        if (!e.isRetryable) {
+          this._minidumpQueue.remove(minidump)
+          this._onerror(e, minidump)
+        }
+        // if the error is retryable: leave things as-is and we will attempt a redelivery
```
imjoehaines, line 52: I think this is an unintentional change? Looks like a recent commit got reverted

packages/plugin-electron-deliver-minidumps/package.json
```diff
@@ -18,7 +18,8 @@
     "send-minidump.js"
   ],
   "dependencies": {
-    "form-data": "^4.0.0"
+    "form-data": "^4.0.0",
+    "@bugsnag/electron-network-status": "^7.10.0"
```
imjoehaines, line 22: We've had issues with putting our own packages in dependencies because it allows you to end up with multiple versions of the same package installed at the same time. Mostly this just causes TS errors so it's probably fine in this case, but usually we'd make this a peer dependency of `plugin-electron-deliver-minidumps` and `delivery-electron`, with the main `@bugsnag/electron` package pulling it in as a real dependency

Some background on PLAT-4872

packages/delivery-electron/package.json
```diff
@@ -18,6 +18,7 @@
   "license": "MIT",
   "devDependencies": {
     "@bugsnag/core": "^7.10.0",
+    "@bugsnag/electron-network-status": "^7.10.0",
```
imjoehaines, line 21: This should also be a peer dependency

packages/plugin-electron-deliver-minidumps/test/minidump-loop.test.ts
```diff
@@ -118,4 +120,57 @@ describe('electron-minidump-delivery: minidump-loop', () => {
...
+      // ensure that nothing is delivered while disconnected
+      await runDeliveryLoop(1)
+      expect(sendMinidump).toBeCalledTimes(0)
+
+      // connect the network
+      emitter.emit('MetadataReplace', { metadata: { device: { online: true } } })
+
+      // check that we've started delivering minidumps
+      await runDeliveryLoop(1)
+      expect(sendMinidump).toBeCalledTimes(0)
```
imjoehaines, line 149: > check that we've started delivering minidumps

Doesn't this check the opposite? I would expect `sendMinidump` to start being called if we're now online

packages/plugin-electron-deliver-minidumps/test/minidump-loop.test.ts
```diff
@@ -118,4 +120,57 @@ describe('electron-minidump-delivery: minidump-loop', () => {
...
+      loop.watchNetworkStatus(statusWatcher)
+
+      // ensure that the first minidump is delivered
+      await runDeliveryLoop(1)
+      expect(sendMinidump).toBeCalledTimes(1)
+
+      // disconnect the network
+      emitter.emit('MetadataUpdate', { section: 'device', values: { online: false } }, null)
+
+      // check that we've started delivering minidumps
```
imjoehaines, line 171:
```suggestion
      // check that we've stopped delivering minidumps
```

packages/electron-network-status/package.json
imjoehaines: You can test this with `npm pack --dry-run <package>` to see which files would end up getting published to NPM

kattrali (approved):

imjoehaines (approved):

PR #1475, opened 2021-07-15
scan all minidumps in crash dir

lemnik:
## Goal
Electron uses different directory layouts for minidumps depending on both platform and how the crash reporter has been configured. To keep things simple: scan all minidump files under the `crashDumps` directory. 

## Testing
Mostly relied on the existing testing (unit tests and integration tests). The filestore unit test now checks that the minidump path is the `crashDumps` directory rather than a subdirectory of it.

### DIFF
--- a/packages/electron-filestore/test/filestore.test.ts
+++ b/packages/electron-filestore/test/filestore.test.ts
@@ -29,7 +29,7 @@ describe('FileStore', () => {
       expect(paths.sessions).toEqual(join(base, 'sessions'))
       expect(paths.device).toEqual(join(base, 'device.json'))
       expect(paths.runinfo).toEqual(join(base, 'runinfo'))
-      expect(dirname(paths.minidumps)).toEqual(crashes)
+      expect(paths.minidumps).toEqual(crashes)
     })
   })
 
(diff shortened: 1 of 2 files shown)


packages/electron-filestore/test/filestore.test.ts
```diff
@@ -148,6 +148,21 @@ describe('FileStore', () => {
...
+      await writeFile(join(base, 'some-other-thing.ps'), 'not a crash')
+      await createMinidump('report01.dmp', '')
+      await createMinidump(join('reports', 'report02.dmp'), id)
+      await createMinidump(join('pending', 'report03.dmp'), id)
+
+      const dumps = await store.listMinidumps() as any[]
+      const dumpPaths = dumps.map(d => d.minidumpPath)
+      expect(dumpPaths).toContain(join(base, 'report01.dmp'))
+      expect(dumpPaths).toContain(join(base, 'reports', 'report02.dmp'))
+      expect(dumpPaths).toContain(join(base, 'pending', 'report03.dmp'))
```
imjoehaines, line 164: To be a bit stricter, is it possible to use `toEqual` here instead of checking each path individually? I forget if Jest requires arrays to have the same order, but if so we could sort them first

```suggestion
      const dumpPaths = dumps.map(d => d.minidumpPath)
      expect(dumpPaths).toEqual([
        join(base, 'report01.dmp'),
        join(base, 'reports', 'report02.dmp'),
        join(base, 'reports', 'report03.dmp')
      ])
```

If not, it'd be good to assert against the length of `dumpPaths`, so we know it doesn't contain "some-other-thing.ps" or any other files

imjoehaines (approved):

PR #614, opened 2019-08-16
fix(plugin-koa): Ensure unhandled Koa errors are logged out

bengourley:
Our plugin adds an `"onerror"` handler to the Koa app, which [stops the default from being called](https://github.com/koajs/koa/blob/a245d18a131341feec4f87659746954e78cae780/lib/application.js#L129).

The default error handler which prints the error is available at `ctx.app.onerror` so by calling through to that after notifying, it exhibits the default behaviour.

Tested manually with `notifyReleaseStages` both allowing and preventing delivery and added a unit test to ensure the original handler is called.

__Output before:__

```
[bugsnag] Report not sent due to releaseStage/notifyReleaseStages configuration
```

__Output after:__

```
[bugsnag] Report not sent due to releaseStage/notifyReleaseStages configuration

  Error: Invalid DSL syntax
      at /Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/app.js:29:29
      at dispatch (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-router/node_modules/koa-compose/index.js:44:32)
      at next (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-router/node_modules/koa-compose/index.js:45:18)
      at /Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-router/lib/router.js:346:16
      at dispatch (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-router/node_modules/koa-compose/index.js:44:32)
      at /Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-router/node_modules/koa-compose/index.js:36:12
      at dispatch (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-router/lib/router.js:351:31)
      at dispatch (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-compose/index.js:42:32)
      at requestHandler (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/dist/bugsnag-koa.js:778:15)
      at dispatch (/Users/bengourley/Development/bugsnag-js/packages/plugin-koa/scratch/node_modules/koa-compose/index.js:42:32)
```

Fixes #602

PR #1482, opened 2021-07-20
Fix the Koa plugin suppressing other error handlers

imjoehaines:
## Goal

The Koa plugin currently works by catching exceptions in the request handler and notifying Bugsnag. This works but prevents Koa from detecting that there was an error and so it doesn't trigger error handlers to run. This also means that Bugsnag's error handler will never be called

This PR overhauls the plugin so that it works primarily through the error handler, not the request handler. The request handler still exists to attach request data as metadata and create `ctx.bugsnag`, but does nothing else. The error handler will then be triggered by Koa when an error is thrown and Koa will then ensure all other error handlers also run

The error handler is now more complicated as logic from the request handler has been moved into it. Specifically:

- it now checks the error status code and only notifies if the code is a 5xx (originally https://github.com/bugsnag/bugsnag-js/pull/541)
- it will delegate to the default Koa error handler if there are no other error handlers. This ensures we don't change the default behaviour of Koa, as it does not attach its internal error handler if any other error handler has been added (originally https://github.com/bugsnag/bugsnag-js/pull/614)

Finally we no longer set the status code ourselves as Koa's default error handling does that already. Assertions against the status code have been added to the MazeRunner tests

## Testing

Unit tests have been added for most functionality (where it makes sense) and the Maze Runner tests have been updated to ensure other error handlers run

### DIFF
--- a/test/node/features/koa.feature
+++ b/test/node/features/koa.feature
@@ -8,7 +8,7 @@ Background:
   And I wait for the host "koa" to open port "80"
 
 Scenario: a synchronous thrown error in a route
-  Then I open the URL "http://koa/err"
+  Then I open the URL "http://koa/err" and get a 500 response
   And I wait to receive an error
   Then the error is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
   And the event "unhandled" is true
@@ -23,7 +23,7 @@ Scenario: a synchronous thrown error in a route
   And the event "request.clientIp" is not null
 
 Scenario: an asynchronous thrown error in a route
-  Then I open the URL "http://koa/async-err"
+  Then I open the URL "http://koa/async-err" and get a 500 response
   And I wait to receive an error
   Then the error is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
   And the event "unhandled" is true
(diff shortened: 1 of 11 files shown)


test/node/features/koa.feature
```diff
@@ -58,27 +64,32 @@ Scenario: an error thrown before the requestHandler middleware
...
+  Then I open the URL "http://koa/throw-non-error" and get a 500 response
   And I wait to receive an error
   Then the error is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
   And the event "unhandled" is true
   And the event "severity" equals "error"
   And the event "severityReason.type" equals "unhandledErrorMiddleware"
-  And the exception "errorClass" equals "InvalidError"
-  And the exception "message" matches "^koa middleware received a non-error\."
+  And the exception "errorClass" equals "Error"
+  And the exception "message" equals 'non-error thrown: \\"error\\"'
```
imjoehaines, line 78: As we rely on Koa catching errors now, it will internally wrap them in an Error instance before they reach Bugsnag (see [Koa's `context.js`](https://github.com/koajs/koa/blob/698ce0afbfac6480400625729a4b8fc4b4203fdc/lib/context.js#L119)). This changes the error class and message, but is otherwise equivalent

djskinner (approved):

PR #1490, opened 2021-07-29
Do not save Instance State in e2e test fixture

twometresteve:
## Goal

Avoid a test flake caused by Android keeping the state Bundle from before a crash and passing it back into the application with a Drawable that does not implement the `getConstantState` method.  This resulted in the following exception in the Appium log:

```
Caught exception
java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable$ConstantState.newDrawable(android.content.res.Resources)' on a null object reference
```

## Changeset

Override `onSaveInstanceState` in all `Activity` classes to prevent and Instance State being saved in the first place.

## Testing

Passing CI shows that the tests are no worse off than they were, although we can't really tell at this stage if the exception is truly resolved.

### DIFF
--- a/examples/reactnative/rn063example/android/app/src/main/java/com/rn063example/MainActivity.java
+++ b/examples/reactnative/rn063example/android/app/src/main/java/com/rn063example/MainActivity.java
@@ -1,5 +1,7 @@
 package com.rn063example;
 
+import android.os.Bundle;
+
 import com.facebook.react.ReactActivity;
 
 public class MainActivity extends ReactActivity {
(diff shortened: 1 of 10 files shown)


examples/reactnative/rn063example/android/app/src/main/java/com/rn063example/MainActivity.java
```diff
@@ -1,5 +1,7 @@
 package com.rn063example;
 
+import android.os.Bundle;
```
twometresteve, line 3: Actually, this seems to have been unnecessary - will remove.

twometresteve: ...although having said that I've then added code to `onCreate` in the next PR to come, so I may as well leave it in.

lemnik (approved):
LGTM

PR #1492, opened 2021-07-29
Dismiss any system (crash) dialog on app launch

twometresteve:
## Goal

Avoid test flakes by dismissing system (crash) dialogs at app launch.

## Design

If a system dialog is present, it prevents Appium from locating elements on the screen.

## Testing

Covered by CI.


lemnik (approved):
LGTM

PR #1495, opened 2021-08-04
Allow the apiKey to be set by event callbacks

lemnik:
## Goal
If `onSend` callbacks to change or set the `apiKey` of an event then this should be the `apiKey` used to send the event. This allows an application to direct events to different projects based on user-defined conditions.

## Testing
Added new unit test coverage.

### DIFF
--- a/packages/plugin-electron-deliver-minidumps/send-minidump.js
+++ b/packages/plugin-electron-deliver-minidumps/send-minidump.js
@@ -4,6 +4,18 @@ const { basename } = require('path')
 const payload = require('@bugsnag/core/lib/json-payload')
 const FormData = require('form-data')
 
+const isValidApiKey = (apiKey, logger) => {
+  if (apiKey) {
+    if (!/^[0-9a-f]{32}$/i.test(apiKey)) {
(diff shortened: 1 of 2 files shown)


packages/plugin-electron-deliver-minidumps/send-minidump.js
```diff
@@ -31,9 +43,12 @@ module.exports = (net, client) => {
   }
 
   const sendMinidump = async (minidumpPath, event) => {
+    const apiKey = event && isValidApiKey(event.apiKey, client._logger)
+      ? event.apiKey
+      : client._config.apiKey
```
imjoehaines, line 48: I don't think this needs to be different to the core Client:

https://github.com/bugsnag/bugsnag-js/blob/34e5f635d67826638008849be60f9adeb2f523be/packages/core/client.js#L328-L329

If we do validate it, it should still be used if it isn't empty:

> If a non-empty apiKey is provided but it is not a string of exactly 32 hexadecimal characters, the notifier shall log a Warning-level validation message but use the value provided.

but if it's wrong then it will log a warning if delivery fails, so I don't think we need to validate this at all

imjoehaines (approved):

PR #1500, opened 2021-08-05
Minidump tests await session requrests

lemnik:
## Goal
Include session counts in the minidump integration tests.

## Changeset
The integration tests can wait for request counts without `expect` behaviour or preset timers.

Allow the test fixture to be started in offline mode. This prevents the `session` request racing against the web-content startup by marking the app `offline` before `Bugsnag.start` in the renderer process.


test/electron/features/native-crash.feature
```diff
@@ -2,26 +2,29 @@ Feature: Native Errors
...
-    And I launch an app
-    Then the total requests received by the server matches:
-      | events    | 0 |
-      | minidumps | 0 |
-    When the server becomes reachable
+    Then the server is unreachable
+    And I launch an app with no network
+    Then the server becomes reachable
     And the app gains network connectivity
+    And I wait for 2 session uploads
```
imjoehaines, line 23: Is this wait necessary? "the total requests received by the server matches" will wait for 3 seconds already

test/electron/features/native-crash.feature
```diff
@@ -2,26 +2,29 @@ Feature: Native Errors
 
   Scenario: A minidump is uploaded on native error
     When I launch an app
+    And I wait for 1 session upload
     And I click "main-process-crash"
     Then I launch an app
+    And I wait for 2 session uploads
```
imjoehaines, line 8: Is this wait necessary? "the total requests received by the server matches" will wait for 3 seconds already

test/electron/features/native-crash.feature
```diff
@@ -2,26 +2,29 @@ Feature: Native Errors
 
   Scenario: A minidump is uploaded on native error
     When I launch an app
+    And I wait for 1 session upload
```
imjoehaines, line 5: Do we need a new step for this? Could we do something like:

```suggestion
    Then the total requests received by the server matches:
      | sessions  | 1 |
```

kattrali: I _think_ I get it - since the session may take a second or two to send (depending on when the electron framework comes up), the hard crash means that the session may or may not be sent, **and** may or may not be persisted. So this step is trying to remove that variability?

imjoehaines: There's a 3 second wait in the "the total requests received by the server matches" step already. AFAICT the only difference with the new step is it waits an extra 1.9 seconds, but otherwise this would be the same

lemnik: The new step returns immediately upon delivery (rather than having a fixed delay), so it tends to be slightly quicker (not that this is a big deal).

It was also intended as a "softer" action rather than a strict assertion (although the result is basically the same).

Happy to change it around though.

imjoehaines: I don't think it's a big deal either way; it's a bit clearer with the new step so lets leave it like that

imjoehaines (approved):

PR #1501, opened 2021-08-06
electron: persist and retry events on network error

kattrali:
## Goal

Ensure events/sessions aren't lost in the event of a connectivity issue

## Design

* Marked network errors as retryable prior to invoking the completion callback. No other changes were needed to ensure other enqueued errors are sent, as payloads are then sent to the back of the queue (ordered by date).

## Tests

Refactored the tests to separate the server reachability cases from the network connectivity cases, so now there are one of each scenario:

* server and app connected from the start, single crash
* no network connection, single crash, connection restored during app lifetime
* no network connection, multiple crashes enqueued, send next connected launch
* no server available, multiple crashes enqueued, send next connected launch

## Discussion

Its somewhat subjective, but I opted to print all types of request counts at each junction to ensure requests were not attempted to be sent when offline. The upside is that its consistent (because I confused myself a few times), the downside is that its very verbose.

### DIFF
--- a/packages/delivery-electron/delivery.js
+++ b/packages/delivery-electron/delivery.js
@@ -6,7 +6,13 @@ const NetworkStatus = require('@bugsnag/electron-network-status')
 
 const delivery = (client, filestore, net, app) => {
   const send = (opts, body, cb) => {
+    const errorHandler = err => {
+      err.isRetryable = true
+      cb(err)
+    }
+
     const req = net.request(opts, response => {
+      req.removeListener('error', errorHandler)
       if (isOk(response)) {
         cb(null)
       } else {
@@ -16,10 +22,7 @@ const delivery = (client, filestore, net, app) => {
       }
     })
 
-    req.on('error', err => {
-      err.isRetryable = true
-      cb(err)
-    })
+    req.on('error', errorHandler)
 
     try {
       req.write(body)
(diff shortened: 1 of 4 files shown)


lemnik (approved):
LGTM

packages/delivery-electron/delivery.js
```diff
@@ -16,7 +16,10 @@ const delivery = (client, filestore, net, app) => {
       }
     })
 
-    req.on('error', cb)
+    req.on('error', err => {
+      err.isRetryable = true
```
imjoehaines, line 20: Do we definitely want to retry for every kind of error?

Some errors can happen after the server has responded, so could possibly lead to duplicate events — not sure whether that's a valid concern though, since a successful response really shouldn't have an error 🤔

kattrali: Huh, that's a case I hadn't considered :thinking:

In that case, maybe `isRetryable` should only be set if the response callback hasn't been invoked? Though I'm struggling to come up with a scenario to test it.

imjoehaines: Yeah that sounds like a good solution, but does seem tricky to test for real. A unit test could probably force it to happen by manuallyemitting an error event after returning a response from the mock server

kattrali: The mock setup in the unit tests seems to perceive an emitted error as an uncaught exception (and automatically fail the test) for some reason. I'll revisit this.

PR #1504, opened 2021-08-10
Deliver non-fatal minidumps

lemnik:
## Goal
Deliver non-fatal minidump files that are produced by crashing child processes (such as the `render` process). The new `MinidumpWatcher` uses file watchers on the minidump directory, and each of its direct child directories (to account for platform differences). The watcher is enabled and disabled based on network availability.

When any `'rename'` event is delivered the watcher ensures that the minidump-delivery-loop is running. The existing minidump loop will then pickup the file and ensure it is delivered to the configured endpoint.

## Testing
Added unit tests and a new end-to-end test to ensure that non-fatal minidumps are delivered without restarting the application.

### DIFF
--- a/packages/plugin-electron-deliver-minidumps/minidump-queue.js
+++ b/packages/plugin-electron-deliver-minidumps/minidump-queue.js
@@ -1,12 +1,13 @@
 module.exports = class MinidumpQueue {
   constructor (filestore) {
     this._filestore = filestore
+    this._removedMinidumpPaths = new Set()
   }
 
   async peek () {
     try {
       const minidumps = await this._filestore.listMinidumps()
-      return minidumps[0]
+      return minidumps.find(minidump => !this._removedMinidumpPaths.has(minidump.minidumpPath))
     } catch (e) {
     }
   }
@@ -15,6 +16,9 @@ module.exports = class MinidumpQueue {
     if (!minidump) {
       return
     }
+
+    // we mark the minidump file as "removed" to avoid problems if we cannot delete the file
+    this._removedMinidumpPaths.add(minidump.minidumpPath)
 
     // we ignore any file delete failures - the file has probably already been deleted
     this._filestore.deleteMinidump(minidump).catch(e => {})
--- a/packages/plugin-electron-deliver-minidumps/minidump-queue.js
+++ b/packages/plugin-electron-deliver-minidumps/minidump-queue.js
@@ -1,30 +1,20 @@
 module.exports = class MinidumpQueue {
   constructor (filestore) {
     this._filestore = filestore
-    this._minidumps = null
   }
 
   async peek () {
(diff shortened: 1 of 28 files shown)


packages/plugin-electron-deliver-minidumps/minidump-queue.js
```diff
@@ -1,31 +1,21 @@
...
   async peek () {
-    if (!this._minidumps) {
-      try {
-        this._minidumps = await this._filestore.listMinidumps()
-      } catch (e) {
-        this._minidumps = []
-      }
+    try {
+      const minidumps = await this._filestore.listMinidumps()
+      return minidumps[0]
```
kattrali, line 9: Rather than removing the upload tracking list altogether, it might be worthwhile to track file paths already attempted, in case a file appears at the top of the queue and fails to be deleted, in which case it will block the queue indefinitely.

test/electron/features/native-crash.feature
```diff
@@ -13,8 +13,15 @@ Feature: Native Errors
...
       | minidumps | 1 |
       | sessions  | 2 |
-    And minidump request 0 contains a file form field named "upload_file_minidump"
-    And minidump request 0 contains a form field named "event" matching "minidump-event.json"
+    Then minidump request 0 contains a file form field named "upload_file_minidump"
+    Then minidump request 0 contains a form field named "event" matching "minidump-event.json"
+
+  Scenario: Non-main minidumps are detected and uploaded
+    Given I launch an app
+    And I click "renderer-process-crash"
```
kattrali, line 21: nit: Starting the app and clicking are both actions caused by the step rather than existing conditions.

```suggestion
    When I launch an app
    And I click "renderer-process-crash"
```

kattrali (commented):
Looks good overall, had one concern around blocked uploads, otherwise just nits :broom:

imjoehaines (commented):
I've not finished review but leaving these comments before this PR gets split up

test/electron/fixtures/app/renderer.js
```diff
@@ -7,11 +7,17 @@ const Bugsnag = require('@bugsnag/electron')
...
-function emulateOnlineStatus (online) {
+function emulateOnlineStatus(online) {
   Object.defineProperty(window.navigator, 'onLine', { value: online, configurable: true })
   window.dispatchEvent(new window.Event(online ? 'online' : 'offline'))
 }
 
+document.getElementById('renderer-process-crash').onclick = () => {
+  setTimeout(() => {
+    window.RunnerAPI.renderProcessCrash()
+  }, 10)
```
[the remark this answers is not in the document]
kattrali: It just needs to be async. Without the timeout, the click event would throw an exception in the automation framework since the listener would not be able to tell that the click "succeeded".

PR #1507, opened 2021-08-12
feat(electron): include session info in minidump events

kattrali:
## Design

Amended `plugin-electron-session` to set native session info in crash events, which was previously not set. The session info is set with an incremented `unhandled` even count, which should be correct in the event of a main process crash.

## Changeset

* In `plugin-electron-session` created a new session delegate which sends the event JSON representation of a session to the crash info cache upon change. This info is sent with a subsequent crash. (The native plumbing was already there, so this just hooks it all together)
* Hooked `Session._track()` to update session info in the native cache when new events are detected

## Testing

* Added additional unit-y tests around pausing/resuming sessions, and asserted the contents of the session in start session events
* Updated the native crash event to assert session event counts

### DIFF
--- a/packages/plugin-electron-session/session.js
+++ b/packages/plugin-electron-session/session.js
@@ -1,11 +1,40 @@
 const sessionDelegate = require('@bugsnag/plugin-browser-session')
+const Session = require('@bugsnag/core/session')
 
 const SESSION_TIMEOUT_MS = 60 * 1000
 
-module.exports = (app, BrowserWindow) => ({
+module.exports = (app, BrowserWindow, NativeClient) => ({
   load (client) {
+    // Ensure native session kept in sync with session changes
+    const oldTrack = Session.prototype._track
+    Session.prototype._track = function (...args) {
+      const result = oldTrack.apply(this, args)
+      NativeClient.setSession(serializeForNativeEvent(this))
+      return result
+    }
+
     // load the actual session delegate from plugin-browser-session
     sessionDelegate.load(client)
+
+    // Override the delegate to perform electron-specific synchronization
+    const defaultDelegate = client._sessionDelegate
+    client._sessionDelegate = {
+      startSession: (client, session) => {
+        const result = defaultDelegate.startSession(client, session)
+        NativeClient.setSession(serializeForNativeEvent(client._session))
+        return result
+      },
+      resumeSession: (client) => {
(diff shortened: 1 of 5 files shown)


packages/plugin-electron-session/session.js
```diff
@@ -1,12 +1,39 @@
...
+        NativeClient.setSession(serializeForNativeEvent(client._session))
+      },
+      resumeSession: (client) => {
+        defaultDelegate.resumeSession(client)
+        NativeClient.setSession(serializeForNativeEvent(client._session))
+      },
+      pauseSession: (client) => {
+        defaultDelegate.pauseSession(client)
+        NativeClient.setSession(serializeForNativeEvent(client._session))
+      }
```
imjoehaines, line 34: These should return whatever the `defaultDelegate` methods return:

```suggestion
      startSession: (client, session) => {
        const result = defaultDelegate.startSession(client, session)
        NativeClient.setSession(serializeForNativeEvent(client._session))
        return result
      },
      resumeSession: (client) => {
        const result = defaultDelegate.resumeSession(client)
        NativeClient.setSession(serializeForNativeEvent(client._session))
        return result
      },
      pauseSession: (client) => {
        const result = defaultDelegate.pauseSession(client)
        NativeClient.setSession(serializeForNativeEvent(client._session))
        return result
      }
```

packages/plugin-electron-session/session.js
```diff
@@ -1,12 +1,39 @@
...
+module.exports = (app, BrowserWindow, NativeClient) => ({
   load (client) {
+    // Ensure native session kept in sync with session changes
+    Session.prototype._track = new Proxy(Session.prototype._track, {
+      apply (target, _this, args) {
+        const result = target.apply(_this, args)
+        NativeClient.setSession(serializeForNativeEvent(_this))
+        return result
+      }
+    })
```
imjoehaines, line 15: I guess really this might be overcomplicating things 😅 We could use the same approach as the session delegate:

```suggestion
    // Ensure native session kept in sync with session changes
    const oldTrack = Session.prototype._track
    Session.prototype._track = function (...args) {
      const result = oldTrack.apply(this, args)
      NativeClient.setSession(serializeForNativeEvent(this))
      return result
    }
```
I'm happy either way though, it doesn't really make a difference 🤷

imjoehaines (approved):

PR #1513, opened 2021-08-18
build(deliver-minidumps): add missing license file

kattrali:


imjoehaines (approved):
👍 good catch!

I wonder if it's worth making a small command to create new packages so we don't miss stuff in future; `lerna create` doesn't seem to have all the options we'd need

kattrali:
That's a good idea!

imjoehaines:
Made PLAT-7150 for this

PR #1515, opened 2021-08-18
feat(deliver-minidumps): support adding minidumps, enforce uniqueness

kattrali:
## Goal

Support adding minidumps which appear while the app runs to a queue for later delivery, avoiding any duplication. 

## Design

This behavior has a lot of constraints given:

* The first disk access to generate the list of minidumps should be asynchronous
* We'll need to verify whether a minidump has been "seen" yet by its path, though the queue should keep whole minidump objects
* Delivery cannot start until "ready" (and that starts the lazy loading of existing minidumps) but a crash can happen at any time, even before that point

So it does something a bit unusual, but hopefully still makes sense

## Changeset

* Added `push()` and `hasSeen()` to `MinidumpQueue`, which if initialized, adds a new minidump to the existing ones. If not, the call is ignored (but the minidump should be picked up automatically by `filestore.listMinidumps()` afterwards)

## Testing

* Added new unit tests and integration tests in a subsequent changeset which uses the queue changes to send background minidumps

### DIFF
--- a/packages/plugin-electron-deliver-minidumps/minidump-queue.js
+++ b/packages/plugin-electron-deliver-minidumps/minidump-queue.js
@@ -2,18 +2,33 @@ module.exports = class MinidumpQueue {
   constructor (filestore) {
     this._filestore = filestore
     this._minidumps = null
+    this._seen = new Set()
   }
 
   async peek () {
     if (!this._minidumps) {
       try {
         this._minidumps = await this._filestore.listMinidumps()
+        this._minidumps.forEach(m => this._seen.add(m.minidumpPath))
       } catch (e) {
         this._minidumps = []
       }
     }
 
     return this._minidumps[0]
+  }
+
+  push (minidump) {
+    // if _minidumps has not yet been lazy-loaded, this minidump will be picked
+    // up automatically when the list loads
+    if (this._minidumps && !this._seen.has(minidump.minidumpPath)) {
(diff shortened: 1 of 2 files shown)


packages/plugin-electron-deliver-minidumps/minidump-queue.js
```diff
@@ -16,6 +18,19 @@ module.exports = class MinidumpQueue {
...
+    // up automatically when the list loads
+    if (this._minidumps && !this._seen.has(minidump.minidumpPath)) {
+      this._seen.add(minidump.minidumpPath)
+      this._minidumps.push(minidump)
+    }
+  }
+
+  hasSeen (minidumpPath) {
+    return this._seen.has(minidumpPath)
+  }
```
imjoehaines, line 32: I don't really understand why this method would be necessary, can you explain what it will be used for?

As I understand it, any new minidumps can be pushed to the queue and then `peek` will return them, so I'm not sure why it would be necessary to check if a minidump has been seen already

kattrali: That's fair - this check is used for determining whether a minidump found after a background crash event is actually new or is already being uploaded. Having it exist means that we can skip unnecessarily writing an extra event info file, which has to be done before pushing it into the queue.

imjoehaines (approved):

PR #1517, opened 2021-08-19
adds aws lambda examples

luke-belton:
## Goal

Bugsnag has a [plugin for AWS Lambda](https://docs.bugsnag.com/platforms/javascript/aws-lambda/) but until now there was no example app to demonstrate Bugsnag + AWS Lambda functionality.

## Design

Adds Lambda functions to demonstrate the following:
- Handled exceptions (with `Bugsnag.notify`)
- Unhandled exceptions
- Lambda timeouts
- Promise rejections

Demonstrates Bugsnag's support for both [async and callback](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html) Lambda handlers.

## Changeset

Example app has been added within the `/examples/aws-lambda/simple-app` directory

## Testing

Tested all functions with `sam local run-api` and sam local invoke "<RESOURCE_NAME>" -e <path/to/event.json>


johnkiely1:
Functionality works as expected for both local and deployed set ups. 

I would suggest adding a bit more to the readme: 
Perhaps listing out each function endpoint with a little description of what it does.
 
The retuned {"message":"Internal server error"} for the majority of the endpoints was a little unexpected. I think this is a consequence of the way the lambda function fails. It would be nice if this could be something a little less vague or at least reference it in the readme that this would be expected. 

There is a [open question](https://bugsnag.slack.com/archives/CL6SB711S/p1646316407198869)/potential bug with platforms regarding caching behaviour for surrounding code retrieval as the /async/handled-error and /callback/handled error can return incorrect code. But we could avoid this by either renaming one of the files or moving the notify away from line 12 on one of them. It's probably worth doing this anyway to avoid confusion.  

Errors do display the missing-sourcemaps banner on the dashboard. I wonder if it possible to get these to be produced with the build? if possible add some additional instructions in the readme on how to upload.

luke-belton:
> I would suggest adding a bit more to the readme
I've updated the readme to give a better description of the functions

> The retuned {"message":"Internal server error"} for the majority of the endpoints was a little unexpected
I don't think there's much we can do about this as it's the way the function fails.

> Errors do display the missing-sourcemaps banner on the dashboard
I can't find a way to generate sourcemaps with `sam build`. I think we'd need to look use something like webpack to build the functions if we wanted to generate sourcemaps

examples/aws-lambda/simple-app/README.md
```diff
@@ -0,0 +1,72 @@
...
+cd bugsnag-js/examples/aws-lambda/simple-app
+```
+- Bugsnag is started for each Lambda function with `Bugsnag.start` and the `@bugsnag/plugin-aws-lambda` plugin
+- Errors in your Lambda are handled by wrapping your Lambda handler inside Bugsnag's handler
+- Both async and callback handlers are supported, with examples given for both inside this app
+
+Replace `<YOUR_BUGSNAG_API_KEY>` in `template.yaml` with your own.
+
+### Install dependencies
+A Lambda layer is used to share dependencies (Bugsnag) across the functions. From the project root, install the dependecies before invoking the functions:
```
johnkiely1, line 26: "dependencies" typo above and below

examples/aws-lambda/simple-app/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+cd dependecies-layer
+npm install
+```
```
johnkiely1, line 30: Might want to explicitly add `cd ..`

examples/aws-lambda/simple-app/template.yaml
```diff
@@ -0,0 +1,160 @@
...
+          Properties:
+            Path: /callback/timeout
+            Method: get
+  
+  NodeDependenciesLayer:
+    Type: AWS::Serverless::LayerVersion
+    Properties:
+      LayerName: node-dependencies-layer
+      Description: Dependencies for the Lambda Functions
+      ContentUri: callback-dependencies-layer/
```
johnkiely1, line 158: I think this should be `dependencies-layer/`

johnkiely1, line 158: I think this should be `dependencies-layer/`

johnkiely1 (commented):
There are some failing tests which need to be looked at. 

I've not been able to retest any errors arriving to the dashboard as I encounter multiple errors using a brand new copy of the repo and running the `sam local` command

Firstly :
Invoking callback-handled-exception.lambdaHandler (nodejs14.x)
NodeDependenciesLayer is a local Layer in the template
Image was not found.

FileNotFoundError: [Errno 2] No such file or directory: '/Users/johnkiely/Documents/test_projects/testAwsExample/bugsnag-js/examples/aws-lambda/simple-app/callback-dependencies-layer'

I think this is solved by the following:

template.yaml -> change to  ContentUri: dependencies-layer/

However then I get: 
Invalid lambda response received: Invalid API Gateway Response Keys: {'errorType', 'trace', 'errorMessage'} in {'errorType': 'Runtime.ImportModuleError', 'errorMessage': "Error: Cannot find module '@bugsnag/js'\nRequire stack:\n- /var/task/async-handled-exception.js\n- /var/runtime/UserFunction.js\n- /var/runtime/Runtime.js\n- /var/runtime/index.js", 'trace': ["Runtime.ImportModuleError: Error: Cannot find module '@bugsnag/js'", 'Require stack:', '- /var/task/async-handled-exception.js', '- /var/runtime/UserFunction.js', '- /var/runtime/Runtime.js', '- /var/runtime/index.js', '    at _loadUserApp (/var/runtime/UserFunction.js:221:13)', '    at Object.module.exports.load (/var/runtime/UserFunction.js:279:17)', '    at Object.<anonymous> (/var/runtime/index.js:43:34)', '    at Module._compile (internal/modules/cjs/loader.js:1085:14)', '    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)', '    at Module.load (internal/modules/cjs/loader.js:950:32)', '    at Function.Module._load (internal/modules/cjs/loader.js:790:12)', '    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:75:12)', '    at internal/main/run_main_module.js:17:47']}

Looks like something is broken with the dependency resolution. 

Also I think we may want to consider not including the package-lock.json file. I was using npm 6 which gave warnings because lockfileVersion:2 is not supported. https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion.  Also npm 8 doesn't support lockfileVersion:1. So users may need to generate their own based on npm version  as 6-8 are all currently widely used. 

Typo's/suggestions in readme. see in line comments

himanshufize (approved):

himanshufize:
I try to implement bugsnag for handled error but it not able to log error when I try to call notify inside catch block. I try block it doing it.

examples/aws-lambda/simple-app/async/handled-exception/async-handled-exception.js
```diff
@@ -0,0 +1,25 @@
+const Bugsnag = require('@bugsnag/js')
+const BugsnagPluginAwsLambda = require('@bugsnag/plugin-aws-lambda')
+
```
himanshufize, line 3: Just one request. Can you add an example in which we call multiple functions and in those function error could occur but they should not break the flow of execution? 

This in regard use case when failure in some place is OK but still be need log regarding those failures - One API request

luke-belton:
@johnkiely1 - I haven't been able to repro the error `Invalid lambda response received: Invalid API Gateway Response Keys:`. Do you know what version of node you were running to test this?

Other than that I've pushed some changes that cover the rest of your comments I hope!

PR #1535, opened 2021-10-01
Thread states

lemnik:
## Goal
Ensure `Thread.state` is visible to the bugsnag-js


fractalwrench (approved):
LGTM pending CI

fractalwrench:
It may also be worth copying the changelog over from #1534 as that PR could then be closed

PR #1546, opened 2021-10-20
Add feature flag delegate

imjoehaines:
## Goal

This is like [the metadata delegate](https://github.com/bugsnag/bugsnag-js/blob/2723b7ea90b33d785198f5104216fbcf533af936/packages/core/lib/metadata-delegate.js), but for feature flags. It holds a common API for both Client & Event to use in future, though is unused as of this PR

## Design

This implements two functions: `add` & `merge`. `add` is used to add a single flag at a time and is given a `name` and optionally a `variant`. `merge` will add new flags and overwrite existing flags from an array of objects with a `name` and optionally a `variant` (e.g. `{ name: 'flag' }` or `{ name: 'flag', variant: 'example' }`)

Removing feature flags should be straight-forward enough that it's actually less duplication to implement in both Client & Event directly, rather than going through a delegate

### DIFF
--- a/packages/core/lib/test/feature-flag-delegate.test.ts
+++ b/packages/core/lib/test/feature-flag-delegate.test.ts
@@ -1,0 +1,195 @@
+import delegate from '../feature-flag-delegate'
+
+describe('feature flag delegate', () => {
+  describe('#add', () => {
+    it('should do nothing if name is not passed', () => {
+      const existingFeatures = { abc: 'xyz' }
+
+      delegate.add(existingFeatures)
+
+      expect(existingFeatures).toStrictEqual({ abc: 'xyz' })
+    })
+
+    it('should do nothing if name is undefined but variant is passed', () => {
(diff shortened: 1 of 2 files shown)


packages/core/lib/test/feature-flag-delegate.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    it('should do nothing if name is undefined but variant is passed', () => {
+      const existingFeatures = { abc: 'xyz' }
+
+      delegate.add(existingFeatures, undefined, '???')
+
+      expect(existingFeatures).toStrictEqual({ abc: 'xyz' })
+    })
+
+    it('should do nothing if name is null but variant is', () => {
```
djskinner, line 21:
```suggestion
    it('should do nothing if name is null but variant is passed', () => {
```

djskinner (approved):

lemnik (approved):

PR #1556, opened 2021-10-27
Bump Expo package versions

foolem:
## Goal

Expo SDK 43 has been launched with some updates on unimodules. New packages rely on expo-core-modules instead of unimodules.

## Design

Just updated the packages versions.

## Changeset

Package.json files

## Testing

Locally tested


imjoehaines:
Thanks for the PR, @foolem

The package versions need to match Expo exactly so I've merged this as https://github.com/bugsnag/bugsnag-js/pull/1566 with the version numbers tweaked

PR #1562, opened 2021-11-03
Browser maze runner tests for feature flags

imjoehaines:
## Goal

Adds some maze runner tests for feature flags to the browser test suite

### DIFF
--- a/test/browser/features/feature_flags.feature
+++ b/test/browser/features/feature_flags.feature
@@ -1,0 +1,62 @@
+@unhandled
+Feature: Feature flags
+
+Scenario: feature flags are attached to unhandled errors
+  When I navigate to the test URL "/feature_flags/script/unhandled_error.html"
+  Then I wait to receive an error
+  And the error is a valid browser payload for the error reporting API
+  And event 0 is unhandled
+  And the error payload field "events.0.featureFlags" is an array with 8 elements
+  And the event "featureFlags.0.featureFlag" equals "from config 1"
+  And the event "featureFlags.0.variant" equals "1234"
+  And the event "featureFlags.1.featureFlag" equals "from config 2"
+  And the event "featureFlags.1" has no variant
+  And the event "featureFlags.2.featureFlag" equals "added at runtime 1"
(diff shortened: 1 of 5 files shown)


test/browser/features/feature_flags.feature
```diff
@@ -0,0 +1,62 @@
...
+  And the event "featureFlags.7.featureFlag" equals "from global on error 3"
+  And the event "featureFlags.7.variant" equals "111"
+
+Scenario: feature flags are attached to handled errors
+  When I navigate to the test URL "/feature_flags/script/handled_error.html"
+  Then I wait to receive an error
+  And the error is a valid browser payload for the error reporting API
+  And event 0 is handled
+  And the error payload field "events.0.featureFlags" is an array with 8 elements
+  And the event "featureFlags.0.featureFlag" equals "from config 1"
```
Cawllec, line 33: We could simplify this with a new step that consumes a table and applies its contents across the featureFlags.  That way this could also be unordered which would remove a source of flakiness.

For now, provided we think flakes due to the ordering of feature flags are unlikely, I propose we raise a ticket to add said step to Maze-runner for future use in this and other notifiers.

Proposed table would look something like:
```
| featureFlag    | variant              |
|--------------|----------------|
| from config 1 | 1234                 |
| from config 2 |                          |
```
etc

test/browser/features/steps/browser_steps.rb
```diff
@@ -41,3 +41,15 @@
   txt = Maze.driver.find_element(id: 'bugsnag-test-state').text
   assert_equal('DONE', txt, "Expected #bugsnag-test-state text to be 'DONE'. It was '#{txt}'.")
 end
+
+Then("the event {string} has no variant") do |field_path|
```
Cawllec, line 45: As mentioned above we should add some steps to maze-runner core to perform these steps unless they're going to seriously vary between each notifier.

Cawllec (approved):
Left a couple of comments regarding integration the feature flags into maze-runner, but everything looks good here and we can use it as a baseline for which steps would be useful.

Issue #1561, opened 2021-11-02
TypeError: undefined is not an object (evaluating 'this["BS~~U"].replace')


olehleskiv:
### Describe the bug
After updating to the last version we started to receive a lot of identical errors in bugsnag web app from different parts of our application - `TypeError: undefined is not an object (evaluating 'this["BS~~U"].replace')`

<img width="708" alt="Screenshot at Nov 02 23-50-54" src="https://user-images.githubusercontent.com/6775110/139956847-38abd43b-0128-478e-91b4-8617e79887e7.png">

I investigated a bit and found that this commit  from 16 Jun is causing this error
https://github.com/bugsnag/bugsnag-js/commit/ece06354bd8c6d402ee79db81663f746e6fe7e0c#diff-9b5eddef7819dd0f66be05dc7119868c717bc9c1f93622f446ddbf88ab72ce41R63

### Environment
* Bugsnag version: 7.11.0
* Browser framework version (if any):
    * Angular: 11.2.14
Reproducable on different browsers

<details><summary>Error messages:</summary>

```
Error · TypeError: undefined is not an object (evaluating 'this["BS~~U"].replace')
```
</details>

mattdyoung:
Hi @olehleskiv 

Thanks for the detailed report - we're taking a look.

imjoehaines:
Hi @olehleskiv,

I've fixed the crash in #1564, which should go out in the next release. However, this error means something about the `XMLHttpRequest` object isn't quite right and will prevent network breadcrumbs being collected

It'd be good to figure out what is causing this to happen so that it's possible to collect network breadcrumbs in your app. Are you using another library that monkey patches `XMLHttpRequest`? How do you make network requests in your application?

To elaborate on the cause of this error:

We [patch the `open` method](https://github.com/bugsnag/bugsnag-js/blob/8a3e0b59db5711a8e9a7e1a78ef507052919b921/packages/plugin-network-breadcrumbs/network-breadcrumbs.js#L33) and attach the URL to the `XMLHttpRequest` instance. The URL can then be read from the same instance [in `handleXHRLoad`](https://github.com/bugsnag/bugsnag-js/blob/8a3e0b59db5711a8e9a7e1a78ef507052919b921/packages/plugin-network-breadcrumbs/network-breadcrumbs.js#L63)

[...]

PR #1564, opened 2021-11-03
Prevent a crash when XMLHttpRequest data goes missing

imjoehaines:
## Goal

This fixes the crash reported in https://github.com/bugsnag/bugsnag-js/issues/1561, though this means the network breadcrumb plugin is in a broken state — the data attached to the `XMLHttpRequest` has been removed and therefore we can't leave network breadcrumbs if this happens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## TBC
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


lemnik (approved):

djskinner (approved):
how peculiar

PR #1587, opened 2021-11-17
Fix in-flight plugin type

skirsten:
Fixed the type of the in-flight plugin.

Before it was exporting the actual type instead of a declaration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


yousif-bugsnag:
Hi @skirsten, thanks for the PR!

This just needs to make the export less ambiguous, e.g.:

```
declare const BugsnagInFlightPlugin: BugsnagInFlight
export default BugsnagInFlightPlugin 
```

We'll then look to merge this as soon as priorities allow

imjoehaines (approved):
Thanks @skirsten!

PR #1589, opened 2021-11-19
Add maze runner tests for feature flags in node

imjoehaines:
## Goal

Adds some maze runner tests for feature flags to the Node test suite. These focus on the request-scoped client having separate lists of feature flags (so each request can have their own set of flags), while also ensuring global flags are supported


Cawllec (approved):
LGTM

PR #1592, opened 2021-11-22
Add 'featureFlags' to Config type definition

imjoehaines:
## Goal

The typescript definition for `Config` wasn't updated in https://github.com/bugsnag/bugsnag-js/pull/1551

This PR adds the missing `featureFlags` property to `Config` and ignores the deliberate errors in our tests (where we pass invalid types on purpose)

### DIFF
--- a/packages/core/test/client.test.ts
+++ b/packages/core/test/client.test.ts
@@ -886,6 +886,7 @@ describe('@bugsnag/core/client', () => {
         featureFlags: [
           { name: 'a', variant: '1' },
           { name: 'b' },
+          // @ts-expect-error
           { name: 'c', variant: 3 }
         ]
       })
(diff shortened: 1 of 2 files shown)


packages/core/test/client.test.ts
```diff
@@ -886,6 +886,7 @@ describe('@bugsnag/core/client', () => {
         featureFlags: [
           { name: 'a', variant: '1' },
           { name: 'b' },
+          // @ts-expect-error
```
imjoehaines, line 889: `variant` should be a string but we pass a number to check it gets coerced into a string

packages/core/test/client.test.ts
```diff
@@ -886,6 +886,7 @@ describe('@bugsnag/core/client', () => {
...
           { name: 'c', variant: 3 }
```
imjoehaines, line 890: `variant` should be a string but we pass a number to check it gets coerced into a string

packages/core/test/client.test.ts
```diff
@@ -906,8 +907,9 @@ describe('@bugsnag/core/client', () => {
         logger,
         featureFlags: [
           { name: 'a', variant: '1' },
+          // @ts-expect-error
           { variant: 'b' },
```
imjoehaines, line 911: This has no `name`

packages/core/test/client.test.ts
```diff
@@ -943,6 +945,7 @@ describe('@bugsnag/core/client', () => {
         error: jest.fn()
       }
 
+      // @ts-expect-error
       const client = new Client({ apiKey: 'a123456789012345678901234567890b', logger, featureFlags })
```
imjoehaines, line 949: Here we're passing junk on purpose

djskinner (approved):

PR #1617, opened 2021-12-09
Allow `featureFlags` in ReactNative config

lemnik:
## Goal
Send `featureFlags` configured as part of `Bugsnag.start` to the `NativeClient` so that they can be included in all future events.

## Testing
Additional unit tests, and some manual testing using additional code to ensure the feature flags can be sent to a native client and will be included in delivered events.

### DIFF
--- a/packages/plugin-react-native-client-sync/test/client-sync.test.ts
+++ b/packages/plugin-react-native-client-sync/test/client-sync.test.ts
@@ -107,6 +107,74 @@ describe('plugin: react native client sync', () => {
       })
       c.leaveBreadcrumb('Spin', { direction: 'ccw', deg: '90' })
     })
+
+    describe('feature flags', () => {
+      it('adds individual feature flags', done => {
+        const c = new Client({
+          apiKey: 'api_key',
+          plugins: [
+            plugin({
+              addFeatureFlag: (name: string, variant?: string) => {
+                expect(name).toBe('feature flag')
+                expect(variant).toBe('flag variant')
+                done()
+              }
+            })
+          ]
+        })
+
+        c.addFeatureFlag('feature flag', 'flag variant')
+      })
+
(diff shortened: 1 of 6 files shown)


packages/react-native/src/notifier.js
```diff
@@ -80,6 +80,10 @@ const _createClient = (opts, jsOpts) => {
     bugsnag.setContext(opts.context)
   }
 
+  if (opts.featureFlags && opts.featureFlags !== opts._originalValues.featureFlags) {
+    NativeClient.addFeatureFlags(opts.featureFlags)
```
imjoehaines, line 84: Should this be `bugsnag.addFeatureFlags` with `@bugsnag/plugin-react-native-client-sync` handling the call to the native client?

lemnik: Yes it probably should be 👍

packages/plugin-react-native-client-sync/test/client-sync.test.ts
```diff
@@ -107,6 +107,78 @@ describe('plugin: react native client sync', () => {
...
+              clearFeatureFlags: () => {
+                done()
+              }
+            })
+          ]
+        })
+
+        c.clearFeatureFlags()
+      })
+    })
```
imjoehaines, line 181: ESLint doesn't like this because it doesn't technically have an assertion 😞

Since it's all synchronous, this could use a mock instead:

```js
it('clears all feature flags', () => {
  const clearFeatureFlags = jest.fn()
  const c = new Client({
    apiKey: 'api_key',
    plugins: [plugin({ clearFeatureFlags })]
  })

  c.clearFeatureFlags()
  expect(clearFeatureFlags).toHaveBeenCalledTimes(1)
})
```

imjoehaines (approved):
LGTM, just a listing issue to fix

PR #1618, opened 2021-12-10
Sync ReactNative feature flags on Android

lemnik:
## Goal
Bridge the feature flag related methods for the Android notifier, handling the
- `addFeatureFlag`
- `addFeatureFlags`
- `clearFeatureFlag`
- `clearFeatureFlags`

methods for the `NativeClient`. `addFeatureFlags` is broken down into multiple `addFeatureFlag(String, String?)` calls to avoid having to reconstruct the list of feature flags to be added, this also keeps the ReactNative-specific logic (`ReadableArray` references) in `bugsnag-js`.

These changes depend on [bugsnag-android#1552](https://github.com/bugsnag/bugsnag-android/pull/1552)

## Testing
Included unit test coverage for the new methods, and manually tested end-to-end with an updated Android notifier.


imjoehaines (approved):

packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
```diff
@@ -182,4 +183,49 @@ class BugsnagReactNative(private val reactContext: ReactApplicationContext) :
...
+    fun clearFeatureFlags() {
+        try {
+            plugin.clearFeatureFlags()
+        } catch (exc: Throwable) {
+            logFailure("clearFeatureFlags", exc)
+        }
+    }
+
+    private fun ReadableMap.safeString(key: String): String? {
+        return if(hasKey(key)) getString(key) else null
```
fractalwrench, line 229:
```suggestion
        return if (hasKey(key)) getString(key) else null
```

packages/react-native/android/src/main/java/com/bugsnag/android/BugsnagReactNative.kt
```diff
@@ -182,4 +183,49 @@ class BugsnagReactNative(private val reactContext: ReactApplicationContext) :
...
+            plugin.addFeatureFlag(name, variant)
+        } catch (exc: Throwable) {
+            logFailure("addFeatureFlag", exc)
+        }
+    }
+
+    @ReactMethod
+    fun addFeatureFlags(flags: ReadableArray) {
+        try {
+            for(index in 0 until flags.size()) {
```
fractalwrench, line 199:
```suggestion
            for (index in 0 until flags.size()) {
```

fractalwrench (approved):
LGTM pending CI + a few style nits

packages/plugin-react-native-client-sync/client-sync.js
```diff
@@ -98,6 +98,16 @@ module.exports = (NativeClient) => ({
...
+        case 'AddFeatureFlag':
+          origAddFeatureFlag.call(client, event.data.name, event.data.variant)
+          break
+        case 'ClearFeatureFlag':
+          if (event.data && event.data.name) {
+            origClearFeatureFlag.call(client, event.data.name)
+          } else {
+            origClearFeatureFlags.call(client)
+          }
+          break
```
nickdowell, line 110: Why are feature flags being synced with distinct add and clear events rather than following the approach used for metadata where a single 'update' event is used to update everything at once?

lemnik: This was mostly because unlike metadata, feature flags are "write only". So while they can be changed, there is no API-level way to read them back. Adding an internal-only read API on the native notifiers would solve this though (and probably make a few other things simpler at the same time).

nickdowell: Thanks for the explanation!

fractalwrench (approved):
LGTM pending resolution of conflicts

PR #1643, opened 2022-01-11
Recursive metadata is now handled in react native

lemnik:
## Goal
Make it safe to specify complex and recursive objects as metadata for breadcrumbs and events in react native applications.

## Design
Added a new `makeSafe` function to the `delivery-react-native` which behaves similarly to the `safe-json-stringify` module but specifically for the react native JS bridge, which does not invoke `toJSON` functions.

## Testing
New unit tests to cover the `makeSafe` function, and adjustments to the existing Mazerunner tests to ensure circular references are correctly flattened and don't cause a crash.

### DIFF
--- a/packages/delivery-react-native/make-safe.js
+++ b/packages/delivery-react-native/make-safe.js
@@ -1,0 +1,83 @@
+const isArray = require('@bugsnag/core/lib/es-utils/is-array')
+
+const isSafeLiteral = (obj) => (
+  typeof obj === 'string' || obj instanceof String ||
+  typeof obj === 'number' || obj instanceof Number ||
+  typeof obj === 'boolean' || obj instanceof Boolean
+)
+
+const isError = o => (
+  o instanceof Error || /^\[object (Error|(Dom)?Exception)]$/.test(Object.prototype.toString.call(o))
+)
+
+const throwsMessage = err => '[Throws: ' + (err ? err.message : '?') + ']'
+
+const safelyGetProp = (obj, propName) => {
+  try {
+    return obj[propName]
+  } catch (err) {
+    return throwsMessage(err)
+  }
+}
+
+/**
+ * Similar to `safe-json-stringify` this function rebuilds an object graph to be safe for the ReactNative JS Bridge.
+ * This requirement is different to `JSON.parse(safeJsonStringify(data))` in three key ways:
+ * - `toJSON` methods are not called
+ * - we leave more types intact for the JS Bridge to handle
+ * - there is no redaction or fixed depth limit
+ *
+ * @param data the value to be made safe for the ReactNative bridge
+ * @returns a safe version of the given `data`
+ */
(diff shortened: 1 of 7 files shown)


packages/delivery-react-native/make-safe.js
```diff
@@ -0,0 +1,79 @@
...
+      return visit({ name: obj.name, message: obj.message })
+    }
+
+    if (seen.includes(obj)) {
+      // circular references are replaced and marked
+      return '[Circular]'
+    }
+
+    // handle arrays, and all iterable non-array types (such as Set)
+    if (isArray(obj) || obj[Symbol.iterator]) {
```
imjoehaines, line 53: Do all versions of RN support `Symbol`?

I can't really find anything very helpful, but there's [this error from not having a `Symbol` global on SO](https://stackoverflow.com/questions/50504532/react-native-cant-find-variable-symbol) which may mean we need to be careful about using `Symbol`

Something like this might be safer:

```suggestion
    if (isArray(obj) || (Symbol && obj[Symbol.iterator])) {
```

test/react-native/features/fixtures/app/scenario_js/app/scenarios/BreadcrumbsJsManualScenario.js
```diff
@@ -6,6 +6,10 @@ export class BreadcrumbsJsManualScenario extends Scenario {
     const metaData = {
       from: 'javascript'
     }
+
+    // ensure that circular references are safely handled
+    metaData.circle = metaData
```
imjoehaines, line 11: Can we add this to [the expected breadcrumb](https://github.com/bugsnag/bugsnag-js/blob/250045158bea921665d51c0383d60a50cbc78e9d/test/react-native/features/fixtures/expected_breadcrumbs/JsManualScenario.json)?

packages/delivery-react-native/test/make-safe.test.ts
```diff
@@ -0,0 +1,86 @@
+import makeSafe from '../make-safe'
+
+describe('delivery: react native makeSafe', () => {
+  it('leaves simple types intact', () => {
```
imjoehaines, line 4: Can we add more types to this test? e.g. `null` & `undefined` (as they are special cased), `Symbol`, maybe a `Map` or other non-array iterable?

packages/delivery-react-native/make-safe.js
lemnik: `Symbol` is available on all the RN versions we support, but I'm happy to addd the secondary check for safety sake.

imjoehaines: If it's supported everywhere then it's probably fine as-is, I just couldn't find anything definitive in a quick search

kattrali (approved):

imjoehaines (approved):

PR #1648, opened 2022-01-13
Fix failing filestore test

pietgk:
Several test failed because storage fixtures directory is not created.
 FAIL  packages/electron-filestore/test/filestore.test.ts
    ENOENT: no such file or directory, stat '/Users/grop/dev/bugsnag-js/packages/electron-filestore/test/fixtures/storage'

## Goal

to get test green.

## Design

create correct test context before running the tests by creating fixtures/storage dir.

## Changeset

filestore.test.ts with test fix

## Testing

test itself is fixed and was returning errors and now succeeds

### DIFF
--- a/packages/electron-filestore/test/filestore.test.ts
+++ b/packages/electron-filestore/test/filestore.test.ts
@@ -12,6 +12,7 @@ describe('FileStore', () => {
   beforeEach(async () => {
     await mkdir(join(crashes, 'reports'), { recursive: true })
     await mkdir(join(crashes, 'pending'), { recursive: true })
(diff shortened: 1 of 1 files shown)


johnkiely1:
Hi @pietgk. 

Thanks for the PR. It would be interesting to know under what conditions you see the test failures?  How are you running the test? Are you using an IDE? 

`store.init()` should create the directory so it should always exist unless so we wouldn't really expect this change to be necessary.

luke-belton:
hey @pietgk - we've managed to reproduce the failing test now. We're going to investigate some more and will keep you updated 👍

PR #1652, opened 2022-01-18
Ensure unhandled JS errors crash the app in e2e test scenarios

twometresteve:
## Goal

Ensure that unhandled JS errors crash the app in the React Native end-to-end test scenarios.

## Design

The click handler for the `run_scenario` button was `async`, leading to unhandled JS errors behaving the same as unhandled Promise rejections and the app not crashing.  The Cucumber steps that follow assume the app crashed and tell it to relaunch, which is likely to have caused a number of flakes.

## Changeset

- Refactors the test fixtures to ensure that scenarios are not run `async`.  This led to separating the Start Bugsnag and Run Scenario steps.
- Renamed the Cucumber step to make it clear that the app (should have) crashed, matching what bugsnag-cocoa now does.
- Updated the Unhandled Promise rejection scenario to cater for the app not crashing (it is not expected to).
- Added a mechanism to wait for Appium to report the app is not running after a crash before relaunching.  This isn't perfect as sometimes it seems not to report as not_running.  Further investigation is needed, but it seems to relate to running scenarios together, as it doesn't happen if running the scenario by itself.

## Testing

Covered by CI.

### DIFF
--- a/test/react-native/features/fixtures/app/react_native_navigation_js/index.js
+++ b/test/react-native/features/fixtures/app/react_native_navigation_js/index.js
@@ -53,6 +53,19 @@ export default class AppScreen extends Component {
     const jsConfig = defaultJsConfig()
     const scenario = new Scenarios[scenarioName](configuration, scenarioMetaData, jsConfig)
     console.log(`  with config: ${JSON.stringify(configuration)} (native) and ${jsConfig} (js)`)
+    this.state.scenario = scenario
+    scenario.run()
+  }
+
+  startBugsnag = () => {
+    console.log(`Starting Bugsnag for scenario: ${this.state.currentScenario}`)
+    console.log(`  with MetaData: ${this.state.scenarioMetaData}`)
+    const scenarioName = this.state.currentScenario
+    const scenarioMetaData = this.state.scenarioMetaData
+    const configuration = this.getConfiguration()
+    const jsConfig = defaultJsConfig()
+    const scenario = new Scenarios[scenarioName](configuration, scenarioMetaData, jsConfig)
+    console.log(`  with config: ${JSON.stringify(configuration)} (native) and ${jsConfig} (js)`)
     NativeModules.BugsnagTestInterface.startBugsnag(configuration)
       .then(() => {
         Navigation.setRoot({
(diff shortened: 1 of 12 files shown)


test/react-native/features/steps/react-native-steps.rb
```diff
@@ -1,68 +1,99 @@
...
+  # max_attempts = 300
+  max_attempts = 100
+  attempts = 0
+  assertion_passed = false
+  until (attempts >= max_attempts) || assertion_passed
+    attempts += 1
+    assertion_passed = yield
+    sleep 0.1
+  end
+  # raise 'Assertion not passed in 30s' unless assertion_passed
```
twometresteve, line 11: I left this in by mistake, but will fix on my next branch (in progress now) to avoid a whole new build, if possible.

lemnik (approved):
LGTM

test/react-native/features/fixtures/app/react_native_navigation_js/index.js
```diff
@@ -53,25 +53,21 @@ export default class AppScreen extends Component {
...
-        })
-        Bugsnag.start(jsConfig)
-        this.state.scenario = scenario
-        scenario.run()
-      })
+          ]
+        }
+      }
+    })
+    this.state.scenario = scenario
```
imjoehaines, line 69: This should probably use `setState`, otherwise the app won't re-render

test/react-native/features/steps/react-native-steps.rb
```diff
@@ -1,68 +1,99 @@
...
+    And I relaunch the app after a crash
   }
 end
 
-When("I relaunch the app") do
-  # This step should only be used when the app has crashed, but the notifier needs a little
-  # time to write the crash report before being forced to reopen.  From trials, 2s was not enough.
-  sleep(5)
+Then('the app is not running') do
+  wait_for_true do
```
imjoehaines, line 32: I think this could be:

```ruby
Maze::Wait.until do
  # stuff ...
end
```

then you don't need `wait_for_true`

test/react-native/features/fixtures/app/react_native_navigation_js/index.js
twometresteve: Yeah, I noticed that some test fixtures used setState and others not.  I fairly lazily left this as-is, largely because I'd already had a few attempts at getting the build to pass - but I'll note it to look at. All the iOS navigation tests are currently skipped and I guess this could be a factor (but if nothing else we could make it consistent).

test/react-native/features/steps/react-native-steps.rb
twometresteve: Good thinking, I'll factor that in when I take the next look at it.

PR #1654, opened 2022-01-18
Avoid jcenter where possible

lemnik:
## Goal
Remove `jcenter` as a dependency when building on CI by replacing it with the Bugsnag private Maven repository.

## Changes

- Replaced `jcenter` with `mavenCentral` for buildscript level repositories
- If available: use the new `MAVEN_REPO_URL` and `MAVEN_REPO_CREDS` environment variables to declare a Maven repository instead of `jcenter`
- Pass the `MAVEN_REPO_URL` and `MAVEN_REPO_CREDS` as environment variables to the BuildKite docker containers

## Testing
Relied on the existing tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -154,6 +154,7 @@ services:
         - REGISTRY_URL
         - REG_BASIC_CREDENTIAL
         - REG_NPM_EMAIL
+        - MAVEN_REPO_URL
     environment:
       - DEBUG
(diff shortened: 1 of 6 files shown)


twometresteve (approved):
Looks ok to me and doesn't seem to do any harm given that the build still succeeded - I just can't see a way of telling for sure that our Maven repo is being used, which would provide additional comfort.

PR #1672, opened 2022-01-24
Force DEVELOPER_DIR for pipeline steps

joshedney:
## Goal

Set the `DEVELOPER_DIR` for each step as required so that we set which version of Xcode we want to use.

## Changeset

Add the environment variable `DEVELOPER_DIR` for each step that requires it.

## Testing

Automated Tests have been run.


.buildkite/react-native-cli-pipeline.yml
```diff
@@ -282,7 +282,7 @@ steps:
     env:
       DEBUG: true
       LANG: "en_US.UTF-8"
-      DEVELOPER_DIR: "/Applications/Xcode11.7.app"
+      DEVELOPER_DIR: "/Applications/Xcode11.app"
```
twometresteve, line 285: It's probably not a problem if the build works, but I notice that RN 0.63 builds with Xcode 12 in the RN CLI pipeline.

twometresteve (approved):
LGTM pending successful CI.

.buildkite/react-native-cli-pipeline.yml
joshedney: Hmm the only reason that this would have been added is because that it was building on the old ms-mac-10-15 queue. I'll double check.

joshedney: This step used to be on the `opensource-mac-rn` queue which was the old 10-15 macs

twometresteve: Yeah, there's a fair bit of history in these pipelines.  If they both build I think it's fine.

PR #1673, opened 2022-01-24
Handle recursive metadata in ReactNative

lemnik:
## Goal
Re-introduce the changes made in [#1643](https://github.com/bugsnag/bugsnag-js/pull/1643) with fixes to the tests, and an additional `makeSafe` call in `plugin-react-native-client-sync` to ensure that metadata included in Breadcrumbs is also safe to cross the React Native JS Bridge.

Make it safe to specify complex and recursive objects as metadata for breadcrumbs and events in react native applications.

## Design
Added a new `makeSafe` function to the `delivery-react-native` which behaves similarly to the `safe-json-stringify` module but specifically for the react native JS bridge, which does not invoke `toJSON` functions.

## Testing
New unit tests to cover the `makeSafe` function, and adjustments to the existing Mazerunner tests to ensure circular references are correctly flattened and don't cause a crash.


packages/plugin-react-native-client-sync/client-sync.js
```diff
@@ -1,4 +1,5 @@
 const { DeviceEventEmitter, NativeEventEmitter, NativeModules, Platform } = require('react-native')
+const makeSafe = require('@bugsnag/delivery-react-native/make-safe')
```
imjoehaines, line 2: I don't think it makes sense for this to import something from `delivery-react-native`

Should `make-safe` be a new plugin? We could put it in core, but that would increase the browser bundle size, so maybe it could go in `@bugsnag/react-native`?

imjoehaines: It'd be good to give it a better name now it's being used in multiple places as well; maybe something like "resolve recursive references" (or something less alliterative!)

imjoehaines (changes requested):

packages/plugin-react-native-client-sync/client-sync.js
lemnik: I mostly agree with moving it out of `delivery`, but moving it into `@bugsnag/react-native` leads us to a circular dependency. Given how tied to the react-native bridge this logic is, it should definitely be kept away from `core`.

Also agree on changing the name, maybe `bridgeSafe`?

Alternatively it needs to become more generic and move completely to its own module?

imjoehaines: > Given how tied to the react-native bridge this logic is&hellip;

I'm not sure it is tied to the react-native bridge; it just resolves recursive references in any JS value, it isn't RN specific AFAICT

lemnik: In the abstract, yes. On the other hand it also converts `Date` objects to strings, changes how `Error` objects are represented, etc. while those are probably useful behaviours everywhere, they're also baked into the logic right now.

imjoehaines (approved):
LGTM, but could use a changelog entry

PR #1678, opened 2022-01-28
Miscellaneous CI improvements

twometresteve:
## Goal

A few miscellaneous CI pipeline improvements:
- Allow retries on known-flaky RN CLI tests, to reduce frustration for developers while a fix is being worked on.
- Bump Android to 11 for RN tests in the hope of quicker/more stable runs.
- Increase maximum wait to receive errors to 60 seconds - the theoretical maximum for the Android notifier.
- Remove use of the resilient Appium driver.  It didn't work as intended and only confused things in the case of a genuine problem.

The retry mechanism in Maze Runner is currently broken, but will be fixed shortly.

## Testing

Covered by CI.

### DIFF
--- a/test/react-native/features/support/env.rb
+++ b/test/react-native/features/support/env.rb
@@ -1,6 +1,6 @@
 BeforeAll do
-  Maze.config.receive_no_requests_wait = 15
-  Maze.config.receive_requests_wait = 60 # Theoretical maximum on Android
+  Maze.config.receive_no_requests_wait = 30
+  Maze.config.receive_requests_wait = 30
 end
 
 Before('@android_only') do |scenario|
--- a/test/react-native/features/support/env.rb
+++ b/test/react-native/features/support/env.rb
@@ -1,5 +1,6 @@
 BeforeAll do
-  Maze.config.receive_no_requests_wait = 15 if Maze.config.respond_to? :receive_no_requests_wait=
+  Maze.config.receive_no_requests_wait = 15
+  Maze.config.receive_requests_wait = 60 # Theoretical maximum on Android
 end
 
 Before('@android_only') do |scenario|
--- a/.buildkite/expo-pipeline.yml
+++ b/.buildkite/expo-pipeline.yml
@@ -114,7 +114,6 @@ steps:
           - --a11y-locator
           - --fail-fast
           - --retry=2
-          - --resilient
     concurrency: 9
     concurrency_group: 'browserstack-app'
     concurrency_method: eager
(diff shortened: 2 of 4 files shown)


test/react-native/features/support/env.rb
```diff
@@ -1,5 +1,6 @@
 BeforeAll do
-  Maze.config.receive_no_requests_wait = 15 if Maze.config.respond_to? :receive_no_requests_wait=
+  Maze.config.receive_no_requests_wait = 15
+  Maze.config.receive_requests_wait = 60 # Theoretical maximum on Android
```
imjoehaines, line 3: Will it ever actually take 60 seconds to deliver events? That seems like way too long to me 😬

If that can legitimately happen, I think the no request wait needs to match it. Otherwise tests that don't expect any events could pass before the notifier has actually tried to deliver them

```suggestion
  Maze.config.receive_no_requests_wait = 60
  Maze.config.receive_requests_wait = 60 # Theoretical maximum on Android
```

twometresteve: Well hopefully it wouldn't ever take 60 seconds, but we did find it made some difference in the Android pipeline (and we could see from logs that the notifier was still trying to send a request at the 30s mark).  

With the `receive_requests_wait` setting, it will only be reached if the scenario is going to fail anyway - i.e. if all requests are received within 3 seconds, say, then Maze Runner will carry on immediately.  With `receive_no_requests_wait`, Maze Runner will _always_ wait for that period of time.

I'd argue that we don't need them to match.  Increasing the wait timeout is there to deal with occasional flakes, not routine behaviour.  In the case of no_wait it's feasible that a request could come in after that period, but most of the time if a request is coming it'll happen quickly, fail the test and we'll immediately look into it (as opposed to blindly retrying when only 3 of 4 requests are received, for example).

That all said, there are only two instances of the `I should receive no requests` step in use in the React Native pipeline.  So we could increase it if you felt the benefit was worth adding 1.5 minutes to every Buildkite step.

imjoehaines: I see what you're saying, but it still feels a bit off to me that we allow events to be sent after a relatively long time except for when we don't expect anything to be sent

How about leaving both options at the 30s default?

twometresteve: OK.  To be fair I was basing the change on previous experience and don't have hard evidence that changing it to 60s would make a difference.  Perhaps I'll monitor flakes going forward and if we can see that the app is shut down while the notifier is still trying to send then we can reassess.

imjoehaines (approved):

PR #1679, opened 2022-01-28
Automatically retry the Electron performance test

imjoehaines:
## Goal

As this runs on GH Actions, it's inherently unreliable as we could be given a very over-provisioned VM so the performance is unpredictable

Until we can migrate to a more stable platform, we can only retry this test until it passes — there is some value in knowing it will pass when retried, rather than always failing

I'd like to avoid increasing the allowed time any further — 150ms is already a lot and is just a compromise to allow the test to pass on CI (locally this test runs _significantly_ quicker)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,6 @@
     "local-npm:start": "verdaccio --config test/electron/local-npm-config.yml --listen 0.0.0.0:5539",
     "local-npm:publish-all": "lerna publish \"$VERSION_IDENTIFIER\" --yes --force-publish --exact --no-push --no-git-reset --no-git-tag-version --registry 'http://0.0.0.0:5539'",
     "local-npm:publish-all-win32": "lerna publish %VERSION_IDENTIFIER% --yes --force-publish --exact --no-push --no-git-reset --no-git-tag-version --registry 'http://0.0.0.0:5539'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -98,6 +98,6 @@
     "local-npm:start": "verdaccio --config test/electron/local-npm-config.yml --listen 0.0.0.0:5539",
     "local-npm:publish-all": "lerna publish \"$VERSION_IDENTIFIER\" --yes --force-publish --exact --no-push --no-git-reset --no-git-tag-version --registry 'http://0.0.0.0:5539'",
     "local-npm:publish-all-win32": "lerna publish %VERSION_IDENTIFIER% --yes --force-publish --exact --no-push --no-git-reset --no-git-tag-version --registry 'http://0.0.0.0:5539'",
-    "test:electron": "xvfb-maybe --auto-servernum -- cucumber-js test/electron/features"
+    "test:electron": "xvfb-maybe --auto-servernum -- cucumber-js test/electron/features/performance.feature --retry 3 --retry-tag-filter @flaky"
```
twometresteve, line 101: I don't think you meant to put `performance.feature` on the end?  No other features will be run.

imjoehaines: 🤦

twometresteve (approved):
LGTM pending CI.

PR #1690, opened 2022-02-21
bump uglify-js

djskinner:
## Goal

- Reduction in bundle size

## Changeset

- Bump uglify-js for 3.15.1 (latest) and update the renamed `--ie` flag.

## Testing

E2E tests passing across all supported browsers

### DIFF
--- a/bin/minify
+++ b/bin/minify
@@ -3,6 +3,6 @@
 cat - | ../../node_modules/.bin/uglifyjs \
   --compress \
   --mangle \
-  --ie8 \
+  --ie \
   --source-map "includeSources,content=inline,url=$(basename "$1").map" \
   --output "$1"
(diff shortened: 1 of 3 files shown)


imjoehaines (approved):

djskinner:
@imjoehaines should this have a change log entry like `- (browser): Update uglify-js [#1690](https://github.com/bugsnag/bugsnag-js/pull/1690)`? I wasn't sure

imjoehaines:
> @imjoehaines should this have a change log entry like `- (browser): Update uglify-js [#1690](https://github.com/bugsnag/bugsnag-js/pull/1690)`? I wasn't sure

This shouldn't make any noticeable difference to users (other than the bundle size reduction), so I don't think it needs one

PR #1696, opened 2022-03-07
Add .ts files to bugsnag/core package

qali-clio:
## Goal

We use `@bugsnag/core` in our project, and use the `onBreadcrumb` option as per [the docs](https://docs.bugsnag.com/platforms/javascript/customizing-breadcrumbs/) to filter and modify breadcrumb metadata. In order to test this behaviour, we need to create `Breadcrumb` instances with specific metadata, however the type definitions in the repo don't include the constructor.

This PR adds the appropriate type definitions to `Breadcrumb`.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,7 +15,8 @@
   "files": [
     "lib",
     "types",
-    "*.js"
+    "*.js",
+    "*.ts"
   ],
   "author": "Bugsnag",
   "license": "MIT",


yousif-bugsnag:
Hi @qali-clio, thanks for the PR - we deliberately do not include the constructor as we don't expect anything else to be instantiating breadcrumbs. 

However we do have a derived class called `BreadcrumbWithInternals` that does define that constructor so that should give you what you need: https://github.com/bugsnag/bugsnag-js/blob/master/packages/core/breadcrumb.d.ts

qali-clio:
@yousif-bugsnag ah, gotcha! it looks like `bugsnag/core` does not include .ts files in the project root, so we don't have that file installed. I've updated the PR to add `*.ts` files to the included files list instead.

xander-jones:
Hey @qali-clio, some TS files are purposely not added the the `files` property of the `package.json` (other than a few exceptions like the [types](https://github.com/bugsnag/bugsnag-js/tree/master/packages/core/types) directory) as they just contain implementation details. For this reason, I don't think this is a PR we'd be looking to accept.

However, a couple of other solutions that might work for you in testing: Creating and then assigning class properties after:

```ts
const a = new Breadcrumb()

a.message = 'hello'
a.metadata = { stuff: [1, 2, 3] }
a.type = 'manual'
a.timestamp = new Date()
```

or, extending the original `Breadcrumb` class:

```ts
class TestBreadcrumb extends Breadcrumb {
   constructor (message: Breadcrumb["message"], metadata: Breadcrumb["metadata"], type: Breadcrumb['type'], timestamp: Breadcrumb["timestamp"]) {
      super()

      this.message = message
      this.metadata = metadata
      this.type = type
      this.timestamp = timestamp
   }
}

const b = new TestBreadcrumb("hello v2", { stuff: [4, 5, 6] }, 'manual', new Date())
```

Let us know if these would work for you?

qali-clio:
it would be nice to be able to create Breadcrumbs for test purposes, but we have a similar workaround in our repo right now. thanks!

PR #1702, opened 2022-03-10
chore(plugin-express): send body in event.request

djskinner:
## Goal

When using the express plugin send the body (if present) as `event.request.body` rather than `event.metadata.request.body` for better alignment with the notifier spec. This should not have any user-facing impact.

## Design

Better alignment with the spec

## Changeset

- send the body as `event.request.body` rather than `event.metadata.request.body`

## Testing

- Covered by an E2E test

### DIFF
--- a/packages/plugin-express/src/express.js
+++ b/packages/plugin-express/src/express.js
@@ -75,10 +75,11 @@ module.exports = {
 }
 
 const getRequestAndMetadataFromReq = req => {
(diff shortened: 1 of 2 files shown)


imjoehaines (approved):
Can you add a changelog entry for this & https://github.com/bugsnag/bugsnag-js/pull/1701?

PR #1708, opened 2022-03-23
React-native test fixture updates

Cawllec:
## Goal

Update the tested versions of react-native to ensure we have coverage up to the newest release of react-native.  0.60 is still our baseline and will have full tests run against it.  Intermediate versions being removed will be covered in the react-native-cli pipeline.


twometresteve (approved):
LGTM

PR #1713, opened 2022-04-01
Remove incorrect type definition

adkenyon:
## Goal

The type definition for `client.notify` method implies that several types are acceptable. Which isn't exactly true. When using an `Error` type `client.notify` works as expected. When using a `string` type, the error reported is `notify() received a non-error. See "notify()" tab for more detail.` while the error details are on the `notify` tab. This is confusing and does not work as expected.

## Design

I removed the `NotifiableError` type in favor of `Error`. This is purely a type definition change with **no** change in functionality.

## Changeset

The `NotifiableError` type was removed in favor of `Error`.

## Testing

I'm having issues building/testing the repo as is. I was able to run the typescript check command with 400+ errors. After my changes, the typescript check did not register any new errors.

### DIFF
--- a/packages/core/types/client.d.ts
+++ b/packages/core/types/client.d.ts
@@ -1,6 +1,5 @@
 import Breadcrumb from './breadcrumb'
 import {
-  NotifiableError,
   BreadcrumbType,
   OnErrorCallback,
   OnSessionCallback,
@@ -16,7 +15,7 @@ declare class Client {
 
   // reporting errors
   public notify(
-    error: NotifiableError,
+    error: Error,
     onError?: OnErrorCallback,
     cb?: (err: any, event: Event) => void
   ): void;
--- a/packages/core/types/common.d.ts
+++ b/packages/core/types/common.d.ts
@@ -73,11 +73,6 @@ export interface SessionPayload {
   sessions: Session[]
 }
 
-export type NotifiableError = Error
-| { errorClass: string, errorMessage: string }
-| { name: string, message: string }
-| string
-
 export type BreadcrumbType = 'error' | 'log' | 'manual' | 'navigation' | 'process' | 'request' | 'state' | 'user';
 
 interface Device {


xander-jones:
Hey @adkenyon, I do agree, an `Error` object should always be the preference for making a call to `Bugsnag.notify()`, especially in `Promise`s so that we have a known object getting sent in (https://eslint.org/docs/rules/prefer-promise-reject-errors). However, I don't think this is a PR we'd be looking to accept as in some cases where an `Error` is not thrown, we still want a notification to be sent, e.g. legacy browsers, packages & code, or even just mistakes where a non-`Error` is tolerable, even if it's not perfect.

When `notify` is called, an `Event` is created:
https://github.com/bugsnag/bugsnag-js/blob/58c0ade57fa1ada78ce3dd912022f3e15b326fc7/packages/core/client.js#L286

When creating the event we make a call to normalize the "error":
https://github.com/bugsnag/bugsnag-js/blob/8bae662a32f939de7371ce542989d2dbed84f03b/packages/core/event.js#L163

https://github.com/bugsnag/bugsnag-js/blob/8bae662a32f939de7371ce542989d2dbed84f03b/packages/core/event.js#L193-L265

If we made the change you're suggesting wouldn't break, but it would limit functionality for anyone using TS.

In your case, where you're seeing `notify() received a non-error. See "notify()" tab for more detail.`, this sounds like a situation where you needed to specify an `Error` but did not, i.e. this specific message you saw was context specific.

adkenyon:
@xander-jones thanks for the quick response!

One point I'd like to convey is the frustration this type definition continues to cause us. Passing an `Error` type compared to a non-Error type results in two very different results on BugSnag. This last week we shipped an app release that passed a `String` type to `.notify()` and the data in BugSnag isn't useful or actionable. We've having to update the code and create another app release. We've effectively wasted our time and lost a week of data. By passing an `Error` type, the data is much more useful in Bugsnag and actionable. This is something also comes up every few months, a dev thinks that a `String` type will work as expected when in actuality it doesn't. 

I see your point on that this may limit functionality, but I'm not sure the functionality is desired? We've never found it useful or actionable

xander-jones:
Hey – thanks for the added context! This has struck up a bit of a discussion internally. Some more thought is needed on how we'd like to proceed on this, so at the moment we won't be accepting the PR directly, but I'll leave this open for future updates on this area.

adkenyon:
@xljones no worries! Thanks for considering it and happy to spur some discussion. This is all purely my team's experience and perspective. Understand ya'll have many more customers, perspectives, and tasks.

gareththackeray:
Hi @adkenyon.  Could you confirm that you simply passed called `Bugsnag.notify('some string')` and got `notify() received a non-error. See "notify()" tab for more detail.` as the message?  
I've tried to reproduce this and can't.  You should only get that message under certain circumstances like unhandled rejections.

PR #1718, opened 2022-04-11
Trigger expo pipeline on push

imjoehaines:
## Goal

Adds a trigger step for bugsnag-expo to the pipeline. This passes over the branch & commit so that the bugsnag-expo tests can be run against any bugsnag-js changes


twometresteve (approved):
LGTM

PR #1731, opened 2022-05-04
Add missing browsers to E2E tests

joshedney:
## Goal

Add missing supported browsers to the E2E within the testing pipeline for additional coverage.

## Changeset

Add chrome 30 browser to the E2E tests
Add Android 4.4, 5.0 and 6.0 browsers to the E2E tests

## Testing

Covered by `[full ci]`


.buildkite/browser-pipeline.yml
```diff
@@ -16,6 +16,111 @@ steps:
       - docker-compose#v3.9.0:
           push:
             - browser-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:ci-browser-${BRANCH_NAME}
+#
+  - label: ':chrome: v30 Browser tests'
```
twometresteve, line 20: I don't think we want to add all these version of Chrome.  Adding them to Maze Runner was just in case we found a problem with v30, but if 30 passes then I would remove 32 to 42.

.buildkite/browser-pipeline.yml
```diff
@@ -231,6 +336,68 @@ steps:
...
+        verbose: true
+        command:
+          - --farm=bs
+          - --browser=iphone_13
+    env:
+      HOST: "bs-local.com"
+    concurrency: 5
+    concurrency_group: 'browserstack'
+
+  - label: ':android: Google Nexus 5 Browser tests'
```
twometresteve, line 356: Could we called these "Android X browser tests" rather than after the device model.  It's the Android version that we care about.

Cawllec (approved):

PR #1737, opened 2022-05-17
[PLAT-8308] Add isStarted to notifiers

gingerbenw:
## Goal

Add the `isStarted` property to notifiers similar to the current android implementation, to check whether Bugsnag has been initialised

## Changeset

Added `isStarted` to the following notifiers

- browser
- node
- react-native
- electron

## Testing

Added appropriate unit tests for each notifier


gareththackeray (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Added
+
+- Added `isStarted()` to notifiers to check whether Bugsnag has initialized [#1737](https://github.com/bugsnag/bugsnag-js/pull/1737)
```
imjoehaines, line 7: Maybe worth being a little more explicit here:

```suggestion
- Added `Bugsnag.isStarted()` to check whether Bugsnag has initialized [#1737](https://github.com/bugsnag/bugsnag-js/pull/1737)
```

packages/browser/test/index.test.ts
```diff
@@ -151,4 +151,12 @@ describe('browser notifier', () => {
...
     })
   })
+
+  it('indicates whether or not the client is started', (done) => {
+    const Bugsnag = getBugsnag()
+    expect(Bugsnag.isStarted()).toBe(false)
+    Bugsnag.start(API_KEY)
+    expect(Bugsnag.isStarted()).toBe(true)
+    done()
+  })
```
imjoehaines, line 161: The `done` function is only needed when writing an asynchronous test that doesn't use promises, so it's not necessary here

```suggestion
  it('indicates whether or not the client is started', () => {
    const Bugsnag = getBugsnag()
    expect(Bugsnag.isStarted()).toBe(false)
    Bugsnag.start(API_KEY)
    expect(Bugsnag.isStarted()).toBe(true)
  })
```

imjoehaines (commented):
The code changes LGTM

I don't think we need both `packages/react-native/src/test/notifier.test.ts` and `packages/react-native/test/index.test.ts` though, since they're both testing the same thing

I'd suggest getting rid of `packages/react-native/src/test/notifier.test.ts` as the other test has a lot more cleanup so there's less likely for issues after calling `start` multiple times

PR #1738, opened 2022-05-17
[PLAT-7801] Add early check for undefined apiKey

gingerbenw:
## Goal

To improve error message shown to end user when the apiKey is undefined/null

## Design

We were trying to use the apiKey to create a file store before validating it

## Changeset

Added early check to `electron/src/client/main.js`

## Testing

Added unit test for specific test case

### DIFF
--- a/packages/electron/src/client/main.js
+++ b/packages/electron/src/client/main.js
@@ -16,6 +16,11 @@ const { schema } = require('../config/main')
 Event.__type = 'electronnodejs'
 
 module.exports = (opts) => {
+  // Sanity check api key has been provided
+  if (typeof opts.apiKey !== 'string') {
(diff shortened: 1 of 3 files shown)


packages/electron/src/client/main.js
```diff
@@ -16,6 +16,11 @@ const { schema } = require('../config/main')
 Event.__type = 'electronnodejs'
 
 module.exports = (opts) => {
+  // Sanity check api key has been provided
```
imjoehaines, line 19: Might be worth noting why we need to do this here but don't in other notifiers, e.g.:

```suggestion
  // check api key has been provided as we need it to create the FileStore
  // which happens before the API key is validated
```

PR #1739, opened 2022-05-17
[PLAT-8277] Add maxEvents error message and update tests

gingerbenw:
## Goal

To better clarify the reason for cancelled events when using `plugin-simple-throttle`

## Changeset

Added a logger message in `throttle.load` when calling `client.addOnError` 

## Testing

Updated unit tests to check number of throttled calls and message wording

### DIFF
--- a/packages/plugin-simple-throttle/test/throttle.test.ts
+++ b/packages/plugin-simple-throttle/test/throttle.test.ts
@@ -3,11 +3,21 @@ import plugin from '../'
 import Client from '@bugsnag/core/client'
 
 describe('plugin: throttle', () => {
-  it('prevents more than maxEvents being sent', () => {
+  describe('addOnError', () => {
     const payloads = []
     const c = new Client({ apiKey: 'aaaa-aaaa-aaaa-aaaa' }, undefined, [plugin])
+    const mockWarn = jest.fn(message => message)
+    c._logger.warn = mockWarn
     c._setDelivery(client => ({ sendEvent: (payload) => payloads.push(payload), sendSession: () => {} }))
-    for (let i = 0; i < 100; i++) c.notify(new Error('This is fail'))
-    expect(payloads.length).toBe(10)
+
+    it('prevents more than maxEvents being sent', () => {
+      for (let i = 0; i < 100; i++) c.notify(new Error('This is fail'))
+      expect(payloads.length).toBe(10)
+      expect(mockWarn).toHaveBeenCalledTimes(90)
+    })
+
+    it('throws an appropriate error message', () => {
+      expect(mockWarn.mock.results[11].value).toBe('Cancelling event send due to maxEvents per session limit of 10 being reached')
+    })
   })
 })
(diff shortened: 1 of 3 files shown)


packages/plugin-simple-throttle/test/throttle.test.ts
```diff
@@ -3,11 +3,21 @@ import plugin from '../'
...
-    expect(payloads.length).toBe(10)
+
+    it('prevents more than maxEvents being sent', () => {
+      for (let i = 0; i < 100; i++) c.notify(new Error('This is fail'))
+      expect(payloads.length).toBe(10)
+      expect(mockWarn).toHaveBeenCalledTimes(90)
+    })
+
+    it('throws an appropriate error message', () => {
+      expect(mockWarn.mock.results[11].value).toBe('Cancelling event send due to maxEvents per session limit of 10 being reached')
```
imjoehaines, line 20: I think you can simplify this by using [`toHaveBeenCalledWith`](https://jestjs.io/docs/26.x/expect#tohavebeencalledwitharg1-arg2-), e.g.:

```suggestion
      expect(mockWarn).toHaveBeenCalledWith('Cancelling event send due to maxEvents per session limit of 10 being reached')
```

packages/plugin-simple-throttle/test/throttle.test.ts
```diff
@@ -3,11 +3,21 @@ import plugin from '../'
 import Client from '@bugsnag/core/client'
 
 describe('plugin: throttle', () => {
-  it('prevents more than maxEvents being sent', () => {
+  describe('addOnError', () => {
     const payloads = []
     const c = new Client({ apiKey: 'aaaa-aaaa-aaaa-aaaa' }, undefined, [plugin])
+    const mockWarn = jest.fn(message => message)
```
imjoehaines, line 9: I don't think you need to provide a function here

```suggestion
    const mockWarn = jest.fn()
```

packages/plugin-simple-throttle/test/throttle.test.ts
```diff
@@ -3,11 +3,21 @@ import plugin from '../'
...
-    for (let i = 0; i < 100; i++) c.notify(new Error('This is fail'))
-    expect(payloads.length).toBe(10)
+
+    it('prevents more than maxEvents being sent', () => {
+      for (let i = 0; i < 100; i++) c.notify(new Error('This is fail'))
+      expect(payloads.length).toBe(10)
+      expect(mockWarn).toHaveBeenCalledTimes(90)
+    })
+
+    it('throws an appropriate error message', () => {
```
imjoehaines, line 19: Maybe use "logs" instead of "throws" here?

```suggestion
    it('logs an appropriate message', () => {
```

packages/plugin-simple-throttle/test/throttle.test.ts
```diff
@@ -3,11 +3,21 @@ import plugin from '../'
 import Client from '@bugsnag/core/client'
 
 describe('plugin: throttle', () => {
-  it('prevents more than maxEvents being sent', () => {
+  describe('addOnError', () => {
```
imjoehaines, line 6: I don't think we need this second describe

gingerbenw: agreed. not a lot of functionality here to describe separately

imjoehaines (approved):

PR #1742, opened 2022-05-18
New nextjs example

djskinner:
## Goal

Update the existing next.js example, which is outdated

NB: This is currently added as a new example side-by-side with the old one for easier reviewing whilst WIP. Eventually, it will replace the existing example.

Credit to https://github.com/vercel/next.js/tree/canary/examples/with-sentry for identifying the range of tests cases and various workaround for error handling in next.js.

# Known issues

## Session started twice on page load

This seems to be because nextjs triggers a router (history) change following load and so two requests to sessions.bugsnag.com can be seen on page load. Not currently sure how to prevent this.

## Source maps

Uploaded source maps do appear to work but there are some issues that need ironing out:

- No source maps for server-side code. It doesn't look like next.js generates these anyway.
- Events generated on the server have source map warnings. This is to do with the project type used in Bugsnag. If a node project type is used then the source map warnings are not shown. It's also possible to avoid this by having separate projects for client and server events (by providing two different API keys).


gingerbenw (approved):

imjoehaines (approved):

tomlongridge (approved):

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,53 @@
...
+| /client/test3 | There is an exception during React lifecycle that is caught by Next.js's React Error Boundary. In this case, when the component mounts. This should cause _error.js to render. |
+| /client/test4 | There is an unhandled Promise rejection during React lifecycle. In this case, when the component mounts.                                                                       |
+| /client/test5 | An Error is thrown from an event handler.                                                                                                                                      |
+
+## Known Issues
+
+## Errors are not captured in some scenarios
+
+With a local production build (`npm run build && npm start`):
+- API Test 2 doesn't work
```
gareththackeray, line 39: When we say "doesn't work" do we mean "fails"?  Or the test itself somehow can't run?

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,53 @@
...
+- Client Test 1 doesn't work
+
+When deployed to vercel:
```
gareththackeray, line 42:
```suggestion
When deployed to Vercel:
```

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,53 @@
+# Next.js
+
+This is an example project showing how to use the universal `@bugsnag/js` notifier to track server, browser, and vercel serverless errors on a Next.js project.
```
gareththackeray, line 3:
```suggestion
This is an example project showing how to use the universal `@bugsnag/js` notifier to track server, browser, and Vercel serverless errors on a Next.js project.
```

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+This example covers various different scenarios where errors can occur in a next.js project. Currently not all scenarios are captured by Bugsnag.
+
+Note: Exceptions in development mode take a different path than in production. These tests should be run on a production build deployed to vercel.
```
gareththackeray, line 7:
```suggestion
Note: Exceptions in development mode take a different path than in production. These tests should be run on a production build deployed to Vercel.
```

examples/js/nextjs-new/README.md
gareththackeray, line 3: Is "universal" a term we generally use?  Might be better to remove it to avoid any confusion with e.g. Universal Angular.

djskinner: That wording was taken from the old example but it's not a particularly useful term. I'll remove it.

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,53 @@
...
+| /client/test1 | There is a top-of-module Promise that rejects, but its result is not awaited.                                                                                                  |
+| /client/test2 | There is a top-of-module exception. _error.js should render.                                                                                                                   |
+| /client/test3 | There is an exception during React lifecycle that is caught by Next.js's React Error Boundary. In this case, when the component mounts. This should cause _error.js to render. |
+| /client/test4 | There is an unhandled Promise rejection during React lifecycle. In this case, when the component mounts.                                                                       |
+| /client/test5 | An Error is thrown from an event handler.                                                                                                                                      |
+
+## Known Issues
+
+## Errors are not captured in some scenarios
```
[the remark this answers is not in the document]
djskinner: > Do we have any idea why the tests don't work / fail? Is it documented anywhere?

Not really. For API Test 1 and API Test 2 there doesn't seem to be any documented approach for catching top-of-module exceptions/rejections. The Sentry example states "Note that 1 and 2 are not expected to work if deployed to Vercel yet."

For API Test 4 I do not understand why the flush mechanism isn't working.

As for why none of the SSR tests pass it's a similar issue of not really knowing where to start to get an understanding of what's happening.

> Also, have we stated anywhere that we don't support a custom server at all?

It's not stated anywhere currently. It could be but it's not that we don't support it at all but rather limitations in the next.js API for custom servers means that errors get swallowed rather than being exposed to be handled by something like `@bugsnag/plugin-express`. Further details: https://github.com/vercel/next.js/issues/1852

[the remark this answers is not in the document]
gareththackeray: > > Do we have any idea why the tests don't work / fail? Is it documented anywhere?
> 
> Not really. For API Test 1 and API Test 2 there doesn't seem to be any documented approach for catching top-of-module exceptions/rejections. The Sentry example states "Note that 1 and 2 are not expected to work if deployed to Vercel yet."
> 
> For API Test 4 I do not understand why the flush mechanism isn't working.
> 
> As for why none of the SSR tests pass it's a similar issue of not really knowing where to start to get an understanding of what's happening.

OK

> > Also, have we stated anywhere that we don't support a custom server at all?
> 
> It's not stated anywhere currently. It could be but it's not that we don't support it at all but rather limitations in the next.js API for custom servers means that errors get swallowed rather than being exposed to be handled by something like `@bugsnag/plugin-express`. Further details: [vercel/next.js#1852](https://github.com/vercel/next.js/issues/1852)

Don't we also have the issue of leakage between requests though?  Because we don't get the isolation that serverless offers.  Although I suppose if it simply doesn't work with SSR it's academic?  Does anything work SSR?  E.g. handled errors?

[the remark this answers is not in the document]
gareththackeray: Sorry I see now that API Test 4 is handled errors.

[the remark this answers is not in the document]
djskinner: > Don't we also have the issue of leakage between requests though?

No, I don't think so. This is from a previous investigation I did on this:

There does not appear to be any issues with context leak when using Bugsnag.addMetaData. This was tested with the following function and verified by making multiple requests and ensuring the count was always one and that there was only one randomly generated metadata header:

```
import { NextApiRequest, NextApiResponse } from 'next'
import Bugsnag from '@bugsnag/js'
import BugsnagPluginAwsLambda from '@bugsnag/plugin-aws-lambda'

let count = 0;
Bugsnag.start({
  apiKey: process.env.BUGSNAG_API_KEY,
  plugins: [BugsnagPluginAwsLambda],
})

const bugsnagHandler = Bugsnag.getPlugin('awsLambda').createHandler()

function handler (
  request: NextApiRequest,
  response: NextApiResponse
) {
  count++
  Bugsnag.addMetadata('count', { count })
  Bugsnag.addMetadata('meta' + Math.random(), { some: 'stuff' + Math.random() })
  throw new Error('oops!');
  response.status(200).json({
    body: request.body,
    query: request.query,
    cookies: request.cookies
  })
}

export default bugsnagHandler(handler)
```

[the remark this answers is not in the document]
gareththackeray: I'm still worried about the leakage / shared context.. Do we have any reason to think that there wouldn't be an issue in a custom server, if errors could be reported?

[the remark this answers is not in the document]
gareththackeray: Same question goes for the api tests since these are using `Busgnag.` rather than `req.busnag.` like in usual node.  Is there any mileage in using `req.bugsnag.` for the API handlers?

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+With a local production build (`npm run build && npm start`):
```
gareththackeray, line 38: I'm slightly concerned about the ambiguity of saying the test fails.  What we really mean is that the error is not reported in this scenario and we know that because the test fails.  Can we perhaps say e.g. "API scenario X doesn't work" rather than "API test X"?  And perhaps rename the tests accordingly?

If you think I'm being overly fussy let me know but I can imagine coming to this and thinking "ok the test fails, but is that an issue with the test or does the functionality actually not work".

examples/js/nextjs-new/README.md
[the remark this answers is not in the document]
djskinner: > Do we have any reason to think that there wouldn't be an issue in a custom server, if errors could be reported?

In a custom server you are running next.js as a middleware in express (or koa etc.) so in that setup you would have the leakage issue an would need to use `req.bugsnag`.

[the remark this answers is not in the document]
djskinner: > Same question goes for the api tests since these are using Busgnag. rather than req.busnag. like in usual node. Is there any mileage in using req.bugsnag. for the API handlers?

I'm not sure I understand the concern here. The serverless environment is not like usual node and as far as I can tell there is no leakage issue. Each request gets a fresh context to run in. This applies to AWS as well as Vercel serverless. For example in our AWS docs we do not mention needing to use `req.bugsnag`. Here is some code from an E2E test for handled exceptions on AWS, which uses `Bugsnag.notify`:

```
const handler = (event, context, callback) => {
  Bugsnag.notify(new Error('Hello!'))

  callback(null, {
    statusCode: 200,
    body: JSON.stringify({ message: 'Did not crash!' })
  })
}
```

From looking at the source for `@bugsnag/plugin-aws-lambda` I don't see the Bugsnag client being attached to the `req` so I don't believe `req.bugsnag` would be defined.

examples/js/nextjs-new/README.md
djskinner: Yeah I think scenario is a better word 👍

examples/js/nextjs-new/README.md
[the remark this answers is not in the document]
gareththackeray: When I said "same question applies" I meant the question about a custom server.  I think the bottom line is that if there is leakage in a custom server we really ought to state that we don't support custom servers for that reason.

HOWEVER, if it were possible to use `req.bugsnag.` in a custom server for the API routes then we could explain how to do that, and just say that SSR is not supported.  But I've no idea if that's a realistic possibility.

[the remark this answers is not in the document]
djskinner: Actually, it's not clear from the docs whether it is possible or not. Perhaps it is, in which case I don't know which is the preferred approach. What is clear from the docs is that it's not normal practice to use a custom server at all.

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,57 @@
+# Next.js
+
+This is an example project showing how to use the `@bugsnag/js` notifier to track server, browser, and Vercel serverless errors on a Next.js project.
```
gareththackeray, line 3: Would this be accurate?  Or could "server" have another meaning beyond "custom server"?
```suggestion
This is an example project showing how to use the `@bugsnag/js` notifier to track browser and Vercel serverless errors on a Next.js project. Custom servers are not supported due to the potential for data leakage between concurrent requests.
```

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,57 @@
...
+| /client/scenario1 | There is a top-of-module Promise that rejects, but its result is not awaited.                                                                                                  |
+| /client/scenario2 | There is a top-of-module exception. _error.js should render.                                                                                                                   |
+| /client/scenario3 | There is an exception during React lifecycle that is caught by Next.js's React Error Boundary. In this case, when the component mounts. This should cause _error.js to render. |
+| /client/scenario4 | There is an unhandled Promise rejection during React lifecycle. In this case, when the component mounts.                                                                       |
+| /client/scenario5 | An Error is thrown from an event handler.                                                                                                                                      |
+
+## Known Issues
+
+## Errors are not captured in some scenarios
```
gareththackeray, line 36: All these subheadings of "Known Issues" need an extra #
```suggestion
### Errors are not captured in some scenarios
```

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+With a local production build (`npm run build && npm start`):
+- API scenario 2 fails
+- Client scenario 1 fails
+
+When deployed to Vercel:
+- API scenario 1, API scenario 2 and API scenario 4 fail
+- All of the SSR scenarios fail
+
+## Session started twice on page load
```
gareththackeray, line 46:
```suggestion
### Session started twice on page load
```

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+This is because nextjs calls `replaceState` following load and so two requests to sessions.bugsnag.com can be seen on page load. We plan to stop `replaceState` from triggering a session by default. In the meantime you can prevent this behaviour by disabling automatic session management and implementing it yourself as required.
+
+## Source maps
```
gareththackeray, line 50:
```suggestion
### Source maps
```

examples/js/nextjs-new/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+- No source maps for server-side code. It doesn't look like next.js generates these anyway.
+- Events generated on the server have source map warnings. This is to do with the project type used in Bugsnag. If a node project type is used then the source map warnings are not shown. It's also possible to avoid this by having separate projects for client and server events (by providing two different API keys).
+
+## Custom Server
```
gareththackeray, line 55: I think this one is a subheading too?
```suggestion
### Custom Server
```

PR #1745, opened 2022-05-20
deps(react-native): Update bugsnag-cocoa to v6.17.1

nickdowell:


nickdowell:
Cannot update to v6.17.1 due to a bug: https://github.com/bugsnag/bugsnag-cocoa/pull/1392

PR #1751, opened 2022-05-27
[PLAT-7565] Rework network breadcrumbs

gingerbenw:
## Goal

To reduce likelihood of xml http requests losing request data in network breadcrumbs

## Testing

Test for errors when losing request data must be skipped as the data can no longer be removed.


djskinner (approved):
👍 seems better

PR #1759, opened 2022-06-10
[PLAT-8549] Error cause

gingerbenw:
## Goal

To support capturing errors included in `Error.cause` within the notifiers

## Changeset

`Event.create` updated within core to format and include error causes as exceptions within the dashboard

## Testing

- Unit tests to validate error causes
- End to end tests to ensure causes are captured on appropriate environments

### DIFF
--- a/packages/core/event.js
+++ b/packages/core/event.js
@@ -9,32 +9,6 @@ const metadataDelegate = require('./lib/metadata-delegate')
 const featureFlagDelegate = require('./lib/feature-flag-delegate')
 const isError = require('./lib/iserror')
 
-function createBugsnagError (errorClass, errorMessage, type, stacktrace) {
-  return {
-    errorClass: ensureString(errorClass),
-    errorMessage: ensureString(errorMessage),
-    type,
-    stacktrace: reduce(stacktrace, (accum, frame) => {
-      const f = formatStackframe(frame)
-      // don't include a stackframe if none of its properties are defined
-      try {
-        if (JSON.stringify(f) === '{}') return accum
-        return accum.concat(f)
-      } catch (e) {
-        return accum
-      }
-    }, [])
-  }
-}
-
-function getCauseStack (error) {
-  if (error.cause) {
-    return [error, ...getCauseStack(error.cause)]
-  } else {
-    return [error]
-  }
-}
-
 class Event {
   constructor (errorClass, errorMessage, stacktrace = [], handledState = defaultHandledState(), originalError) {
     this.apiKey = undefined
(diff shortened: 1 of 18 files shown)


packages/core/event.js
```diff
@@ -181,6 +193,19 @@ Event.create = function (maybeError, tolerateNonErrors, handledState, component,
...
+  if (error.cause) {
+    const causes = getCauseStack(maybeError).slice(1)
+    const normalisedCauses = map(causes, (cause) => {
+      const [error] = normaliseError(cause, tolerateNonErrors, 'error cause')
+      if (error.name === 'InvalidError') {
+        event.addMetadata('error cause', makeSerialisable(cause))
+      }
+      return error
+    })
+    const causeErrors = map(normalisedCauses, (cause) => createBugsnagError(cause.name, cause.message, Event.__type, [])) // do not include stacktrace for error causes
```
gareththackeray, line 205: If the cause is a JS `Error` it's fine to include the stacktrace.  It's only if it is some other type we don't want to generate it.  But looking at the implementation that generation is only done in `Event.create()` at the top level anyway.  The stack should already be on the normalised error I think?  It's hard to follow it completely.

As related issue, looking at the `normaliseError` function, do we need to do something different with the `internalFrames` if it's a cause?  I don't know the answer but worth checking.

gingerbenw: refactored the normalisation of the causes, which checks for the existence of an error stack before passing to `normaliseError` and have updated test cases to ensure stacks are present where expected.

packages/core/test/event.test.ts
```diff
@@ -335,4 +335,77 @@ describe('@bugsnag/core/event', () => {
...
+        stacktrace: [],
+        type: 'browserjs'
+      })
+      expect(event.getMetadata('error cause')).toEqual({ error: 'I am not a real cause' })
+    })
+
+    it('handles invalid cause errors regardless of tolerateNonErrors', () => {
+      const err = new Error('I am the error')
+      // @ts-ignore
+      err.cause = 'I am not a real cause'
```
djskinner, line 397: despite the test name this is not an invalid cause error

djskinner (approved):
just one comment about the purpose of one of the tests

packages/core/test/event.test.ts
```diff
@@ -335,4 +335,78 @@ describe('@bugsnag/core/event', () => {
...
+      expect(event.errors).toContainEqual({
+        errorClass: 'InvalidError',
+        errorMessage: 'error cause was a non-error. See "error cause" tab for more detail.',
+        stacktrace: [],
+        type: 'browserjs'
+      })
+      expect(event.getMetadata('error cause')).toEqual({ error: 'I am not a real cause' })
+    })
+
+    it('tolerates non-errors regardless of tolerateNonErrors being true/false', () => {
```
gareththackeray, line 394: You've missed out the fact that this is testing causes specifically.
```suggestion
    it('tolerates non-error causes regardless of tolerateNonErrors being true/false', () => {
```

packages/core/test/event.test.ts
```diff
@@ -335,4 +335,78 @@ describe('@bugsnag/core/event', () => {
...
+      const err = new Error('I am the error')
+      // @ts-ignore
+      err.cause = 'I am not a real cause'
```
gareththackeray, line 397: Probably wouldn't have mentioned / noticed this if the other discussions hadn't drawn attention to it, but it's not really true that it's not a cause... The key thing is it's not an `Error` 😄 
```suggestion
      err.cause = 'I am not an Error'
```

packages/core/event.js
```diff
@@ -181,6 +193,19 @@ Event.create = function (maybeError, tolerateNonErrors, handledState, component,
   if (error.name === 'InvalidError') {
     event.addMetadata(`${component}`, 'non-error parameter', makeSerialisable(maybeError))
   }
+  if (error.cause) {
```
gareththackeray, line 196: nit: it seems mildly odd that you're checking for `error.cause` but passing `maybeError` to `getCauseStack`.  Perhaps this is the right thing to do, but just looks slightly surprising.

gingerbenw: that is surprising! it should have been `error`

PR #1772, opened 2022-06-27
slint code, fix get component name when use script setup way

sald19:
## Goal

Get component name in vue3 when use script setup

CleanShot 2022-06-27 at 15 41 38@2x

### DIFF
--- a/packages/plugin-vue/src/vue.js
+++ b/packages/plugin-vue/src/vue.js
@@ -1,28 +1,55 @@
 module.exports = (app, client) => {
-  const prev = app.config.errorHandler
+  const prev = app.config.errorHandler;
 
   const handler = (err, vm, info) => {
-    const handledState = { severity: 'error', unhandled: true, severityReason: { type: 'unhandledException' } }
-    const event = client.Event.create(err, true, handledState, 'vue error handler', 1)
+    const handledState = {
+      severity: "error",
+      unhandled: true,
+      severityReason: { type: "unhandledException" },
+    };
+    const event = client.Event.create(
+      err,
+      true,
+      handledState,
+      "vue error handler",
+      1
+    );
 
-    event.addMetadata('vue', {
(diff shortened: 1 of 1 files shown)


luke-belton:
Hi @sald19 - thanks for raising this. Please could you tell us a bit more about the problem you're trying to solve here - perhaps you could share a cut-down reproduction example? I'm still seeing the `vue.component` field set correctly in Bugsnag when using the new script syntax, i.e. `<script setup>`.

sald19:
the `vm`  instance (ViewModel) does not have the  property `name`
CleanShot 2022-06-29 at 13 12 49

only get `__name` or `__file`

luke-belton:
Hi @sald19 if I import a `HelloWorld` component which contains a simple call to `Bugsnag.notify(new Error('Test'))` like this:
```
<script setup>
   import HelloWorld from "./components/HelloWorld.vue";
</script>
```
I see the `vm.$options.name` property set to `HelloWorld`:

Screenshot 2022-07-05 at 15 43 00

What version of Vue are you using? Are you able to share a reproduction example to demonstrate the behaviour you're seeing?

sald19:
my current setup is 

composer.json
```
 "inertiajs/inertia-laravel": "^0.6.3",
 "laravel/framework": "^9.19",
```

package.json
```
"@bugsnag/js": "^7.17.0",
"@bugsnag/plugin-vue": "^7.17.0",
```

johnkiely1:
Hi @sald19, I think it would be beneficial if you could share a cut down reproduction example demonstrating the issue you are seeing, as we have been unable to reproduce it. This is likely the best way for us to progress on this. Thanks.

luke-belton:
Hi @sald19 - it's been a little while since we heard from you on this, so I'm going to close this for now. Feel free to reopen in the future if you're able to create a reproduction example 👍

PR #1775, opened 2022-06-28
Add RN 0.69 Test Fixtures and CI coverage

joshedney:
## Goal

Add RN 0.69 test fixtures and CI coverage

## Changeset

Add RN and RN CLI 0.69 test fixtures and CI coverage

## Testing

Covered by Full CI


gingerbenw (approved):
👍🏻

starting to wonder if we could use `react-native init` to install various versions and install with the bugsnag cli tool, rather than having to commit all these fixtures in different versions 🤔

PR #1777, opened 2022-06-29
deps(react-native): Update bugsnag-cocoa to v6.19.0

nickdowell:


gingerbenw (approved):
yep, that's a _lot_ cleaner!

PR #1784, opened 2022-07-07
Fix the version of vue in the vue-webpack4 tests

Cawllec:
## Goal

A simple fix to our current CI issues, as something within the latest release of Vue causes the es3ify webpack plugin to fail.

I've attempted to replace this, but unfortunately that meant tests wouldn't run on certain IE versions.


twometresteve (approved):
LGTM to get CI running again and we've started an internal conversation to look more deeply into the underlying issue.

PR #1785, opened 2022-07-07
e2e tests: revert vue dependency locks

djskinner:
## Goal

Remove the dependency lock that has been added to the E2E tests so that our test suite can again run against the latest versions of vue2 and vue 3.

- typescript_vue3: Vue3 now includes ECMAScript 2018 syntax (specifically https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals). Whilst this seems to go against the claims Vue3 makes about browser support https://vuejs.org/about/faq.html#what-browsers-does-vue-support browsers that do not support this syntax are not excluded.
- plugin_vue/webpack4: New syntax in the vue bundle seems not supported by `es3ify-webpack-plugin` so use `babel` instead.
- typescript_vue2: New syntax in the vue bundle triggers a bug in Safari 10

## Design

- Continuing with the existing method of checking if a test can run by `eval`ing code in the fixture. It didn't seem worthwhile or appropriate to change the way things are being done as part of this.

## Changeset

- Remove the dependency lock from the typescript_vue3 fixture
- Remove the dependency lock from the typescript_vue2 fixture
- Remove the dependency lock from the plugin_vue/webpack4 fixture
- Add a check for _spread in object literals_ support to the typescript_vue3 fixture
- Replace `es3ify-webpack-plugin` in the plugin_vue/webpack4 fixture with `babel`
- Do not run the typescript_vue2 fixture if it would trigger the Safari 10 bug

## Testing

- e2e testing and browserify live


test/browser/features/fixtures/plugin_vue/typescript_vue3/index.html
```diff
@@ -12,11 +12,14 @@
     <script>
       var el = document.getElementById('bugsnag-test-should-run')
       var isClassSupported = false
+      var isObjectSpreadSupported = false
       try {
         eval('class Foo extends Object {}')
         isClassSupported = true
+        eval('const a = {}; const b = { ...a }')
```
djskinner, line 19: this syntax has started to appear in the Vue bundles. This is despite it being an ECMAScript 2018 addition and Vue claiming to support all browsers that support ES2015

test/browser/features/fixtures/plugin_vue/typescript_vue2/index.html
```diff
@@ -11,7 +11,16 @@
...
     <script>
       var el = document.getElementById('bugsnag-test-should-run')
-      el.textContent = el.innerText = typeof Proxy !== 'undefined'
+      var isLetSupported = false
+      var hasSafari10Bug = false
+      try {
+        eval('let a = "foo"')
+        isLetSupported = true
+        // See https://bugs.webkit.org/show_bug.cgi?id=171041
+        eval('function(a) { for(let a of []) { console.log(a) } }')
```
djskinner, line 20: Bug in Safari that is being triggered by later versions of the Vue2 bundle, which now include const/let/class etc. I'm not sure if this is a mistake by Vue2 as it seemed it should continue to support older browsers

test/browser/features/fixtures/plugin_vue/typescript_vue2/index.html
```diff
@@ -11,7 +11,16 @@
...
     <script>
       var el = document.getElementById('bugsnag-test-should-run')
-      el.textContent = el.innerText = typeof Proxy !== 'undefined'
+      var isLetSupported = false
+      var hasSafari10Bug = true
+      try {
+        eval('let a = "foo"')
+        isLetSupported = true
+        // See https://bugs.webkit.org/show_bug.cgi?id=171041
+        eval('let e = e => { console.log(e); for (let e of [1, 2, 3]) console.log(e) }')
```
djskinner, line 20: Bug in Safari that is being triggered by later versions of the Vue2 bundle, which now include const/let/class etc. I'm not sure if this is a mistake by Vue2 as it seemed it should continue to support older browsers

gingerbenw: that's some weird syntax!

gingerbenw (approved):
seems fine to me if the tests are passing and the vue version isn't locked down! 👍🏻

PR #1795, opened 2022-07-28
BitBar web and updated test versions

Cawllec:
## Goal

Moves latest versions of each browser to using BitBar Web, and removes intermediate versions that have been determined to be unnecessary.

Note - This will remain WIP until Maze-runner v7 can be fully released.

## Design

Since the dependencies for Maze-runner V7 (for BitBar Web) and V6 (for BrowserStack) are mutually exclusive we're going to be using a mix of V7 and V6 going forward, until we can reconcile these differences.

## Changeset

Removed browsers:
- Chrome v30
- Chrome v61
- Edge v14
- Edge v15
- Safari v6
- Safari v13
- Firefox v30
- Firefox v56

Added browsers:
- Chrome v72
- Edge v17
- Firefox v78

Moved to BitBar web:
- Chrome latest -> Chrome v102
- IE 11 -> IE 11
- Edge v102 (No previous version)
- Safari v15 -> Safari v15
- Firefox latest -> Firefox 101


.buildkite/browser-pipeline.yml
```diff
@@ -16,228 +16,206 @@ steps:
...
+          cache-from:
+            - browser-maze-runner-v6:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:ci-browser-v6-${BRANCH_NAME}
+      - docker-compose#v3.9.0:
+          push:
+            - browser-maze-runner-v6:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:ci-browser-v6-${BRANCH_NAME}
 
-  - label: ':chrome: v61 Browser tests'
+# BitBar tests
+
+  - label: ':firefox: 101 Browser tests'
```
twometresteve, line 39: Could we somehow make it obvious on the Buildkite build page which steps were run on BB and which were run on BS?  Could with maybe prefix the label of each step with either "BB -" or "BS -"?  Or maybe Buildkite annotations could be used here, like providing a header for each set of steps (I've never tried this, so don't how how it would actually look in practice).

twometresteve (commented):
Changes look good to me, just made one minor comment (which in hindsight may be unnecessary given that all the BB steps are grouped together at the top).

Happy to see this merged, but I'll let @gingerbenw make the approval so that he has a chance to check that he's happy with the test matrix as it now stands.

dockerfiles/Dockerfile.browser
```diff
@@ -45,7 +45,7 @@ RUN find . -name package.json -type f -mindepth 2 -maxdepth 3 ! -path "./node_mo
 RUN rm -fr **/*/node_modules/
 
 # The maze-runner browser tests
-FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:latest-v6-cli as browser-maze-runner
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:integration-library-updates-cli as browser-maze-runner
```
twometresteve, line 48: Oh, remove to update this for maze v7 once released, of course.

gingerbenw: worth a separate ticket, or a TODO in the code?

dockerfiles/Dockerfile.browser-v6
```diff
@@ -0,0 +1,51 @@
+# CI test image for unit/lint/type tests
+FROM node:14-alpine as browser-feature-builder
```
gingerbenw, line 2: just a question - what's the purpose of aliasing the base image?

gingerbenw (approved):
thanks! I'm happy here :)

dockerfiles/Dockerfile.browser
twometresteve: It'll need changing before this PR is merged.

dockerfiles/Dockerfile.browser-v6
twometresteve: See L50

gingerbenw: ahh! got it. thanks!

Cawllec: There's not really one, except for clarity.

Cawllec:
I've used the `group` annotation in Buildkite to group the steps at the top of the page.  Hopefully this should make it a bit more readable.  @gingerbenw Looking at the buildkite run (for browsers) does that look better?

gingerbenw:
ooh, yeah! that's neat 👌🏻

PR #1802, opened 2022-08-12
[PLAT-8735] Preserve feature flag order

gingerbenw:
## Goal

To preserve the order of feature flags as they were added

## Design

Changed internal feature storage from an object to an array, added relevant poly fills for IE11 compatibility

## Testing

Unit and integration tests updated


packages/core/lib/feature-flag-delegate.js
```diff
@@ -14,7 +14,16 @@ function add (existingFeatures, name, variant) {
     variant = jsonStringify(variant)
   }
 
-  existingFeatures[name] = variant
+  var found = false
+  for (var i = 0; i < existingFeatures.length; i++) {
+    if (existingFeatures[i].name === name) {
+      found = true
+      existingFeatures[i].variant = variant
```
nickdowell, line 21: Would it be safer to replace rather than mutate the object, since an event.features may also contain it?

e.g.

```suggestion
      existingFeatures[i].variant = { name, variant }
```

gingerbenw: I don't think it makes any difference here, as the flag will have been _cloned_ from the event, rather than referenced.

packages/core/test/client.test.ts
```diff
@@ -1095,13 +1095,13 @@ describe('@bugsnag/core/client', () => {
...
-            { featureFlag: 'd', variant: '3' },
-            { featureFlag: 'e' }
-          ])
+          expect(event.featureFlags).toEqual(expect.arrayContaining([
+            expect.objectContaining({ featureFlag: 'a', variant: '1' }),
+            expect.objectContaining({ featureFlag: 'b', variant: '2' }),
+            expect.objectContaining({ featureFlag: 'c' }),
+            expect.objectContaining({ featureFlag: 'd', variant: '3' }),
+            expect.objectContaining({ featureFlag: 'e' })
+          ]))
```
djskinner, line 1104: why no longer a strict check here?

packages/plugin-electron-ipc/bugsnag-ipc-main.js
```diff
@@ -102,7 +102,7 @@ module.exports = class BugsnagIpcMain {
       }
       event.context = event.context || this.client._context
       event._metadata = { ...event._metadata, ...this.client._metadata }
-      event._features = { ...event._features, ...this.client._features }
+      event._features = [...event._features, ...this.client._features]
```
djskinner, line 105: In `client.js` there is `[...this._features, ...event._features]`. Should the order of the spread be the same here and there i.e. event first then client or client first then event?

gingerbenw: that's a really good question, actually. @gareththackeray would this kind of thing be defined in the spec?

packages/core/test/client.test.ts
gingerbenw: good spot - I updated this test when I was trying to use a Map, but this can now be changed back to strictEqual

packages/plugin-electron-ipc/bugsnag-ipc-main.js
kattrali: Any client features with a given name should be overridden by event features with the same name, and client features should be ordered first (as they are added before the event is created).

gingerbenw: should that be the same with all the other metadata, or just features?

gareththackeray: Yes you'd expect anything that was added to the event to override the client level stuff.

packages/core/lib/feature-flag-delegate.js
```diff
@@ -14,7 +14,16 @@ function add (existingFeatures, name, variant) {
     variant = jsonStringify(variant)
   }
 
-  existingFeatures[name] = variant
+  var found = false
+  for (var i = 0; i < existingFeatures.length; i++) {
```
gareththackeray, line 18: Should we be concerned about the performance of this?  I guess it would be possible to have a secondary data structure (i.e. a map) that allowed quick lookup of whether the feature exists at all, so that you don't need to iterate over all the features every time you add one?

gingerbenw: https://jsben.ch/cDI8i

interesting, my (admittedly limited) benchmarking shows the array search to be faster than using a Map

packages/plugin-electron-ipc/bugsnag-ipc-main.js
gingerbenw: so I should update this to perform a proper merge at this point? currently this could have duplicate flags.

packages/core/lib/feature-flag-delegate.js
gareththackeray: If you've only got 2 features it'll be basically a wash.  I'm thinking about when you've got 100 or 1000 features.

packages/plugin-electron-ipc/bugsnag-ipc-main.js
gareththackeray: Oh, yes.

gareththackeray (approved):

PR #1815, opened 2022-09-16
[PLAT-8736] add getFeatureFlags method to event

gingerbenw:
## Goal

To enable reading feature flags attached to a specific event passed to `onError` or `onSend`

## Testing

Unit test added to event in core


djskinner (approved):

djskinner:
integration test looks good to me

gareththackeray (approved):

PR #1823, opened 2022-09-29
remove js payload limit

djskinner:
## Goal

- Remove the check that prevented attempting delivery of event and session payloads > 1MB
- For delivery implementations that support retries, ensure that any failed event is not retried if > 1MB

## Design

- Remove code in json-payload that throws for oversized events thus allowing for attempted send
- Check payload size of failed-to-send events in each delivery method and handle them accordingly: logging and ensuring they are not retried

## Testing

- Manual tests
  - [x] Electron: Attempt to send over-sized (breadcrumb metadata) payload and verify log output and no retry
  - [x] Node
  - [x] x-domain-request: Tested on ie8/ie9 via Browserstack
  - [x] xml-http-request: Tested on Firefox, Chrome (latest), IE11 via Browserstack
  - [x] expo: tested with a local expo app force pulling a local version of @bugsnag/core with these changes in and verifying no issues upon the 400 response (though no logging as that only gets added in https://github.com/bugsnag/bugsnag-expo/pull/65)
- [x] e2e test coverage for much of the above


packages/delivery-node/delivery.js
```diff
@@ -3,8 +3,13 @@ const request = require('./request')
 
 module.exports = (client) => ({
   sendEvent: (event, cb = () => {}) => {
+    const body = payload.event(event, client._config.redactedKeys)
+
     const _cb = err => {
       if (err) client._logger.error(`Event failed to send…\n${(err && err.stack) ? err.stack : err}`, err)
+      if (body.length > 10e5) {
+        client._logger.warn(`Discarding over-sized event (${body.length / 10e5}) after failed delivery`)
+      }
```
djskinner, line 12: the spec doesn't actually mention logging for delivery methods that don't support persistence so I'm not sure if this is wanted cc @gareththackeray and if it is perhaps the wording needs to be slightly different. Same question for the two browser delivery methods.

packages/delivery-x-domain-request/delivery.js
```diff
@@ -3,14 +3,22 @@ const payload = require('@bugsnag/core/lib/json-payload')
...
   sendEvent: (event, cb = () => {}) => {
     const url = getApiUrl(client._config, 'notify', '4', win)
+    const body = payload.event(event, client._config.redactedKeys)
+
     const req = new win.XDomainRequest()
     req.onload = function () {
       cb(null)
     }
+    req.onerror = function () {
+      client._logger.error('Event failed to send…\n')
```
djskinner, line 13: node does `logger.error(`Event failed to send…` so I thought maybe the browser delivery methods should too

packages/delivery-electron/test/delivery.test-main.ts
```diff
@@ -207,6 +207,39 @@ describe('delivery: electron', () => {
     })
   })
 
+  it('does not attempt to re-send oversized payloads', done => {
```
djskinner, line 210: this test fails because the delivery is depending on an old version. Am I missing something here? Do we need to make and merge the changes to core separately, then bump the version of core used by various components such as delivery-electron?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Changed
+
+- Attempt to send all event and session payloads, regardless of size
```
djskinner, line 7: not sure about this wording due to the stripping of the first event's metadata

gingerbenw (approved):
changelog wording seems fine to me, but @gareththackeray might have a better idea

gareththackeray (approved):

test/node/features/fixtures/express/scenarios/app.js
```diff
@@ -70,6 +104,26 @@ app.get('/throw-non-error', function (req, res, next) {
...
+    return a.join('');
+  }
+
+  var big = {};
+  var i = 0;
+  while (JSON.stringify(big).length < 5*10e5) {
+    big['entry'+i] = repeat('long repetitive string', 1000);
+    i++;
+  }
+  req.bugsnag.leaveBreadcrumb('big thing', big);
```
djskinner, line 122: works even though apparently not supported on node because any cloned clients don't get the patched method

test/node/features/delivery.feature
```diff
@@ -0,0 +1,21 @@
...
+  And I set the HTTP status code for the next "POST" request to 400
+  When I open the URL "http://express/oversized"
+  And I wait for 5 seconds
+  Then I wait to receive an error
+
+  # Check that Bugsnag is discarding the event
+  And I wait to receive 3 logs
+  Then I discard the oldest log
+  Then I discard the oldest log
+  And the log payload field "message" equals "Event oversized (5.02 MB)"
```
djskinner, line 21: @Cawllec whilst persistence is not enabled in node yet I'd be inclined (if it's easy) to add further steps to assert when the app is reloaded no sends are attempted. Is that currently possible?

Cawllec: Absolutely, although you may need to add a step locally that calls `Maze::Docker.down_all_services`, something like:
```
Then('I stop all docker services') do
    Maze::Docker.down_all_services
end
```
and then empty out the old stack of payloads and restart the service:
```
Then I stop all docker services
And I discard the oldest error
And I start the service "express"
And I wait for the host "express" to open port "80"
And I should receive 0 requests
```

test/electron/features/delivery.feature
```diff
@@ -0,0 +1,15 @@
...
+        And I set the HTTP status code for the next "POST" request to 400
+        And I click "oversized-notify"
+        Then I wait to receive an error
+
+        # Check that Bugsnag is discarding the event
+        And I wait to receive 3 logs
+        Then I discard the oldest log
+        Then I discard the oldest log
+        And the log payload field "message" equals "Event oversized (2.00 MB)"
+
```
djskinner, line 15: @Cawllec this doesn't run because the steps are undefined for electron. Perhaps this is something we can add later as I'm happy the manual testing has shown no issues

djskinner:
@imjoehaines @gingerbenw I've requested a re-review. The implementation hasn't changed since you last looked but there has been some work on the e2e tests so I thought it would be good for some more eyes on it again before we merge

test/node/features/fixtures/express/scenarios/app.js
imjoehaines: Yeah I noticed this with another monkey patched method recently — I think it's actually a good thing in this case though, since the breadcrumbs would be scoped to the request client so it works as expected?

djskinner: Yes, it works as you would expect. I have manually tested it for a different purpose.

PR #1831, opened 2022-10-07
[PLAT-8023] Prevent session/event delivery with invalid configuration

gingerbenw:
## Goal

If only one endpoint config option is set, no events or sessions should be sent from the app to avoid leaking data to SaaS Bugsnag.

## Changeset

Do not set delivery method when incorrect configuration is used

## Testing

Updated unit tests


packages/core/test/client.test.ts
```diff
@@ -72,6 +72,46 @@ describe('@bugsnag/core/client', () => {
...
+
+      client.notify(new Error('111'), () => {}, (err, event) => {
+        expect(err).toBe(null)
+        expect(event).toBeTruthy()
+        expect(event.errors[0].errorMessage).toBe('111')
+        done()
+      })
+
+      // give the event loop a tick to see if the event gets sent
+      process.nextTick(() => done())
```
djskinner, line 93: not sure i understand this. won't this just cause the test to pass regardless of what assertions above fail?

gingerbenw: I'm not sure either - I copied it from a similar test but I think it needs to go altogether.

packages/core/test/client.test.ts
```diff
@@ -72,6 +72,44 @@ describe('@bugsnag/core/client', () => {
...
+      client._setDelivery(client => ({
+        sendSession: () => { },
+        sendEvent: () => { done('sendEvent() should not be called') }
+      }))
+
+      client.notify(new Error('111'), undefined, (err, event) => {
+        expect(err).toBe(null)
+        expect(event).toBeTruthy()
+        expect(event.errors[0].errorMessage).toBe('111')
+        done()
```
djskinner, line 110: could we have a `expect(sendEvent).not.toHaveBeenCalled()` here?

gingerbenw: sure - I think it's redundant with `sendEvent: () => { done('sendEvent() should not be called') }` but it won't hurt.

gareththackeray (approved):

PR #1836, opened 2022-10-17
Update Browser e2e tests to use Maze Runner v7 exclusively

twometresteve:
## Goal

Port all browser tests to Maze Runner v7, removing the use of v6 for browsers.

## Changeset

Some unrelated changes were also needed to fix the CI build.

## Testing

Covered by CI.


test/electron/features/support/app.js
```diff
@@ -21,15 +21,15 @@ class TestApp {
     // install this first. electron has a lengthy postinstall script, and doing
     // a bare `npm install` first will wildcard to the latest / last version
     // installed, doubling the install time
-    await this._exec(npmRunner, [...installArgs, `electron@${electronVersion}`])
+    await this._exec(npmRunner, [...installArgs, `electron@${electronVersion}`], {}, 180)
```
Cawllec, line 24: Given Electron is continuing to fail, would it be worth removing any potential fixes from this PR and allowing work toward fixing the issue to be conducted on a fully separate branch?

twometresteve: I think it's a positive change, but I definitely take your point - how can we be sure that it doesn't make things worse in some way.  I could potentially leave us with a hidden issue that's difficult to resolve.

I'll revert the change and take a note of it to roll into the next set of changes.

Cawllec (approved):
Electron test failures aren't connected to the changes made here, so approving this in the context of wanting to move the buildkite pipeline to Maze-v7

PR #1844, opened 2022-10-21
Dynamically load Buildkite pipelines rather than trigger

twometresteve:
## Goal

Dynamically load the Browser, Electron, Node, RN and RN CLI pipelines into the main pipeline, rather than trigger them.

## Design

This has a few benefits:
- Less clicking around to see the overall state of a CI build
- Less annoyance from Buildkite's new behaviour of defaulting to the "Issues" view

Possible drawback:
- More on one page and harder to spot a section you may be more interested in.  I've tried to mitigate this by using blue circle enclosed titles to highlight sections (Buildkite doesn't currently provide a way of dividing sections up).

We could also use the Buildkite `group:` mechanism to group together sections in the banner at the top.

## Changeset

- Some job key changes needed to avoid clashes across pipelines (now they are dynamically loaded into the same "namespace"
- Made the use of quotes consistent (double) in the Buildkite yaml files.
- Browser/Device specific BrowserStack creds environment variables needed to be let through into the Docker containers (I've updated the secrets bucket that sets the variables - as there are separate folders in the bucker for each pipeline).

## Testing

Covered by CI.  Please follow the link from GitHub's checks section for a view of what the Buildkite build page now looks like.


.buildkite/pipeline.yml
```diff
@@ -47,75 +47,24 @@ steps:
...
-  - label: ':aws-lambda: AWS Lambda tests'
+  - label: ":aws-lambda: AWS Lambda tests"
     if: build.env("BUILD_RN_WITH_LATEST_NATIVES") != "true"
     timeout_in_minutes: 35
     agents:
-      queue: 'opensource-mac-cocoa-10.15'
+      queue: "opensource-mac-cocoa-10.15"
     commands:
-      # force the NPM registry as the default on CI is artifactory, which can't
+      # force the NPM registry as the default on CI is artifactory, which can"t
```
gingerbenw, line 67: slightly aggressive find and replace ;)

.buildkite/pipeline.yml
```diff
@@ -145,29 +94,61 @@ steps:
...
       docker-compose#v3.9.0:
         run: ci
-    command: 'npm run test:types'
+    command: "npm run test:types"
+
+  - label: ":large_blue_circle: :large_blue_circle: :large_blue_circle: BROWSER STEPS :large_blue_circle: :large_blue_circle: :large_blue_circle:"
+    if: build.env("BUILD_RN_WITH_LATEST_NATIVES") != "true"
+    depends_on: "package-js"
+    commands:
+      - buildkite-agent pipeline upload .buildkite/browser-pipeline.yml
```
gingerbenw, line 128: Would be happy to see stages grouped using buildkite rather than our own workaround

.buildkite/pipeline.yml
twometresteve: Ooh, I did realise that was a possibility and tried to catch that when looking over the resulting diff!

.buildkite/pipeline.yml
twometresteve: Talked in IRL to confirm it was the `:large_blue_circle:` we see as a workaround (I agree).  Unfortunately though Buildkite doesn't currently offer a better solution.  However, I will implement a group for each pipeline to see what we think (I think it'll probably be a good thing).

.buildkite/node-pipeline.yml
```diff
@@ -1,78 +1,80 @@
...
-  - label: ':node: Node 4'
-    depends_on: "node-maze-runner-image"
-    timeout_in_minutes: 30
-    plugins:
-      docker-compose#v3.9.0:
-        run: node-maze-runner
-        use-aliases: true
-    env:
-      NODE_VERSION: "4"
+      - label: ":node: Node 4"
```
imjoehaines, line 22: Not related to this PR but this looks like it could be easily matrix-ed

.buildkite/electron-pipeline.yml
```diff
@@ -1,42 +1,44 @@
...
-      - npm run build:electron
-      - defaults write com.apple.CrashReporter DialogType none
-      - npm run test:unit:electron-runner
-      - npm run test:electron
+  - group: "Electron Tests"
+    steps:
+      #
+      # Node 14
+      #
+      - label: "Electron 12 tests - macOS - Node 14"
```
imjoehaines, line 8: Not related to this PR but this looks like it could be easily matrix-ed

imjoehaines (approved):

.buildkite/pipeline.yml
imjoehaines: Groups look good to me and definitely better than the 🔵🔵🔵 markers — I'm guessing those steps will always have to be visible though as they're what uploads the pipelines?

It does seems a bit weird to me that the number in the group is (presumably) the number of finished tests instead of "passed/total" but that's easy to get used to 🤷

.buildkite/pipeline.yml
```diff
@@ -152,3 +141,14 @@ steps:
       - "android-builder-base"
     commands:
       - buildkite-agent pipeline upload .buildkite/react-native-cli-pipeline.yml
+
+  - label: ":large_blue_circle: :large_blue_circle: :large_blue_circle: EXPO PIPELINE :large_blue_circle: :large_blue_circle: :large_blue_circle:"
+    if: build.env("BUILD_RN_WITH_LATEST_NATIVES") != "true"
+    depends_on: "publish-js"
+    trigger: "bugsnag-expo"
```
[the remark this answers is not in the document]
twometresteve: When you say "every maintained branch" do you mean each supported version (46-next, 45-next, etc)?  Yes, I can see the argument for that, have raised internal ticket PLAT-9114 to do that.

.buildkite/electron-pipeline.yml
twometresteve: Agreed, should well well here, have raised PLAT-9115.

.buildkite/node-pipeline.yml
twometresteve: Agreed, should well well here, have raised PLAT-9115.

.buildkite/pipeline.yml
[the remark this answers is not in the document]
imjoehaines: > When you say "every maintained branch" do you mean each supported version (46-next, 45-next, etc)?

Yeah, exactly

.buildkite/pipeline.yml
twometresteve: A step of some kind would be needed, but the horrible blue circles could I think be moved into the groups themselves to make the initial banner more readable.  I'd quite like to keep some kind of section separator in, as I generally use the body of the build page rather than the header.  But admittedly I have a different perspective to most and perhaps I need to look at changing my habit a little.

This PR certainly doesn't need to be the end of the discussion though, I'll continue to try out different approaches - this was very much just the first iteration.  Shrinking the 7 pipeline pages down to 2 was the main thing to achieve here.

PR #1856, opened 2022-10-31
Do not add JCenter dependency when building RN apps

twometresteve:
## Goal

Respond to yesterday's JCenter outage.

## Design

The intention of the code in `packages/react-native/android/build.gradle` was to shield our test fixture builds from JCenter outages, but it also unnecessarily adding a dependency on JCenter for app builds in general.

## Changeset

The build script snippet is now moved to only reside within our test fixture build scripts.  The get all fixtures to build on CI during the (potentially partial) outage I found it necessary to replace `jcenter()` with `mavenCentral()` in a number of locations, as well as removing Flipper from the builds (we don't need it anyway).

## Testing

Covered by CI.


gingerbenw (approved):
if it's building and passing I'm happy - but review from @lemnik would be appreciated.

PR #1861, opened 2022-11-04
[PLAT-6697] Prevent excess change detection cycles

gingerbenw:
## Goal

Improve performance of angular applications when calling `Bugsnag.notify`

## Design

Leverage built in angular zones to move `notify` out of unnecessary change detection


packages/plugin-angular/src/index.ts
```diff
@@ -42,6 +53,24 @@ export class BugsnagErrorHandler extends ErrorHandler {
...
+    const originalNotify = client._notify
+
+    client._notify = function (event) {
+      if (isNgZoneEnabled) {
+        // The `Zone.root.run` basically will run all asynchronous tasks in the most parent zone.
+        // The Angular's zone is forked from the `Zone.root`. In this case, `zone.js` won't
+        // trigger change detection, and `ApplicationRef.tick()` will not be run.
+        // Caretaker note: we're using `Zone.root` as opposed to `NgZone.runOutsideAngular` since this
+        // will require injecting the `NgZone` facade. That will create a breaking change for
+        // projects already using the `BugsnagErrorHandler`.
```
imjoehaines, line 65:
```suggestion
        // run notify in the root zone to avoid triggering change detection
```

packages/plugin-angular/src/index.ts
```diff
@@ -1,6 +1,17 @@
 import { ErrorHandler, Injectable } from '@angular/core'
 import Bugsnag, { Client, Plugin } from '@bugsnag/js'
 
+// That's the `global.Zone` exposed when the `zone.js` package is used.
+declare const Zone: any
```
imjoehaines, line 5:
```suggestion
// angular uses zones to watch for changes in asynchronous tasks so it can
// update the UI in response
// Bugsnag uses a lot of asynchronous tasks when notifying, which trigger change
// detection multiple times. This causes a potential performance problem, so we
// need to run `notify` outside of the current zone if zones are being used
// see https://angular.io/guide/zone
declare const Zone: any
```

packages/plugin-angular/src/index.ts
```diff
@@ -1,6 +1,17 @@
...
+
+// There're 2 types of Angular applications:
+// 1) zone-full (by default)
+// 2) zone-less
+// The developer can avoid importing the `zone.js` package and tells Angular that
+// he is responsible for running the change detection by himself. This is done by
+// "nooping" the zone through `CompilerOptions` when bootstrapping the root module.
```
imjoehaines, line 12:
```suggestion
// zones are optional, so we need to detect if they are being used
// see https://angular.io/guide/zone#noopzone
```

packages/plugin-angular/src/index.ts
```diff
@@ -42,6 +53,24 @@ export class BugsnagErrorHandler extends ErrorHandler {
...
+        // The Angular's zone is forked from the `Zone.root`. In this case, `zone.js` won't
+        // trigger change detection, and `ApplicationRef.tick()` will not be run.
+        // Caretaker note: we're using `Zone.root` as opposed to `NgZone.runOutsideAngular` since this
+        // will require injecting the `NgZone` facade. That will create a breaking change for
+        // projects already using the `BugsnagErrorHandler`.
+        Zone.root.run(() => {
+          originalNotify(event)
+        })
+      } else {
+        originalNotify(event)
```
imjoehaines, line 70:
```suggestion
        // if zones are not enabled, change detection will not run anyway
        originalNotify(event)
```

imjoehaines (commented):
Changes look good, just not sure that the comments are that useful — I've tried to make them a bit clearer for future & link to the docs but feel free to reword them some more if you want

imjoehaines (approved):

packages/plugin-angular/src/index.ts
```diff
@@ -42,6 +54,20 @@ export class BugsnagErrorHandler extends ErrorHandler {
...
+
+    client._notify = function (event) {
+      if (isNgZoneEnabled) {
+        // run notify in the root zone to avoid triggering change detection
+        Zone.root.run(() => {
+          originalNotify(event)
+        })
+      } else {
+        // if zones are not enabled, change detection will not run anyway
+        originalNotify(event)
```
imjoehaines, line 67: [`_notify` actually has more than one argument](https://github.com/bugsnag/bugsnag-js/blob/next/packages/core/client.js#L292); probably best to use `arguments` to avoid having to update multiple places if they change

```suggestion
    client._notify = function () {
      const notifyArguments = arguments
      if (isNgZoneEnabled) {
        // run notify in the root zone to avoid triggering change detection
        Zone.root.run(() => {
          originalNotify(notifyArguments)
        })
      } else {
        // if zones are not enabled, change detection will not run anyway
        originalNotify(notifyArguments)
```

imjoehaines (changes requested):

imjoehaines (approved):

PR #1866, opened 2022-11-09
Update browser delivery tests to respond with a log, add delivery tests to node

Cawllec:
## Goal

This updates the delivery browser tests to:
- Wait for several seconds to ensure the delivery process is not repeated
- Expect a log back from the page showing receipt of the rejection
This is currently very flakey on IE10, hence this test being skipped for the time being and a separate ticket will be created for it.

EDIT - Now implements the changes for node as well


djskinner (approved):
looks good. I'm going to fold this into my PR and assess next steps

PR #1871, opened 2022-11-17
Update node maze-runner version to v7

Cawllec:


dockerfiles/Dockerfile.node
```diff
@@ -21,7 +21,7 @@ RUN npm pack --verbose packages/plugin-koa/
 RUN npm pack --verbose packages/plugin-restify/
 
 # The maze-runner node tests
-FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:latest-v6-cli as node-maze-runner
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:latest-v7-cli as node-maze-runner
```
twometresteve, line 24: I don't know if anyone runs the Node tests locally, but I think we should also update the `Gemfile`in `test/node`.

Cawllec: I've updated that, as well as a typo which was triggering a non-fatal step missing

twometresteve (approved):
LGTM pending CI

PR #1874, opened 2022-11-18
Node improvements

djskinner:
## Goal

Node notifier:

- replace the deprecated domain API (used for adding context to errors) with async-hook-domain
- add the above capability to koa where it had been previously missing. This means that a wider class of errors will now receive request context in koa applications.
- Enable breadcrumbs (and use AsyncLocalStorage to support request context). This means that you can do `Bugsnag.leaveBreadcrumb` in an express/koa/restify request handler (or in any code initiated by the request) and get request context added to errors
- Use the same AsyncLocalStorage mechanism to add request context to unhandled promise rejections (where such rejection originate from a request)

## Work in progress

Implementations:
- [x] express
- [ ] koa
- [ ] restify
- [ ] plugin-contextualize

## Changeset

Breaking change: Minimum supported version of node is: [...]


packages/node/src/config.js
```diff
@@ -45,6 +50,14 @@ module.exports = {
...
     message: 'should be a function',
     validate: value => typeof value === 'function'
+  },
+  enabledBreadcrumbTypes: {
+    defaultValue: () => null,
+    message: `should be null or a list of available breadcrumb types (${NODE_BREADCRUMB_TYPES.join(',')})`,
+    validate: value => value === null || (isArray(value) && reduce(value, (accum, maybeType) => {
+      if (accum === false) return accum
+      return includes(NODE_BREADCRUMB_TYPES, maybeType)
+    }, true))
```
djskinner, line 60: This is overridden from the base config because the breadcrumb types will be different for node (I presume)

packages/plugin-node-unhandled-rejection/unhandled-rejection.js
```diff
@@ -3,16 +3,19 @@ module.exports = {
   load: client => {
     if (!client._config.autoDetectErrors || !client._config.enabledErrorTypes.unhandledRejections) return
     _handler = err => {
-      const event = client.Event.create(err, false, {
+      // if we are in an async context, use the client from that context
+      const c = (client._clientContext && client._clientContext.getStore()) ? client._clientContext.getStore() : client
```
djskinner, line 7: allows us to send request context for unhandled rejections originating from a request

test/node/features/express.feature
```diff
@@ -33,6 +33,9 @@ Scenario: an asynchronous thrown error in a route
   And the exception "message" equals "async"
   And the exception "type" equals "nodejs"
   And the "file" of stack frame 0 equals "scenarios/app.js"
+  And the event "request.url" equals "http://express/async"
+  And the event "request.httpMethod" equals "GET"
+  And the event "request.clientIp" is not null
```
djskinner, line 38: to verify that request context is sent in these scenarios (this is existing behaviour)

packages/node/src/config.js
```diff
@@ -1,8 +1,13 @@
 const { schema } = require('@bugsnag/core/config')
+const reduce = require('@bugsnag/core/lib/es-utils/reduce')
+const isArray = require('@bugsnag/core/lib/es-utils/is-array')
+const includes = require('@bugsnag/core/lib/es-utils/includes')
```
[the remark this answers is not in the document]
imjoehaines: Yeah, they'll be available

packages/node/src/config.js
imjoehaines: I think they'd be the same actually — even if we don't use them in plugins now, we might in future so it makes sense to me to allow users to configure them up front

djskinner: Yeah good point, in which case I can remove this and leave it to pick up the default config

PR #1878, opened 2022-11-30
[PLAT-8994] Support service workers

gingerbenw:
## Goal

To support the ability to use Bugsnag within a service worker, particularly use cases such as background scripts in chrome extensions 

## Design

The browser notifier is bundled with many plugins which are either unnecessary, or would need to be rewritten to support service workers, so a new simple worker notifier provides the opportunity to define worker related features off of a stable base

## Changeset

- new web-worker notifier package
- new delivery-fetch package, consumed by the web-worker notifier

## Testing

- Added unit tests for delivery-fetch and web-worker packages
- Added end to end tests for web-worker package


packages/worker/src/notifier.ts
```diff
@@ -0,0 +1,68 @@
...
+import delivery from '@bugsnag/delivery-fetch'
+
+const name = 'Bugsnag JavaScript'
+const version = '__VERSION__'
+const url = 'https://github.com/bugsnag/bugsnag-js'
+
+// extend the base config schema with some worker-specific options
+const schema = { ...coreSchema }
+
+// TODO: Fix interface - extend off ClientWithInternals?
```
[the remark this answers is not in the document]
imjoehaines: Actually I think this should be able to use the existing `BugsnagStatic` interface: https://github.com/bugsnag/bugsnag-js/blob/8d607ca4fd3fc1470207fdadca7aff78573a5542/packages/core/types/bugsnag.d.ts#L4-L8

If there are any worker-specific options then you'd need to add a new interface extending `BugsnagStatic`, e.g. like node: https://github.com/bugsnag/bugsnag-js/blob/8d607ca4fd3fc1470207fdadca7aff78573a5542/packages/node/types/bugsnag.d.ts#L14-L17

packages/worker/src/notifier.ts
```diff
@@ -0,0 +1,68 @@
...
+interface Notifier {
+  _client: Client | null
+  createClient: (opts: {}) => Client
+  start: (opts: {}) => Client
+  isStarted: () => Boolean
+}
+
+const Bugsnag: Notifier = {
+  _client: null,
+  createClient: (opts: {}) => {
```
imjoehaines, line 24: I think `opts` would need to be the `Config` type from core?

packages/worker/src/notifier.ts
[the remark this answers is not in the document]
imjoehaines: Also any new types we add here might also need to be exported for users to use, e.g. if we needed a `WorkerBugsnagStatic` then users might need to reference that

packages/worker/src/notifier.ts
```diff
@@ -0,0 +1,68 @@
+
+import type { Config, Plugin } from '@bugsnag/core'
+import Client from '@bugsnag/core/client'
+import { schema as coreSchema } from '@bugsnag/core/config'
+import delivery from '@bugsnag/delivery-fetch'
+
+const name = 'Bugsnag JavaScript'
+const version = '__VERSION__'
```
imjoehaines, line 8: The translation from this to the actual version is done by [`browserify-versionify`](https://www.npmjs.com/package/browserify-versionify) here: https://github.com/bugsnag/bugsnag-js/blob/8d607ca4fd3fc1470207fdadca7aff78573a5542/bin/bundle#L16

So we'd need a webpack equivalent for replacing this when it's built

packages/worker/README.md
```diff
@@ -0,0 +1,7 @@
+# @bugsnag/worker
+
+This package contains the worker implementation of the Bugsnag notifier for JavaScript. The normal use case is to install this package via `@bugsnag/js`, but you can install it directly if you want to.
+
+## License
+
+This package is free software released under the MIT License. See [LICENSE.txt](./LICENSE.txt) for details.
```
imjoehaines, line 7: This file is missing — it's the same license in every package so you can just copy paste it

packages/worker/README.md
```diff
@@ -0,0 +1,7 @@
+# @bugsnag/worker
+
+This package contains the worker implementation of the Bugsnag notifier for JavaScript. The normal use case is to install this package via `@bugsnag/js`, but you can install it directly if you want to.
```
imjoehaines, line 3:
```suggestion
This package contains the worker implementation of the Bugsnag notifier for JavaScript.
```

packages/delivery-fetch/test/delivery.test.ts
```diff
@@ -0,0 +1,37 @@
+import delivery from '../src/delivery'
+import type { Client } from '@bugsnag/core'
+
+// Prevent typescript from complaining about fetch not existing on global object
+const globalAny: any = global
+
+describe('delivery:fetch', () => {
+  it('sends events successfully', done => {
```
imjoehaines, line 8: Can we add more tests?

packages/delivery-fetch/package.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "@bugsnag/delivery-fetch",
```
imjoehaines, line 2: This package is missing a readme & license

packages/worker/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "devDependencies": {
+        "ts-loader": "^9.4.1",
+        "typescript": "^4.9.3",
+        "webpack": "^5.75.0",
+        "webpack-cli": "^5.0.0"
+    },
+    "dependencies": {
+        "@bugsnag/core": "^7.18.0",
+        "@bugsnag/delivery-fetch": "^7.18.0"
+    }
```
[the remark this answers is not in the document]
gingerbenw: I think it depends on how we use it. if we bundle the whole script with webpack, it only needs to be a devDependency

packages/delivery-fetch/README.md
```diff
@@ -0,0 +1,7 @@
+# @bugsnag/deliery-fetch
```
imjoehaines, line 1:
```suggestion
# @bugsnag/delivery-fetch
```

packages/worker/src/prevent-discard.js
```diff
@@ -0,0 +1,14 @@
...
+  load: client => {
+    client.addOnError(event => {
+      event.errors[0].stacktrace = event.errors[0].stacktrace.map(function (frame) {
+        frame.file = frame.file.replace(/chrome-extension:/g, 'chrome_extension:')
+        frame.file = frame.file.replace(/moz-extension:/g, 'moz_extension:')
+        frame.file = frame.file.replace(/safari-extension:/g, 'safari_extension:')
+        frame.file = frame.file.replace(/safari-web-extension:/g, 'safari_web_extension:')
+        return frame
+      })
+    }, true)
```
imjoehaines, line 12: This can be simpler & more efficient by mutating in place and using 1 regex:

```suggestion
    const extensionRegex = /^(chrome|moz|safari|safari-web)-extension:/

    client.addOnError(event => {
      event.errors[0].stacktrace.forEach(function (frame) {
        frame.file = frame.file.replace(extensionRegex, '$1_extension:')
      })
    }, true)
```

packages/delivery-fetch/delivery.js
```diff
@@ -0,0 +1,44 @@
...
+const delivery = (client) => ({
+  sendEvent: (event, cb = () => {}) => {
+    const url = client._config.endpoints.notify
+
+    global.fetch(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'Bugsnag-Api-Key': client._config.apiKey,
+        'Bugsnag-Payload-Version': '4',
```
gareththackeray, line 12: Minor: Would it make sense to make this a constant somewhere so it can be shared between here and XHR (and any other places)?

packages/worker/README.md
```diff
@@ -0,0 +1,7 @@
+# @bugsnag/worker
+
+This package contains the worker implementation of the Bugsnag notifier for JavaScript.
```
gareththackeray, line 3: Is "worker" a specific enough term?  If this plugin works for [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) in general then IMO we should say "Web Worker" and if it's specifically for Service Workers we should say "Service Worker".

As an aside, if it does work for Web Workers, do you have a feeling for 

* how that works?
* whether it's useful?

gareththackeray: TBC I'm talking about the description rather than the package name I think that is fine.

gingerbenw: the test fixture is using a worker, as opposed to a service worker. so I'm happy to change it to "web worker / service worker"

it should work in much the same way between web/service workers and should also handle uncaught errors in a web worker.

packages/delivery-fetch/delivery.js
gingerbenw: I was questioning this as well actually - sounds like a good idea to me.

gingerbenw:
Apologies, got a lot of commits on this one now - but will squash when I merge into the integration branch

packages/web-worker/src/unhandled.js
```diff
@@ -0,0 +1,88 @@
+const ERROR_NAME = 'worker onerror'
+
+/*
+ * Automatically captures unhandled worker errors
+ */
+module.exports = {
```
imjoehaines, line 6: Should be able to use the window onerror plugin by passing `self` to it when importing it in the notifier

packages/delivery-fetch/delivery.js
```diff
@@ -0,0 +1,44 @@
+import payload from '@bugsnag/core/lib/json-payload'
+
+const delivery = (client) => ({
+  sendEvent: (event, cb = () => {}) => {
+    const url = client._config.endpoints.notify
+
+    global.fetch(url, {
```
imjoehaines, line 7: Other plugins that use globals allow passing a replacement when they are imported, e.g. in `plugin-window-onerror`:

https://github.com/bugsnag/bugsnag-js/blob/0599faeb4019e571bfd7111c2663b5ddcbb78901/packages/plugin-window-onerror/onerror.js#L5

So we should pass `fetch` to this function with a default value of `fetch` (i.e. the global)

packages/web-worker/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Within a web worker, unhandled errors will also bubble up to the script that initialised the worker, so if you are also using Bugsnag in the parent script, you may wish to prevent these errors from being reported a second time
+
+```js
+const worker = new Worker('worker.js', { type: 'module' })
+worker.onerror = function (e) {
+    e.preventDefault()
+}
+```
+
```
gingerbenw, line 26: @tomlongridge could you check over wording of this section please?

packages/delivery-fetch/delivery.js
gingerbenw: will raise a separate ticket to make this enhancement outside the scope of this PR

packages/web-worker/src/config.js
```diff
@@ -0,0 +1,28 @@
+const { schema } = require('@bugsnag/core/config')
+const map = require('@bugsnag/core/lib/es-utils/map')
+const assign = require('@bugsnag/core/lib/es-utils/assign')
+
+module.exports = {
+  appType: {
+    ...schema.appType,
+    defaultValue: () => 'worker-js'
```
imjoehaines, line 8: this is all one word in other notifiers, e.g. `browserjs` or `electronrendererjs`

```suggestion
    defaultValue: () => 'workerjs'
```

packages/web-worker/src/prevent-discard.js
```diff
@@ -0,0 +1,14 @@
+const extensionRegex = /^(chrome|moz|safari|safari-web)-extension:/
+
+module.exports = {
+  name: 'preventDiscard',
+  load: client => {
+    client.addOnError(event => {
```
imjoehaines, line 6: As discussed earlier, this isn't appropriate for non-browser extensions — we'll need to figure out how to handle that

packages/web-worker/src/unhandled.js
gareththackeray: Do you mean we can get rid of all the copy-pasted code below?  I think that would be good as some of the comments are strange in the context of a worker.

packages/web-worker/README.md
```diff
@@ -0,0 +1,53 @@
...
+**This package is still in the beta phase, so should be used with caution in production apps.**
+
+We would welcome feedback on this package via GitHub or directly to our [Support](mailto:support@bugsnag.com) team.
+
+## Features and limitations
+
+This early release offers basic functionality for web workers and service workers, with the intent to support Chrome extension development using manifest v3. Using this library, you will be able to:
+
+- notify errors from within service workers and web workers, including browser extensions
+- notify unhandled errors from service workers and web workers, excluding Chrome browser extensions
```
gareththackeray, line 14:
```suggestion
- detect and automatically notify unhandled errors from service workers and web workers, excluding Chrome browser extensions
```

packages/web-worker/README.md
```diff
@@ -0,0 +1,53 @@
...
+function myFunction() {
+    try {
+        // something that will throw an error
+    } catch (err) {
+        Bugsnag.notify(err)
+    }
+}
+```
+
+Within a web worker, unhandled errors will also bubble up to the script that initialised the worker, so if you are also using Bugsnag in the parent script, you may wish to prevent these errors from being reported a second time:
```
gareththackeray, line 42:
```suggestion
Within a web worker, unhandled errors will also bubble up to the script that initialized the worker, so if you are also using Bugsnag in the parent script, you may wish to prevent these errors from being reported a second time:
```

gareththackeray, line 42: > Within a web worker

Is this excluding service workers?  If not and it's all types of workers, lets just remove it so that it begins:

> Unhandled errors ...

gingerbenw: this excludes service workers - they aren't run within the context of the parent app

packages/web-worker/README.md
```diff
@@ -0,0 +1,53 @@
...
+We would welcome feedback on this package via GitHub or directly to our [Support](mailto:support@bugsnag.com) team.
+
+## Features and limitations
+
+This early release offers basic functionality for web workers and service workers, with the intent to support Chrome extension development using manifest v3. Using this library, you will be able to:
+
+- notify errors from within service workers and web workers, including browser extensions
+- detect and automatically notify unhandled errors from service workers and web workers, excluding Chrome browser extensions
+
+Due to limitations in the Chrome Runtime API, it is not possible to catch unhandled errors so you will need to catch these from within your code and notify these errors manually.
```
gareththackeray, line 16: Is this paragraph pertaining to Chrome browser extensions?  If so I guess it's redundant?

gingerbenw: correct, `chrome.runtime` is used in service workers hosting background tasks for browser extensions, and we are unable to catch unhandled errors here. perhaps we could specify `chrome.runtime` in the above "excluding Chrome browser extensions" ?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Enhancements
 
 - (react-native) Update bugsnag-android from v5.28.1 to [v5.28.3](https://github.com/bugsnag/bugsnag-android/blob/master/CHANGELOG.md#5283-2022-11-16)
+- (delivery-fetch) Create fetch based notifier delivery package [#1878](https://github.com/bugsnag/bugsnag-js/pull/1878)
+- (web-worker) Create web-worker notifier package to facilitate changes in google chrome extension development using [manifest v3](https://developer.chrome.com/docs/extensions/mv3/intro) [#1878](https://github.com/bugsnag/bugsnag-js/pull/1878)
```
gingerbenw, line 9: @gareththackeray happy to link to the chrome documentation here?

packages/web-worker/README.md
gareththackeray: I think we should either 

1. just remove this paragraph entirely
2. ensure that the link is made explicit (e.g. by saying "Due to limitations in the Chrome Runtime API, which is used for service workers hosting background tasks in Chrome browser extensions")
3. remove this paragraph, but append "due to limitations  in the Chrome Runtime API" to the bullet point

gingerbenw: option 3 sounds clean to me. happy @tomlongridge ?

tomlongridge: Agree

CHANGELOG.md
tomlongridge: IMO, I wouldn't mention browser extensions here at all. The feature is for web workers and has wider users beyond the browser extension part.

PR #1879, opened 2022-12-02
Fix parts of request metadata being missing from some events

imjoehaines:
## Goal

Request metadata is stored in two places on an Event — some goes in `event.request` and other bits are added as metadata. This is because the Bugsnag API accepts some request data as a top-level field, but we add more data than the API supports so these bits have to go in metadata

We were adding this metadata on the `requestClient` instead of the `event` and were doing so in an on error callback, which is too late to affect the already-created event. This would affect any subsequent events, though, so the data was present in some cases. This was broken when we moved from adding request data when our middleware was called to adding it in a callback, which was done to ensure we reported the final state of the request as other middleware could mutate it

### DIFF
--- a/packages/plugin-restify/src/restify.js
+++ b/packages/plugin-restify/src/restify.js
@@ -27,7 +27,7 @@ module.exports = {
       requestClient.addOnError((event) => {
         const { request, metadata } = getRequestAndMetadataFromReq(req)
         event.request = { ...event.request, ...request }
-        requestClient.addMetadata('request', metadata)
+        event.addMetadata('request', metadata)
       }, true)
 
       if (!client._config.autoDetectErrors) return next()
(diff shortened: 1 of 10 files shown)


packages/plugin-restify/src/restify.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
       requestClient.addOnError((event) => {
         const { request, metadata } = getRequestAndMetadataFromReq(req)
         event.request = { ...event.request, ...request }
-        requestClient.addMetadata('request', metadata)
+        event.addMetadata('request', metadata)
```
gingerbenw, line 30: are there any side effects of removing the metadata from the requestClient?

imjoehaines: Hmm, yes actually though I'm not sure it could be a problem in practice

The way it worked before would mean users could get this data from the request client after an error had been reported, which wouldn't work anymore

djskinner: Given this is a bug it seems reasonable to change the behaviour here (without a significant version bump) as anyone doing this would be relying on a bug

djskinner (approved):

packages/plugin-restify/src/restify.js
imjoehaines: Yeah, especially because even if someone was relying on this they'd already have to handle the metadata being missing sometimes

PR #1880, opened 2022-12-02
Add unit tests for clone-client

imjoehaines:
## Goal

Added some unit tests for clone-client in preparation for some upcoming changes

In practice a lot of this is already covered by other tests that use clone-client, but hopefully these will allow easier debugging if something does break

### DIFF
--- a/packages/core/client.d.ts
+++ b/packages/core/client.d.ts
@@ -50,7 +50,8 @@ export default class ClientWithInternals<T extends Config = Config> extends Clie
   _user: User
 
   _metadata: { [key: string]: any }
-  _features: FeatureFlag | null[]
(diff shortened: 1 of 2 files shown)


packages/core/client.d.ts
```diff
@@ -50,7 +50,8 @@ export default class ClientWithInternals<T extends Config = Config> extends Clie
   _user: User
 
   _metadata: { [key: string]: any }
-  _features: FeatureFlag | null[]
+  _features: Array<FeatureFlag | null>
+  _featuresIndex: { [key: string]: number }
```
imjoehaines, line 54: `_featuresIndex` needed adding here to get the unit test to compile

PR #1886, opened 2022-12-07
Update RN pipeline to use Maze Runner v7

twometresteve:
## Goal

Update RN pipeline to use Maze Runner v7.

## Design

This was a fairly standard upgrade to v7, following the path already walked by the likes of bugsnag-android.  

I did find, however, that at least one of the test scenarios (`Unhandled native error`) became flaky when moving to the new `terminate_app`/`terminate_app` Appium driver calls.  This will need further investigation, so for now I have just stuck with the "old" `close_app` and `launch_app` calls (they seem to work absolutely fine and just output a deprecation warning).  The scenario identified then passed 60/60 times for me when it was flaking at least 1/10 times.

## Testing

Covered by CI.


test/react-native/features/steps/react-native-steps.rb
```diff
@@ -16,27 +16,36 @@
...
+  max_attempts = 20
+  attempts = 0
+  state = get_app_state
+  until (attempts >= max_attempts) || state == expected_state
+    attempts += 1
+    state = get_app_state
+    sleep 0.5
+  end
+  $logger.warn "App state #{state} instead of #{expected_state} after 10s" unless state == expected_state
+  state
```
imjoehaines, line 32: I think this is equivalent:

```suggestion
  wait = Maze::Wait.new(interval: 0.5, timeout: 10)
  state = get_app_state

  success = wait.until do
    state = get_app_state
    state == expected_state
  end

  $logger.warn "App state #{state} instead of #{expected_state} after 10s" unless success

  state
```

imjoehaines (approved):

test/react-native/features/steps/react-native-steps.rb
twometresteve: Thanks - I shall try to remember to update this the next time I'm there, rather than suffer a full rebuild on CI right now.

Issue #1875, opened 2022-11-27
Express/Koa/Restify plugin adds new error callbacks on each request without removing old callbacks, resulting in the wrong request being reported and timeouts


hansottowirtz:
### Describe the bug

These lines are run on every request in the express plugin:

https://github.com/bugsnag/bugsnag-js/blob/8d607ca4fd3fc1470207fdadca7aff78573a5542/packages/plugin-express/src/express.js#L22-L32

When `autoTrackSessions: true` (which is the default), the same `requestClient` is used as for previous requests. This means that a new callback is added to the error callbacks array on every request. This has the following consequences:

- The wrong request is reported in the Request tab on Bugsnag (the body of the first request is always used, regardless of the request that caused the error)
- After many requests, the Bugsnag flush times out (we think this is the reason)

Is it possible that old callbacks are not removed when they should be?

### Steps to reproduce

I think any express/koa/restify app without `autoTrackSessions: false` will have this problem.

### Environment

We're using the Express plugin together with the AWS Lambda plugin.

* Bugsnag version: 7.17.0

PR #1887, opened 2022-12-08
Clone clients in AWS Lambda session implementation when a server plugin is used

imjoehaines:
## Goal

This PR fixes a conflict between the AWS Lambda plugin and server plugins (Express, Koa & Restify) where the server plugins expect the session delegate to create a cloned client but the AWS Lambda session delegate doesn't do that

This is a bit more complicated than simply cloning the client in the AWS Lambda session delegate — doing so causes unexpected behaviour when _not_ using a server plugin as the client created by `Bugsnag.start` would not have access to the session information and therefore session data doesn't get attributed to events properly

See #1875

### DIFF
--- a/packages/plugin-aws-lambda/test/session.test.ts
+++ b/packages/plugin-aws-lambda/test/session.test.ts
@@ -56,59 +56,57 @@ describe('plugin: aws lambda sessions', () => {
     })
   })
 
-  ;['express', 'koa', 'restify'].forEach(serverPlugin => {
-    describe(`with server plugin (${serverPlugin})`, () => {
-      it('can start a session', () => {
-        const client = createClient(serverPlugin)
+  describe.each(['express', 'koa', 'restify'])('with server plugin (%s)', serverPlugin => {
+    it('can start a session', () => {
+      const client = createClient(serverPlugin)
 
-        expect(client._session).toBeNull()
+      expect(client._session).toBeNull()
 
-        const cloned = client.startSession()
+      const cloned = client.startSession()
 
-        expect(cloned).not.toBe(client)
-        expect(client._session).toBeNull()
-        expect(cloned._session).toBeInstanceOf(Session)
-        // @ts-ignore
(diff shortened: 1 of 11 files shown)


test/aws-lambda/features/promise-rejection.feature
```diff
@@ -83,5 +83,4 @@ Scenario: promise rejections are reported when using serverless-express
     Then the session is valid for the session reporting API version "1" for the "Bugsnag Node" notifier
     And the session "id" is not null
     And the session "startedAt" is a timestamp
-    And the event "session.events.handled" equals 0
-    And the event "session.events.unhandled" equals 1
+    And the event "session.events" is null
```
imjoehaines, line 86: This promise rejection isn't associated with a session anymore, because the session is started by the Express plugin in the request but an unhandled promise rejection is caught globally and therefore by a different client with no knowledge of the session

packages/plugin-aws-lambda/test/session.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    it('does not clone the client with resumeSession if there is no active session', () => {
+      const client = createClient()
+      const cloned = client.resumeSession()
+
+      expect(cloned).toBe(client)
+      expect(client._session).toBeInstanceOf(Session)
+    })
+  })
+
+  ;['express', 'koa', 'restify'].forEach(serverPlugin => {
```
djskinner, line 59:
```suggestion
  ['express', 'koa', 'restify'].forEach(serverPlugin => {
```
?

djskinner (approved):

packages/plugin-aws-lambda/test/session.test.ts
imjoehaines: You can't start a line with `[` without a semicolon before it, otherwise it gets interpreted like this:

```js
describe('', function () {})['express', ...].forEach(...)
```

but in this case it can be an `it.each` and sidestep that problem

PR #1893, opened 2022-12-21
[PLAT-7663] Fix state/metadata set in renderer config not being cleared 

yousif-bugsnag:
## Goal

Any state/metadata that was added in a renderer's configuration could no be removed from that renderer

This is because we send the renderer config values to the main process, but they were also stored locally in the renderer client (because it’s using the same code as all other JS platforms)

This means both the renderer and main processes had a copy of the state, but methods like `clearMetadata` and `clearFeatureFlags` only talk to the main process

This affected `user`, `context` `metadata` and `featureFlags` config

## Changeset

Updated `plugin-electron-renderer-client-state-updates`  to clear any state from the renderer client once it has been synched to the main process

## Testing

- Manually tested clearing `user`, `context` `metadata` and `featureFlags` set in renderer config
- Added a unit test to ensure that this state is not stored in the renderer client 
- Added integration tests to ensure that any state set in the renderer config can be cleared and will not be attached to subsequent events from that renderer

### DIFF
--- a/test/electron/features/feature-flags.feature
+++ b/test/electron/features/feature-flags.feature
@@ -94,10 +94,10 @@ Scenario: feature flags are attached to unhandled errors in a renderer process
         | Bugsnag-Integrity | {BODY_SHA1}                      |
     And the event contains the following feature flags:
         | featureFlag                | variant    |
-        | from renderer config       |            |
         | from main config 1         | 1234       |
         | from main config 2         |            |
         | from main at runtime       | runtime 1  |
+        | from renderer config       |            |
         | from renderer at runtime 1 | runtime    |
         | from renderer at runtime 2 |            |
         | from renderer on error     | on error   |
@@ -119,10 +119,10 @@ Scenario: feature flags are attached to handled errors in a renderer process
         | Bugsnag-Integrity | {BODY_SHA1}                      |
     And the event contains the following feature flags:
         | featureFlag                | variant    |
-        | from renderer config       |            |
         | from main config 1         | 1234       |
         | from main config 2         |            |
         | from main at runtime       | runtime 1  |
+        | from renderer config       |            |
         | from renderer at runtime 1 | runtime    |
         | from renderer at runtime 2 |            |
         | from renderer on error     | on error   |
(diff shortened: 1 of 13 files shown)


test/electron/features/feature-flags.feature
```diff
@@ -94,10 +94,10 @@ Scenario: feature flags are attached to unhandled errors in a renderer process
         | Bugsnag-Integrity | {BODY_SHA1}                      |
     And the event contains the following feature flags:
         | featureFlag                | variant    |
-        | from renderer config       |            |
         | from main config 1         | 1234       |
         | from main config 2         |            |
         | from main at runtime       | runtime 1  |
+        | from renderer config       |            |
```
yousif-bugsnag, line 100: One effect of this change is that feature flags added in the main process config will now always appear before those added in the renderer config. This is because all feature flags (and their sort order) are now only stored in the main process.

When an error occurs in the renderer process it receives payload data (including all feature flags) from the main process.

Previously the renderer client stored its own features, and its own separate `client._featuresIndex` object containing only the renderer's feature flags. So when feature flags were merged back across from the main process they would be sorted according to the renderer's `client._featuresIndex`.

test/electron/fixtures/app/index.html
```diff
@@ -59,5 +59,10 @@ <h2>actions</h2>
         <li><a href="#" id="main-process-clear-feature-flags-now" onclick="RunnerAPI.mainProcessClearFeatureFlagsNow()">immediately clear feature flags in main process</a></li>
         <li><a href="#" id="renderer-clear-feature-flags-now">immediately clear feature flags in renderer</a></li>
     </ul>
+    <ul>
+        <li><a href="#" id="renderer-clear-user">Clear user data set in renderer config</a></li>
+        <li><a href="#" id="renderer-clear-context">Clear context set in renderer config</a></li>
+        <li><a href="#" id="renderer-clear-metadata">Clear metadata set in renderer config</a></li>
```
imjoehaines, line 65: Actually, maybe this is clearer for the metadata link?

```suggestion
        <li><a href="#" id="renderer-clear-metadata">Clear "renderer" metadata section</a></li>
```

test/electron/features/feature-flags.feature
imjoehaines: So this is fixing a bug really — feature flags are now entirely handled in the main process so are always chronologically ordered

yousif-bugsnag: yep!

test/electron/features/renderer-config.feature
```diff
@@ -16,3 +16,25 @@ Feature: Setting config options in renderers
             | property | config                      |
             | appType  | { "appType": "real great" } |
             | codeBundleId  | { "codeBundleId": "1.0.0-r0123" } |
+            | user  | { "user": { "id": "3", "name": "Bugs Nag", "email": "bugs.nag@bugsnag.com" } } |
+            | context  | { "context": "renderer context" } |
+            | metadata  | { "metadata": { "renderer": { "key": "value" } } } |
+
+    Scenario Outline: Clearing config options set in renderer config
```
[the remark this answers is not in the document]
yousif-bugsnag: Oh sorry, that test is checking that those values do exist - I noticed those config options were missing from that existing scenario outline so I added them in. Maybe I should have left that out as it's not directly related to this fix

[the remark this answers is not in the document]
yousif-bugsnag: But I think this does apply to the `clear-context` test in the new scenario as I think that will revert to a default context if it's cleared. Just trying to figure out what that would be...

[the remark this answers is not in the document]
imjoehaines: Yeah, that's what I'm commenting on — the "Clearing config options set in renderer config" scenario

- `clear-user.json` has a `user` object with an ID, but it will still pass if there's an `email` & `name`
- `clear-context.json` doesn't have a `context` key, so it will pass regardless of the value
- `clear-metadata.json` has a `metadata` object with some automatically collected stuff in, but it will still pass if there's a `renderer` section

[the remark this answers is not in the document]
yousif-bugsnag: Aah yes I see what you mean. Updated.

test/electron/fixtures/events/renderer/config/clear-context.json
```diff
@@ -0,0 +1,93 @@
...
+        "id": "{REGEX:[0-9a-f]{64}}",
+        "freeMemory": "{TYPE:number}",
+        "time": "{TIMESTAMP}",
+        "totalMemory": "{TYPE:number}",
+        "osVersion": "{REGEX:\\d+\\.\\d+}"
+      },
+      "user": {
+        "id": "{REGEX:[0-9a-f]{64}}"
+      },
+      "context": "Runner",
```
yousif-bugsnag, line 34: If you clear the renderer context it reverts to the default context, which is the html `<title>` tag

test/electron/fixtures/app/index.html
yousif-bugsnag: Updated

imjoehaines (approved):

PR #1896, opened 2023-01-09
Create web-worker package

gingerbenw:
## Goal

To support the ability to use bugsnag within a web worker or service worker, including background scripts in chrome extensions 

## Design

The browser notifier is bundled with many plugins which are either unnecessary, or would need to be rewritten to support service workers, so a new simple worker notifier provides the opportunity to define worker related features off of a stable base

## Changeset

- new web-worker notifier package

## Testing

- Added unit tests web-worker package
- Added end to end tests for web-worker package

### DIFF
--- a/packages/plugin-browser-device/device.js
+++ b/packages/plugin-browser-device/device.js
@@ -1,4 +1,3 @@
-const cuid = require('@bugsnag/cuid')
 const assign = require('@bugsnag/core/lib/es-utils/assign')
 const BUGSNAG_ANONYMOUS_ID_KEY = 'bugsnag-anonymous-id'
 
@@ -14,6 +13,7 @@ const getDeviceId = () => {
       return id
     }
 
+    const cuid = require('@bugsnag/cuid')
     id = cuid()
 
     storage.setItem(BUGSNAG_ANONYMOUS_ID_KEY, id)
@@ -27,7 +27,7 @@ const getDeviceId = () => {
 /*
  * Automatically detects browser device details
  */
-module.exports = (nav = navigator, screen = window.screen) => ({
+module.exports = (nav = navigator, screen = window && window.screen) => ({
   load: (client) => {
     const device = {
       locale: nav.browserLanguage || nav.systemLanguage || nav.userLanguage || nav.language,
@@ -37,10 +37,14 @@ module.exports = (nav = navigator, screen = window.screen) => ({
     if (screen && screen.orientation && screen.orientation.type) {
       device.orientation = screen.orientation.type
     } else {
-      device.orientation =
+      try {
+        device.orientation =
         document.documentElement.clientWidth > document.documentElement.clientHeight
           ? 'landscape'
           : 'portrait'
+      } catch (err) {
+        // document is undefined (inside a worker)
+      }
     }
 
     if (client._config.generateAnonymousId) {
--- a/packages/plugin-browser-device/device.js
+++ b/packages/plugin-browser-device/device.js
@@ -1,7 +1,7 @@
 const assign = require('@bugsnag/core/lib/es-utils/assign')
 const BUGSNAG_ANONYMOUS_ID_KEY = 'bugsnag-anonymous-id'
 
-const getDeviceId = (win = window) => {
+const getDeviceId = (win) => {
   try {
     const storage = win.localStorage
 
@@ -38,7 +38,7 @@ module.exports = (nav = navigator, win = window) => ({
       device.orientation = win.screen.orientation.type
     } else if (win && win.document) {
       device.orientation =
-        window.document.documentElement.clientWidth > window.document.documentElement.clientHeight
+        win.document.documentElement.clientWidth > win.document.documentElement.clientHeight
           ? 'landscape'
           : 'portrait'
     }
--- a/packages/plugin-browser-device/device.js
+++ b/packages/plugin-browser-device/device.js
@@ -1,9 +1,9 @@
 const assign = require('@bugsnag/core/lib/es-utils/assign')
 const BUGSNAG_ANONYMOUS_ID_KEY = 'bugsnag-anonymous-id'
 
-const getDeviceId = () => {
+const getDeviceId = (win = window) => {
   try {
-    const storage = window.localStorage
+    const storage = win.localStorage
 
     let id = storage.getItem(BUGSNAG_ANONYMOUS_ID_KEY)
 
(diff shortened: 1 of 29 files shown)


packages/web-worker/src/notifier.js
```diff
@@ -0,0 +1,62 @@
...
+const schema = { ...coreSchema, ...config }
+
+export const Bugsnag = {
+  createClient: (opts) => {
+    // handle very simple use case where user supplies just the api key as a string
+    if (typeof opts === 'string') opts = { apiKey: opts }
+    if (!opts) opts = {}
+
+    // eslint-disable-next-line no-undef
+    const internalPlugins = [pluginClientIp, pluginBrowserDevice(navigator, null), pluginPreventDiscard, pluginWindowOnError(self)]
```
gingerbenw, line 25: `plugin-window-onerror` creates events with a component of `'window onerror'` - should we pass in `'worker onerror'` or similar?

packages/plugin-browser-device/device.js
```diff
@@ -14,6 +13,7 @@ const getDeviceId = () => {
       return id
     }
 
+    const cuid = require('@bugsnag/cuid')
```
imjoehaines, line 16: Have you checked what this gets compiled to? I think it's fine as all of our requires get flattened (so there's no actual runtime cost of this) but worth double checking that this won't do something unexpected

packages/web-worker/TODO.md
```diff
@@ -0,0 +1,6 @@
+- [x] Update existing packages to support worker notifier (window-onerror, plugin-browser-device)
```
imjoehaines, line 1: This file can go 🙂

packages/web-worker/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@bugsnag/web-worker",
+    "version": "7.18.0",
+    "description": "Bugsnag error reporter for JavaScript web workers and service workers",
```
imjoehaines, line 4:
```suggestion
    "description": "BugSnag error reporter for JavaScript web workers and service workers",
```

packages/web-worker/src/notifier.js
imjoehaines: Yeah, or maybe `self onerror` since that's the object with the `onerror` event? I'm not sure if that's more confusing than `worker onerror` though

packages/web-worker/types/notifier.ts
```diff
@@ -0,0 +1,18 @@
+import type { Client, Config, BugsnagStatic } from '@bugsnag/core'
+
+interface WorkerConfig extends Config {
+  collectUserIp?: boolean
+  generateAnonymousId?: boolean
+
```
imjoehaines, line 6:
```suggestion
```

packages/web-worker/src/notifier.js
gingerbenw: @gareththackeray, @tomlongridge - thoughts on the above?

imjoehaines: For context, this is only used as the name of the metadata tab when you pass a non-error to `notify` (it's the value of `component` in this snippet):

https://github.com/bugsnag/bugsnag-js/blob/aced3db494a4a779cb2911dbfbdd9e6c529c807c/packages/core/event.js#L228-L234

gareththackeray: `worker onerror` works for me

packages/plugin-browser-device/device.js
gingerbenw: this does indeed get flattened - it just changes the scope of the function here.

packages/plugin-browser-device/device.js
```diff
@@ -1,10 +1,9 @@
-const cuid = require('@bugsnag/cuid')
 const assign = require('@bugsnag/core/lib/es-utils/assign')
 const BUGSNAG_ANONYMOUS_ID_KEY = 'bugsnag-anonymous-id'
 
-const getDeviceId = () => {
+const getDeviceId = (win = window) => {
```
imjoehaines, line 4: This doesn't need a default since it will always be passed `win` from the main plugin function

```suggestion
const getDeviceId = (win) => {
```

packages/plugin-browser-device/device.js
```diff
@@ -27,24 +27,24 @@ const getDeviceId = () => {
...
 
-    if (screen && screen.orientation && screen.orientation.type) {
-      device.orientation = screen.orientation.type
-    } else {
+    if (win && win.screen && win.screen.orientation && win.screen.orientation.type) {
+      device.orientation = win.screen.orientation.type
+    } else if (win && win.document) {
       device.orientation =
-        document.documentElement.clientWidth > document.documentElement.clientHeight
+        window.document.documentElement.clientWidth > window.document.documentElement.clientHeight
```
imjoehaines, line 41:
```suggestion
        win.document.documentElement.clientWidth > win.document.documentElement.clientHeight
```

packages/web-worker/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+This early release offers basic functionality for web workers and service workers, with the intent to support Chrome extension development using manifest v3. Using this library, you will be able to:
+
+- notify errors from within service workers and web workers, including browser extensions
+- detect and automatically notify unhandled errors from service workers and web workers, excluding Chrome browser extensions due to limitations in the Chrome Runtime API.
+
+## Installation
+
+```bash
+npm i @bugsnag/web-worker@next
```
imjoehaines, line 15: This won't need the `@next` when it's released, or are we planning another prerelease before then?

```suggestion
npm i @bugsnag/web-worker
```

packages/web-worker/README.md
```diff
@@ -0,0 +1,47 @@
...
+function myFunction() {
+    try {
+        // something that will throw an error
+    } catch (err) {
+        Bugsnag.notify(err)
+    }
+}
+```
+
+Within a web worker, unhandled errors will also bubble up to the script that initialized the worker, so if you are also using BugSnag in the parent script, you may wish to prevent these errors from being reported a second time:
```
gareththackeray, line 36: It might be worth just replacing this paragraph and snippet with a link to the docs.

packages/web-worker/README.md
```diff
@@ -33,14 +35,7 @@ function myFunction() {
...
 
-Within a web worker, unhandled errors will also bubble up to the script that initialized the worker, so if you are also using BugSnag in the parent script, you may wish to prevent these errors from being reported a second time:
-
-```js
-const worker = new Worker('worker.js', { type: 'module' })
-worker.onerror = function (e) {
-    e.preventDefault()
-}
-```
+Check out the [documentation](https://docs.bugsnag.com/platforms/javascript/web-worker/)
```
gareththackeray, line 38:
```suggestion
For further information check out the [documentation](https://docs.bugsnag.com/platforms/javascript/web-worker/).
```

gareththackeray (approved):

imjoehaines (approved):

PR #1897, opened 2023-01-09
Pin electron-forge version to 6.0.0-beta.54 in test fixture

yousif-bugsnag:
## Goal

Pin the Electron test fixture’s electron-forge dependency to `v6.0.0-beta.54` because newer versions are incompatible with webpack 4 (which the test fixture uses)

## Testing

Relied on existing CI tests


twometresteve (commented):
No comments from me, but I don't feel particularly well placed to judge.

imjoehaines (approved):

PR #1898, opened 2023-01-10
replace node domain API with AsyncLocalStorage

djskinner:
## Goal

Replace all usages of the Domain API in the node notifier. The Domain API is deprecated and there are now stable alternative APIs for achieving the functionality that the Domain API was providing.

### Background

The main usage of the domain API was in `plugin-express` where we used the error handling capabilities to
a) add request context to async errors originating from a request handler
b) return an error response in the case of an error

Note that b) actually changes the behaviour of an application, which is not something we want to do in BugSnag, so this functionality has been removed i.e. BugSnag will no longer prevent the node process from exiting in the case on an uncaught exception.

## Design

- replace the deprecated domain API (used for adding context to errors) with AsyncLocalStorage. Instead of using Domain, `plugin-express` and `plugin-restify` now run the request handlers in the AsyncLocalStorage context with the cloned request client as the store value.
- add the above capability to `plugin-koa` where it had been previously missing. This means that a wider class of errors will now receive request context in koa applications.
- In `plugin-node-uncaught-exception` use the request context from `AsyncLocalStorage` if present. This replaces the request-context-adding functionality Domain provided. Such errors will now have a `"severityReason.type"` of `unhandledException ` instead of `unhandledErrorMiddleware `
- In `plugin-node-unhandled-rejection` use the request context from `AsyncLocalStorage` if present. This new capability allows including the request context for unhandled promise rejections that occur within a request. Works in node >= 16.

## Changeset

- Minimum supported node version is now v12.17.0

## Testing

- Manually tested end-to-end ensuring the request context is present in uncaught exceptions and unhandled promise rejections (node >= 16) from errors originating within request handlers
- Full automated e2e coverage of the changes and expected error capture scenarios


packages/plugin-koa/src/koa.js
```diff
@@ -27,7 +27,11 @@ module.exports = {
         event.addMetadata('request', metadata)
       }, true)
 
-      await next()
```
djskinner, line 30: koa plugin wasn't using domain previously so with this change we now improve error capturing in koa significantly

test/node/features/connect.feature
```diff
@@ -22,16 +22,18 @@ Scenario: a synchronous thrown error in a route
   And the event "request.httpMethod" equals "GET"
 
 Scenario: an asynchronous thrown error in a route
-  Then I open the URL "http://connect/async"
+  Then I open the URL "http://connect/async" tolerating any error
```
djskinner, line 25: tolerating any error is not required because the node process is allowed to crash

test/node/features/connect.feature
```diff
@@ -22,16 +22,18 @@ Scenario: a synchronous thrown error in a route
...
   And I wait to receive an error
   Then the error is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
   And the event "unhandled" is true
   And the event "severity" equals "error"
-  And the event "severityReason.type" equals "unhandledErrorMiddleware"
+  And the event "severityReason.type" equals "unhandledException"
```
djskinner, line 30: because the error is captured by the top-level handler where we are now able to add request context if available

test/node/features/restify.feature
```diff
@@ -1,3 +1,4 @@
+@skip_node_18
```
djskinner, line 1: the current restify fixture doesn't work on node 18. It has been verified manually that these tests pass with an updated fixture with the latest restify but it has been left to a separate task to adjust the e2e tests to add additional fixtures to cover restify on all supported node versions

djskinner:
In this PR an aws-lambda test is failing. Details and fix for that are in an upstream PR that will need review and merge before this PR passes: https://github.com/bugsnag/bugsnag-js/pull/1900

test/node/features/connect.feature
imjoehaines: Do you mean "is now required"?

test/node/features/express.feature
```diff
@@ -28,17 +28,20 @@ Scenario: a synchronous thrown error in a route
...
 
 Scenario: an asynchronous thrown error in a route
-  Then I open the URL "http://express/async"
+  Then I open the URL "http://express/async" tolerating any error
   And I wait to receive an error
   Then the error is valid for the error reporting API version "4" for the "Bugsnag Node" notifier
   And the event "unhandled" is true
   And the event "severity" equals "error"
-  And the event "severityReason.type" equals "unhandledErrorMiddleware"
+  And the event "severityReason.type" equals "unhandledException"
```
imjoehaines, line 36: These `unhandledErrorMiddleware` &rarr; `unhandledException` look like a regression to me, since it affects what's shown in the dashboard:

Screenshot 2023-02-21 at 10 02 20

I _think_ adding an `onError` in the server plugins that sets the handled state for unhandled events would fix this and be safe?

test/node/features/support/env.rb
```diff
@@ -17,3 +17,8 @@
   node_version = ENV['NODE_VERSION'].to_i
   skip_this_scenario("Skipping scenario on Node #{node_version}") if node_version < 16
 end
+
+Before('@skip_node_18') do |_scenario|
```
imjoehaines, line 21: Lets just future proof:

```ruby
(12..30).each do |version|
  Before("@skip_node_#{version}") do
    actual_version = ENV['NODE_VERSION'].to_i

    if actual_version == version
      skip_this_scenario(
        "Skipping scenario on Node #{actual_version}"
      )
    end
  end
end
```

imjoehaines: We can also get rid of `skip_before_node_6` & `skip_before_node_8`

imjoehaines: Or even:

```ruby
(12..30).each do |version|
  Before("@skip_node_#{version}") do
    actual_version = ENV['NODE_VERSION'].to_i

    if actual_version == version
      skip_this_scenario(
        "Skipping scenario on Node #{actual_version}"
      )
    end
  end

  Before("@skip_before_node_#{version}") do
    actual_version = ENV['NODE_VERSION'].to_i

    if actual_version < version
      skip_this_scenario(
        "Skipping scenario on Node #{actual_version}"
      )
    end
  end
end
```

packages/plugin-express/src/express.js
```diff
@@ -33,20 +29,11 @@ module.exports = {
 
       if (!client._config.autoDetectErrors) return next()
```
imjoehaines, line 30: This isn't correct anymore — we have to run `next` in the context otherwise APIs like `Bugsnag.addMetadata` will be global instead of request scoped

imjoehaines: In fact, isn't this check completely unnecessary now? Unhandled errors will bubble up to the global exception handler, which checks for this itself

packages/node/src/notifier.js
```diff
@@ -52,6 +59,8 @@ const Bugsnag = {
 
     const bugsnag = new Client(opts, schema, internalPlugins, { name, version, url })
 
+    bugsnag._clientContext = clientContext
```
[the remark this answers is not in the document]
djskinner: Seems better though I don't think it would have any practical difference

packages/plugin-express/src/express.js
```diff
@@ -33,20 +29,11 @@ module.exports = {
...
-          req.bugsnag._config.onUncaughtException(err, event, client._logger)
-        })
-        if (!res.headersSent) {
-          res.statusCode = 500
-          res.end('Internal server error')
-        }
-      })
-
-      return dom.run(next)
+      if (client._clientContext) {
```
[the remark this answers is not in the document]
djskinner: Make sense. At this point it should be defined.

packages/plugin-express/src/express.js
```diff
@@ -29,24 +25,10 @@ module.exports = {
         const { metadata, request } = getRequestAndMetadataFromReq(req)
         event.request = { ...event.request, ...request }
         event.addMetadata('request', metadata)
+        event._handledState.severityReason = { type: 'unhandledErrorMiddleware' }
```
imjoehaines, line 28: I don't think this is quite right — if it's a handled error then it shouldn't be set to `unhandledErrorMiddleware` and we need the `attributes` set so the dashboard knows the error came from Express

There's a `handledState` variable already, so I think this needs to be:

```js
if (event.unhandled) {
  event.severity = 'error'
  event._handledState = handledState
}
```

djskinner: Hmm, forcing the severityReason here has opened up an issue for unhandled promise rejections. In node versions prior to 16 the AsyncLocalStorage context is not available in the unhandled rejection handler so you get an error with severityReason `unhandledPromiseRejection` and none of the additional context (e.g. request). However, for  node >= 16 the context is available so this new logic runs and sets `unhandledErrorMiddleware`. At this point I'm not sure which is more correct. Furthermore, presumably, we'd like this to not be dependent on the node version i.e. it's either always `unhandledPromiseRejection` or always `unhandledErrorMiddleware`.

djskinner: Right, I've addressed this so the behaviour is the same as before. But it did then raise the issue of errors coming from the unhandled promise rejection handler, which are of the following handled state:

```
{
  severity: 'error',
  unhandled: true,
  severityReason: {
    type: 'unhandledPromiseRejection',
  }
}
```

I think it's useful that we keep these as unhandled promise rejections as this type of error can be lacking in detail so knowing it was an unhandled promise rejection is useful information. So the total logic is now:

```
        if (event.unhandled) {
          const originalSeverityReasonType = event._handledState.severityReason.type        
          event.severity = 'error'
          event._handledState = handledState
          if (originalSeverityReasonType === 'unhandledPromiseRejection') {
            event._handledState.severityReason.type = 'unhandledPromiseRejection'
          }
        }
```

Which results in:

```
{
  severity: 'error',
  unhandled: true,
  severityReason: {
    type: 'unhandledPromiseRejection',
    attributes: { framework: 'Express/Connect' }
  }
}
```

Note that the framework attribute remains.

In the dashboard this type of error currently looks like:

Screenshot 2023-02-23 at 14 11 17

With product steer I could make a change in the dashboard to also include the framework attribute, so that we surface the same data we do for an uncaught exception:

Screenshot 2023-02-23 at 14 12 44

packages/plugin-express/src/express.js
```diff
@@ -29,24 +25,13 @@ module.exports = {
...
-          res.end('Internal server error')
+        if (event._handledState.severityReason.type === 'unhandledException') {
+          event.severity = 'error'
+          event._handledState = handledState
         }
-      })
+      }, true)
 
-      return dom.run(next)
+      client._clientContext.run(requestClient, () => next())
```
imjoehaines, line 34:
```suggestion
      client._clientContext.run(requestClient, next)
```

packages/plugin-restify/src/restify.js
```diff
@@ -28,28 +25,12 @@ module.exports = {
...
-          })
-          res.end(body)
+        if (event._handledState.severityReason.type === 'unhandledException') {
+          event._handledState = handledState
         }
-      })
+      }, true)
 
-      return dom.run(next)
+      client._clientContext.run(requestClient, () => next())
```
imjoehaines, line 33:
```suggestion
      client._clientContext.run(requestClient, next)
```

imjoehaines (approved):

PR #1900, opened 2023-01-11
capture uncaught exceptions in aws-lambda

djskinner:
## Goal

AWS add a default unhandledRejection listener that forcefully exits the process. To workaround this the aws-lambda plugin removes all the listeners and only runs them after our own unhandled rejection handler has run.

It turns out the same thing is happening with uncaught exceptions but we weren't handling that case. This PR adds support for uncaught exceptions in the same manner as unhandled rejections.

To be clear the aws-lambda plugin previously was not reporting any errors that were caught by node's global uncaught exception handler. This PR fixes that.

## Design

- Same approach as used for unhandled rejections

## Testing

- e2e test was added covering this case
- Verified locally using aws sam


packages/plugin-node-uncaught-exception/uncaught-exception.js
```diff
@@ -12,12 +12,15 @@ module.exports = {
...
+      return new Promise(resolve => {
+        c._notify(event, () => {}, (e, event) => {
+          if (e) c._logger.error('Failed to send event to Bugsnag')
+          c._config.onUncaughtException(err, event, c._logger)
+          resolve()
+        })
       })
     }
-    process.on('uncaughtException', _handler)
+    process.prependListener('uncaughtException', _handler)
```
djskinner, line 23: This is a breaking change for the node notifier so we might want to avoid this change by creating a separate uncaught-exception plugin just for aws lambda. Thoughts?

imjoehaines: I'm not sure why it'd be breaking; do you just mean that `prependListener` isn't available on old Node versions?

In the unhandled rejection plugin we have a guard for this:

https://github.com/bugsnag/bugsnag-js/blob/aced3db494a4a779cb2911dbfbdd9e6c529c807c/packages/plugin-node-unhandled-rejection/unhandled-rejection.js#L21-L26

djskinner: I was thinking if a user's application code or other third-party library also adds listeners that this change could alter the order and break some implicit assumptions

imjoehaines: I'd say that's pretty unlikely and as we've already made this change in the unhandled rejection plugin, it'd be OK to do here as well

We could postpone this change until a v8.0.0 though, since it should be very rare — IIUC this is only really relevant if you have top-level code that throws? I guess also using regular callbacks, but doing that without waiting for the result is likely to cause other issues in a Lambda function

test/aws-lambda/features/unhandled.feature
```diff
@@ -91,13 +117,12 @@ Scenario Outline: unhandled exceptions are reported when using serverless-expres
 Scenario: unhandled asynchronous exceptions are reported when using serverless-express
     Given I setup the environment
     When I invoke the "ExpressFunction" lambda in "features/fixtures/serverless-express-app" with the "events/unhandled-async.json" event
-    Then the lambda response "errorMessage" equals "busted"
```
djskinner, line 94: The domain error handler was (incorrectly) preventing the application from crashing and returning a response. Now, the exception goes to the top-level uncaught exception handler before the process exists as it should

packages/plugin-aws-lambda/src/index.js
```diff
@@ -38,6 +38,20 @@ const BugsnagPluginAwsLambda = {
...
+    if (client._config.autoDetectErrors && client._config.enabledErrorTypes.unhandledExceptions) {
+      const listeners = process.listeners('uncaughtException')
+      process.removeAllListeners('uncaughtException')
+
+      // This relies on our unhandled rejection plugin adding its listener first
+      // using process.prependListener, so we can call it first instead of AWS'
+      process.on('uncaughtException', async (reason, promise) => {
+        for (const listener of listeners) {
+          await listener.call(process, reason, promise)
+        }
```
imjoehaines, line 51:
```suggestion
      process.on('uncaughtException', async (err) => {
        for (const listener of listeners) {
          await listener.call(process, err)
        }
```

imjoehaines (approved):

PR #1903, opened 2023-01-17
add request duration metadata to network breadcrumbs

yousif-bugsnag:
## Goal

Adds a new metadata field `duration` to network breadcrumbs for JS and Electron - this is the elapsed time in milliseconds between the request start and completion or error

## Design

- `plugin-network-breadcrumbs`
    - For `XMLHttpRequest` : 
        - Override request object's `send` method in order to record the start time of the request, and get the duration in milliseconds when `handleXHRLoad` / `handleXHRError` are invoked
    - For `fetch`:
    	- Record start of request when calling native `fetch` and get the duration in milliseconds when`handleFetchSuccess` / `handleFetchError` are invoked

- `plugin-electron-net-breadcrumbs`
	-  Override request object's `request.end` method in order to record the start time of the request, and get the duration in milliseconds when the `request` / `abort` / `error` handlers are invoked

## Testing

- Unit tests:
	- Extended existing unit tests to check for new metadata field

- e2e tests
	- Extended existing e2e tests to check for new metadata field
	- Added basic e2e tests for network breadcrumbs on Electron

### DIFF
--- a/packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
+++ b/packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
@@ -32,12 +32,20 @@ module.exports = net => ({
         return request
       }
 
+      const originalEnd = request.end
+      let requestStart
+
+      request.end = (...args) => {
+        requestStart = new Date()
+        originalEnd.apply(request, args)
+      }
+
       request.on('response', response => {
         const success = response.statusCode < 400
 
         client.leaveBreadcrumb(
           `net.request ${success ? 'succeeded' : 'failed'}`,
-          { request: `${method} ${url}`, status: response.statusCode },
+          { request: `${method} ${url}`, status: response.statusCode, duration: getDuration(requestStart) },
           BREADCRUMB_REQUEST
         )
       })
@@ -45,7 +53,7 @@ module.exports = net => ({
       request.on('abort', () => {
         client.leaveBreadcrumb(
           'net.request aborted',
-          { request: `${method} ${url}` },
+          { request: `${method} ${url}`, duration: getDuration(requestStart) },
           BREADCRUMB_REQUEST
         )
       })
@@ -53,7 +61,7 @@ module.exports = net => ({
       request.on('error', (error) => {
         client.leaveBreadcrumb(
           'net.request error',
-          { request: `${method} ${url}`, error: error.message },
+          { request: `${method} ${url}`, error: error.message, duration: getDuration(requestStart) },
           BREADCRUMB_REQUEST
         )
       })
@@ -62,3 +70,5 @@ module.exports = net => ({
     }
   }
 })
+
+const getDuration = (startTime) => startTime && new Date() - startTime
--- a/packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
+++ b/packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
@@ -32,11 +32,19 @@ module.exports = net => ({
         return request
       }
 
-      const originalEnd = request.end
       let requestStart
 
+      // For chunked requests the request begins on the first write operation,
+      // otherwise the request begins when the request is finalised
+      const originalWrite = request.write
+      request.write = (...args) => {
+        if (request.chunkedEncoding && !requestStart) requestStart = new Date()
+        originalWrite.apply(request, args)
+      }
+
+      const originalEnd = request.end
       request.end = (...args) => {
-        requestStart = new Date()
+        if (!requestStart) requestStart = new Date()
         originalEnd.apply(request, args)
       }
 
--- a/test/electron/fixtures/app/main.js
+++ b/test/electron/fixtures/app/main.js
@@ -1,4 +1,5 @@
 const { join } = require('path')
+const { net } = require('electron')
 const {
   uncaughtException,
   unhandledRejection,
@@ -41,6 +42,14 @@ function createWindow () {
 
   // eslint-disable-next-line no-undef
   win.loadFile(join(__dirname, htmlRelativePath))
+}
+
+function makeSimpleGetRequest (fail = false) {
+  const url = fail ? 'https://non.existent.url/' : 'https://google.com/'
+  const request = net.request(url)
+  request.on('response', notify)
+  request.on('error', notify)
+  request.end()
 }
 
 app.whenReady().then(createWindow)
@@ -92,4 +101,8 @@ ipcMain.on('main-process-clear-feature-flags', () => {
 
 ipcMain.on('main-process-clear-feature-flags-now', () => {
   Bugsnag.clearFeatureFlags()
+})
+
+ipcMain.on('main-process-get-request', (_event, fail) => {
+  makeSimpleGetRequest(fail)
 })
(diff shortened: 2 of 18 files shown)


packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
```diff
@@ -62,3 +70,5 @@ module.exports = net => ({
     }
   }
 })
+
+const getDuration = (startTime) => startTime && new Date() - startTime
```
imjoehaines, line 74: I think this needs to be stricter if we expect `startTime` to be missing — right now it can return `false`, which will end up in metadata and visible in the dashboard

If `startTime` is missing, it should probably return `undefined` so that it gets omitted from the JSON payload entirely

```suggestion
const getDuration = (startTime) => startTime ? new Date() - startTime : undefined
```

test/electron/fixtures/app/main.js
```diff
@@ -43,6 +44,14 @@ function createWindow () {
   win.loadFile(join(__dirname, htmlRelativePath))
 }
 
+function makeSimpleGetRequest (fail = false) {
+  const url = fail ? 'https://non.existent.url/' : 'https://google.com/'
```
imjoehaines, line 48: I think we should use URLs we control if we're making requests for real, e.g. something like:

```suggestion
  const url = fail
    ? 'https://does.not.exist.bugsnag.com/'
    : 'https://bugsnag.com/'
```

packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
imjoehaines: This could also use a test

test/electron/fixtures/app/main.js
imjoehaines: Maybe we should be hitting a local server though? That way the network conditions on CI shouldn't cause issues

packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
yousif-bugsnag: Hmm this is already returning `undefined` when `startTime` is `undefined`. The `&&` operator just returns the value of the first operand if it's falsey - it doesn't coerce it to a boolean value.

Two of the existing tests are already testing for this (albeit implicitly) because they abort the request without calling `request.end`, meaning `startTime` is `undefined`: https://github.com/bugsnag/bugsnag-js/blob/PLAT-9443/network-breadcrumb-duration/packages/plugin-electron-net-breadcrumbs/test/net-breadcrumbs.test-main.ts#L118-L162

These tests would fail if the breadcrumb's `duration` was `false` - do you think it needs its own explicit test/check?

imjoehaines: Yeah you're right — not sure what I was thinking here! That test seems fine to me

test/electron/fixtures/app/main.js
yousif-bugsnag: Yeah I went with those urls because that's what the Android e2e tests do, but this is a much better idea - will also be super useful when we come to add tests for response content length etc.

I've updated the tests to use the mock server and also created separate scenarios for a failed request (4xx response from the server) and a request error (hitting an invalid port on localhost)

test/electron/fixtures/events/main/breadcrumbs/network/error.json
```diff
@@ -0,0 +1,84 @@
...
+        "type": "handledException"
+      },
+      "breadcrumbs": [
+        {
+          "type": "request",
+          "name": "net.request error",
+          "timestamp": "{TIMESTAMP}",
+          "metaData": {
+            "request" : "GET http://locahost:65536/",
+            "duration": "{TYPE:number}"
```
imjoehaines, line 64: Should this have a `status`? If not you can use `{TYPE:undefined}` to assert that it's missing

```suggestion
            "request": "GET http://locahost:65536/",
            "status": "{TYPE:undefined}",
            "duration": "{TYPE:number}"
```

imjoehaines (approved):

packages/plugin-electron-net-breadcrumbs/net-breadcrumbs.js
```diff
@@ -32,28 +32,36 @@ module.exports = net => ({
         return request
       }
 
+      const originalEnd = request.end
+      let requestStart
+
+      request.end = (...args) => {
+        requestStart = new Date()
+        originalEnd.apply(request, args)
+      }
```
gingerbenw, line 41: we save the `requestStart` date on `request.end`? if that's correct I think a comment here would be nice

yousif-bugsnag: Aah this is a good point actually - if the request is using chunked encoding that would only be the duration of the final chunk that was sent. Maybe it's better to set `requestStart` on the first write operation instead, or do something like this (cc @imjoehaines):
```suggestion
      let requestStart

      // For chunked requests the request begins when the first chunk is written,
      // otherwise the request begins when the last chunk is written
      const originalWrite = request.write
      request.write = (...args) => {
        if (request.chunkedEncoding && !requestStart) requestStart = new Date()          
        originalWrite.apply(request, args)
      }

      const originalEnd = request.end
      request.end = (...args) => {
        if (!requestStart) requestStart = new Date()
        originalEnd.apply(request, args)
      }
```

imjoehaines: `request.end` is what finalises the request, i.e. it makes sure that everything is actually sent. This is still a good point though — `request.write` _can_ start the process of sending the request:

> The first write operation ***may*** cause the request headers to be issued on the wire
> https://www.electronjs.org/docs/latest/api/client-request#requestwritechunk-encoding-callback

So only using `request.end` means the duration may be wrong if we want to capture the duration from the point the request actually starts being sent — I doubt it will make a different most of the time in a real app TBH, especially with `Date` only having millisecond precision, but if there's a large request body it could take a non-trivial amount of time from writing it to the request with `request.write` and finishing the request with `request.end`

Maybe we need to use both `request.write` & `request.end`, recording the start time from whichever is called first? You have to call `request.end` but don't necessarily have to call `request.write`, so we can't just use `request.write`

imjoehaines: Replied at the same time as you @yousif-bugsnag!

What you suggested LGTM if it's only when using chunked encoding that `write` can start the request. Electron's docs aren't that specific so I'm not sure if there are other cases where `write` can start sending the request?

yousif-bugsnag: Cool, yep, looks that way - I've been doing some manual testing with a local http server and that tallies up with what I've observed (the connection is established on `write` only when `request.chunkedEncoding = true`)

Although looks like I could have just read the source! 😅

yousif-bugsnag:
I've added some unit tests for the new duration logic but was finding that the unit tests for `plugin-electron-net-breadcrumbs` started flaking heavily again so I've reverted the previous change to reinstate these and updated the PR comment accordingly

imjoehaines (approved):

PR #1904, opened 2023-01-17
add autoTrackSessions to web-worker notifier and disable as default

gingerbenw:
## Goal

Add the ability to start and auto track sessions from the `web-worker` notifier

## Testing

- Unit tests for `Bugsnag.startSession()`
- Unit tests for `autoTrackSession` config option
- End to end test for `autoTrackSession` config option

### DIFF
--- a/packages/web-worker/src/notifier.js
+++ b/packages/web-worker/src/notifier.js
@@ -38,7 +38,9 @@ export const Bugsnag = {
 
     bugsnag._logger.debug('Loaded!')
 
-    return bugsnag
+    return bugsnag._config.autoTrackSessions
+      ? bugsnag.startSession()
+      : bugsnag
   },
   start: (opts) => {
     if (Bugsnag._client) {
--- a/packages/web-worker/src/notifier.js
+++ b/packages/web-worker/src/notifier.js
@@ -8,6 +8,7 @@ import pluginWindowOnError from '@bugsnag/plugin-window-onerror'
 import pluginWindowUnhandledRejection from '@bugsnag/plugin-window-unhandled-rejection'
 import config from './config'
 import pluginBrowserDevice from '@bugsnag/plugin-browser-device'
+import pluginBrowserSession from '@bugsnag/plugin-browser-session'
 import pluginPreventDiscard from './prevent-discard'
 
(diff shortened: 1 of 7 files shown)


packages/web-worker/src/notifier.js
```diff
@@ -31,7 +31,9 @@ export const Bugsnag = {
 
     bugsnag._logger.debug('Loaded!')
 
-    return bugsnag
+    return bugsnag._config.autoTrackSessions
+      ? bugsnag.startSession()
+      : bugsnag
```
imjoehaines, line 36: I don't think this works since the `@bugsnag/plugin-browser-session` isn't installed for `@bugsnag/web-worker` — there's no `_sessionDelegate` on `bugsnag` here, so it'll crash in core

This could use some unit & end to end tests 😀

imjoehaines (changes requested):

packages/web-worker/test/notifier.test.ts
```diff
@@ -82,6 +82,43 @@ describe('worker notifier', () => {
     Bugsnag.start(testConfig)
     expect(Bugsnag.isStarted()).toBe(true)
   })
+
+  describe('session management', () => {
+    it('successfully starts a session', (done) => {
+      const Bugsnag = getBugsnag()
+      Bugsnag.start(API_KEY)
+      Bugsnag.startSession()
```
imjoehaines, line 90: It might be worth checking that `startSession` is what actually caused the request, e.g.:

```suggestion
      Bugsnag.start(API_KEY)

      expect(typedGlobal.fetch).not.toHaveBeenCalled()

      Bugsnag.startSession()
```

test/browser/features/web_worker.feature
```diff
@@ -21,3 +21,9 @@ Scenario: setting collectUserIp option to false
   Then the error is a valid browser payload for the error reporting API
   And the event "request.clientIp" equals "[REDACTED]"
   And the event "user.id" equals "[REDACTED]"
+
+Scenario: setting autoTrackSessions option to true
```
imjoehaines, line 25: Similar to the above — maybe worth adding `And I should receive no sessions` to the other scenarios?

PR #1909, opened 2023-01-23
Update NextJS example with correct options for IgnorePlugin

camro:
Adds in an explicit key for NextJS example for webpack IgnorePlugin.

Was throwing errors without, and matches the [official documentation](https://webpack.js.org/plugins/ignore-plugin/)


luke-belton:
Hi @camro - thanks for bringing this to our attention, we'll look to get this change merged 👍

PR #1914, opened 2023-01-25
PLAT-9410 Modularise browser upload script

gingerbenw:
## Goal

To make the `cdn-upload` script more agnostic, and support multiple packages uploading to the CDN

## Changeset

- Refactored upload script and moved to root `bin` directory
- Updated `cdn-upload` scripts in relevant packages
- Updated tests
- Removed unnecessary packages   

## Testing

- Manually uploaded build artefacts from the `web-worker` and `browser` packages using the new script in a sandbox environment
- Updated e2e tests to use `importScripts` syntax, improving test coverage


bin/cdn-upload
```diff
@@ -0,0 +1,94 @@
...
+  if (response.statusCode >= 400) {
+    throw new Error(`Failed to upload ${relativePath}: ${response.statusCode} ${response.statusMessage}`)
+  }
+
+  return response
+}
+
+const invalidate = (paths) => {
+  console.log(`invalidating CloudFront cache for the following objects:\n${paths.join('\n')}`)
+  return promisify(cfClient.createInvalidation.bind(cfClient))(invalidationKey, `bugsnag-js@${version}`, paths) // TODO: Do we need to get the package name? i.e. bugsnag-js
```
gingerbenw, line 55: Does this invalidation need to be different for the web-worker package?

imjoehaines: Yeah, I think they need to be different

The "`invalidationKey`" (called "`distributionId`" in AWS docs) will be the same, but the "[CallerReference](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CreateInvalidation.html#API_CreateInvalidation_RequestSyntax)" [should be different](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CreateInvalidation.html#API_CreateInvalidation_RequestSyntax):

> CloudFront uses the value to prevent you from accidentally resubmitting an identical request. Whenever you create a new invalidation request, you must specify a new value for CallerReference and change other values in the request as applicable
> &hellip;
> If CallerReference is a value you already sent in a previous invalidation batch request but the content of any Path is different from the original request, CloudFront returns an InvalidationBatchAlreadyExists error.

It's completely arbitrary though, so using the package name + version should be fine

```suggestion
  return promisify(cfClient.createInvalidation.bind(cfClient))(invalidationKey, `${packageName}@${packageVersion}`, paths)
```

bin/cdn-upload
```diff
@@ -0,0 +1,100 @@
...
+const S3 = require('aws-sdk/clients/s3')
+const Cloudfront = require('aws-sdk/clients/cloudfront')
+
+// INPUT PARAMETERS
+const packageDir = process.cwd()
+const packageJson = require(packageDir + "/package.json")
+const packageName = packageJson.name
+const packageVersion = packageJson.version
+const bucketName = 'bugsnagcdn'
+const distributionId = 'E205JDPNKONLN7'
```
gingerbenw, line 16: Could these be added using environment variables?

bin/cdn-upload
```diff
@@ -0,0 +1,100 @@
...
+  
+  const fileStream = await fs.createReadStream(localPath)
+  const response = await s3Client.putObject({ 
+    Bucket: bucketName, 
+    Key: remotePath, 
+    Body: fileStream,
+    ACL: 'public-read',
+    CacheControl: 'public, max-age=315360000'
+    // TODO: Is it possible to add the Access-Control-Allow-Origin header?
+    //   'Access-Control-Allow-Origin': '*',
```
gingerbenw, line 36: This header was previously set on the old upload script along with the acl and cache control, however as we're using cloudfront I'm not convinced this is required

bin/cdn-upload
imjoehaines: They're static values for each package, so I think we can set them automatically based on the package name

It makes the script not _entirely_ generic, but I think it's better to be safe against accidentally uploading something to the wrong bucket or invalidating the wrong cache

So we could do something like this:

```js
const packageConfigs = new Map([
  ['@bugsnag/browser', { distributionId: '...', bucketName: '...' }],
  ['@bugsnag/web-worker', { distributionId: '...', bucketName: '...' }],
])

const config = packageConfigs.get(packageName)

if (!config) {
  throw new Error(`Unknown package name: ${packageName}`)
}

const distributionId = config.distributionId
const bucketName = config.bucketName
```

bin/cdn-upload
imjoehaines: Yeah, there's no public access to this bucket so I _think_ it's unnecessary

bin/cdn-upload
```diff
@@ -0,0 +1,100 @@
...
+    // if this is a release (as opposed to a prerelease), update the major/minor aliases
+    const major = `v${semver.major(packageVersion)}`
+    
+    for (const file of files) {
+      const localPath = getLocalPath(file)
+      const uploadPath = `${major}/${path.basename(file)}`
+      await upload(localPath, track(uploadPath))
+    }
+
+    const minor = `v${semver.major(packageVersion)}.${semver.minor(packageVersion)}`
```
imjoehaines, line 88:
```suggestion
    const minor = `v${major}.${semver.minor(packageVersion)}`
```

imjoehaines (approved):

PR #1924, opened 2023-02-10
rewrite plugin-contextualize without domain

djskinner:
## Goal

- rewrite plugin-contextualize to not use the deprecated node domain API
- use a similar approach and the same AsyncLocalStorage instance as in https://github.com/bugsnag/bugsnag-js/pull/1898

## Changeset

- rewrite plugin-contextualize to not use the deprecated node domain API.
  - The usage API remains the same (`contextualize: (fn: () => void, onError: OnErrorCallback) => void`) but there are changes in behaviour.
  - Before, the plugin would capture unhandled errors regardless of the config values `autoDetectErrors` and `enabledErrorTypes.unhandledExceptions`. Now, the plugin will only capture unhandled exceptions if both `autoDetectErrors` and `enabledErrorTypes.unhandledExceptions` are enabled.
  - Before, the plugin was unable to add context to unhandled promise rejections. Now, (since node 16) unhandled promise rejections are captured with context if `autoDetectErrors` and `enabledErrorTypes.unhandledRejections` are enabled.

## Testing

- e2e coverage covering both uncaught exceptions and unhandled rejections
- no unit test coverage - it appears impossible to cover this in a meaningful way with jest


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Changed
 
 - (plugin-navigation-breadcrumbs) calling `pushState` or `replaceState` no longer triggers a new session when `autoTrackSessions` is enabled [#1820](https://github.com/bugsnag/bugsnag-js/pull/1820)
+- (plugin-contextualize) reimplement without relying on the deprecated node Domain API [#1924](https://github.com/bugsnag/bugsnag-js/pull/1924)
```
imjoehaines, line 12: Maybe also worth noting it now supports promise rejections on Node 16+?

packages/plugin-contextualize/contextualize.js
```diff
@@ -9,21 +8,20 @@ module.exports = {
...
+      const clonedClient = clone(client)
+
+      // add the stacktrace to the cloned client so it can be used later
+      // by the uncaught exception handler. Note the unhandled rejection
+      // handler does not need this because it gets a stacktrace
+      clonedClient.fallbackStack = fallbackStack
+
+      clonedClient.addOnError(onError)
+
+      if (client._clientContext) {
```
imjoehaines, line 20: As in #1898, falling back to just calling `fn` isn't right because it means the `onError` added here won't ever be called, so the plugin silently does nothing

test/aws-lambda/features/promise-rejection.feature
```diff
@@ -42,7 +42,7 @@ Scenario Outline: unhandled promise rejections are not reported when autoDetectE
     When I invoke the "<lambda>" lambda in "features/fixtures/simple-app" with the "events/<type>/promise-rejection.json" event
     Then the lambda response "errorMessage" equals "Error: yikes"
     And the lambda response "errorType" equals "Runtime.UnhandledPromiseRejection"
-    And the lambda response "trace" is an array with 6 elements
+    And the lambda response "trace" is an array with 5 elements
```
djskinner, line 45: The change here is that the line `"    at process.EventEmitter.emit (domain.js:483:12)"` is no longer present. This frame was being added due to the mere presence of `require('domain')` in the contextualize plugin, which is part of the node notifier and therefore also instantiated in aws lambda.

imjoehaines (approved):

PR #1925, opened 2023-02-13
add requestContentLength breadcrumb metadata for xhr and fetch requests

yousif-bugsnag:
## Goal

Adds `requestContentLength` metadata to network breadcrumbs for XHR and fetch requests. This is the length of the request body in bytes. 

## Design

If a body is passed to `XMLHTTPRequest.send()` or `fetch()` the length in bytes of the body is calculated and recorded prior to calling into the respective native methods.

The content length is not calculated if the request body is a `ReadableStream` as these cannot be read without interfering with the successful completion of request.

The content length is also omitted if the request body is `FormData` or `Document` object as it's difficult to calculate accurately for these object types. 

Where possible the byte length of the request body is read directly, (e.g. for `Blob`, `TypedArray`, `ArrayBuffer` and `DataView` objects). Otherwise the request body is serialized and a `Blob` is used to get the byte length of the UTF-8 encoded body.

## Testing

Manually tested that the calculated content length matches the `Content-Length` header for the various request types.

Since the unit tests run in Node and not in the browser, this is covered by e2e tests. New e2e tests have been added to check that the content length is calculated for the various request types.

### DIFF
--- a/packages/plugin-network-breadcrumbs/network-breadcrumbs.js
+++ b/packages/plugin-network-breadcrumbs/network-breadcrumbs.js
@@ -206,13 +206,13 @@ module.exports = (_ignoredUrls = [], win = window) => {
 
   const getByteLength = (body) => {
     // body could be any of the types listed here: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send#parameters
-    // Omit content length metadata if body is an unsupported type
+    // ReadableStreams cannot be read safely and it's difficult to get an accurate byte length for FormData and Document inputs
     if ((body === null || typeof body === 'undefined') ||
       (win.ReadableStream && body instanceof win.ReadableStream) ||
       (win.FormData && body instanceof win.FormData) ||
       (win.Document && body instanceof win.Document)) return undefined
 
-    // See if we can get the byte length directly
+    // Try to read the byte length directly
     if (typeof body.byteLength === 'number') {
       // ArrayBuffer, DataView, TypedArray
       return body.byteLength
@@ -222,10 +222,11 @@ module.exports = (_ignoredUrls = [], win = window) => {
       return undefined
     }
 
-    // Stringify the input and construct a Blob to get the utf-8 encoded byte length
-    // This may fail if the input object has no prototype or a broken toString
     try {
+      // do a simple stringification - this may fail if the input object has no prototype or a broken toString
       const stringified = String(body)
+
+      // use a Blob to get the utf-8 encoded byte length
       return new win.Blob([stringified]).size
     } catch (e) {
       return undefined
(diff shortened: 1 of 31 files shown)


packages/plugin-network-breadcrumbs/network-breadcrumbs.js
```diff
@@ -190,6 +204,35 @@ module.exports = (_ignoredUrls = [], win = window) => {
     }
   }
 
+  const getByteLength = (body) => {
+    // body could be any of the types listed here: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send#parameters
+    // ReadableStreams cannot be read safely and it's difficult to get an accurate byte length for FormData and Document inputs
+    if ((body === null || typeof body === 'undefined') ||
+      (win.ReadableStream && body instanceof win.ReadableStream) ||
+      (win.FormData && body instanceof win.FormData) ||
+      (win.Document && body instanceof win.Document)) return undefined
```
gingerbenw, line 213: Should we test this is an instance of the class rather than an instance of itself?
```suggestion
      (win.ReadableStream && body instanceof ReadableStream) ||
      (win.FormData && body instanceof FormData) ||
      (win.Document && body instanceof Document)) return undefined
```

packages/plugin-network-breadcrumbs/network-breadcrumbs.js
```diff
@@ -190,6 +204,35 @@ module.exports = (_ignoredUrls = [], win = window) => {
...
+
+    // Try to read the byte length directly
+    if (typeof body.byteLength === 'number') {
+      // ArrayBuffer, DataView, TypedArray
+      return body.byteLength
+    } else if (win.Blob && body instanceof win.Blob) {
```
gingerbenw, line 219:
```suggestion
    } else if (win.Blob && body instanceof Blob) {
```

packages/plugin-network-breadcrumbs/test/network-breadcrumbs.test.ts
```diff
@@ -501,7 +514,7 @@ describe('plugin: network breadcrumbs', () => {
 
     const request = new XMLHttpRequest()
     request.open('GET', '/')
-    request.send(false, 200)
+    request.send(undefined, false, 200)
```
gingerbenw, line 517: why do we need to pass undefined here?

gingerbenw (commented):
Forgive me, but what does `reflect` do?

packages/plugin-network-breadcrumbs/network-breadcrumbs.js
yousif-bugsnag: This wouldn't pass the linting checks as those types are not globally defined (but they are exposed on `window`) - discussed here: https://bugsnag.slack.com/archives/G03JXULFW06/p1674737016675659

packages/plugin-network-breadcrumbs/test/network-breadcrumbs.test.ts
yousif-bugsnag: That's the request body - the mock `XMLHttpRequest` class had some existing optional parameters to set the desired failure state and status code but the real `XMLHTTPRequest.send` method just accepts a single (optional) `body` parameter - without this the plugin would interpret the request body as `false` and try to calculate the content length for that.

gingerbenw (approved):

PR #1927, opened 2023-02-16
enable breadcrumbs and context-scoped calls for node

djskinner:
## Goal

- Enable full support for breadcrumbs on node.js

## Design

- Remove the noop method override on `leaveBreadcrumb` that prevented usage of breadcrumbs on the main client instance. Note cloned instances such as those inside a request handler didn't have this limitation though such usage was undocumented.
- When calling method on the main client instance, such as `Bugsnag.leaveBreadrumb` or `Bugsnag.addFeatureFlag` from within an AsyncLocalStorage context (e.g. when using the express plugin or plugin-contextualize) forward the method onto the cloned client instance thereby achieving scoped context without having to pass the cloned instance of the client (e.g. `req.bugsnag`) around.
- Add a `runInContext` middleware for express that can be used to re-establish the AsyncLocalStorage context which can be lost in some scenarios (where such context loss prevents the correct operation of `Bugsnag.` calls, such as `Bugsnag.leaveBreadcrumb`)

## Testing

- e2e test coverage
- manual testing


imjoehaines (commented):
Changes look good, but I think we need to test the APIs we're changing here — adding some unit tests to [`notifier.test.ts`](https://github.com/bugsnag/bugsnag-js/blob/next/packages/node/test/notifier.test.ts) for `Bugsnag.addMetadata` & friends should be fine

packages/plugin-express/src/express.js
```diff
@@ -55,7 +55,11 @@ module.exports = {
       next(err)
     }
 
-    return { requestHandler, errorHandler }
+    const runInContext = (req, res, next) => {
+      client._clientContext.run(req.bugsnag, next)
+    }
```
djskinner, line 60: This is a new express middleware that can be used to restore async local storage context if it is lost. See E2E tes showing usage

packages/node/src/notifier.js
```diff
@@ -63,12 +60,6 @@ const Bugsnag = {
 
     bugsnag._logger.debug('Loaded!')
 
-    bugsnag.leaveBreadcrumb = function () {
-      bugsnag._logger.warn('Breadcrumbs are not supported in Node.js yet')
-    }
-
-    bugsnag._config.enabledBreadcrumbTypes = []
```
djskinner, line 70: remove the patches that were preventing breadcrumbs from being used with the node notifier

test/node/features/fixtures/express/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "bugsnag-test",
   "dependencies": {
-    "body-parser": "^1.19.0",
+    "body-parser": "^1.20.2",
```
djskinner, line 4: Bumping this to fix the context-loss present in the older versions

test/node/features/fixtures/express/scenarios/app.js
```diff
@@ -144,11 +144,11 @@ app.post('/features/unhandled', bodyParser.urlencoded(), function (req, res, nex
   // the request body is an object of feature flag name -> variant
   const featureFlags = Object.keys(req.body).map(name => ({ name, variant: req.body[name] }))
 
-  req.bugsnag.addFeatureFlags(featureFlags)
-  req.bugsnag.clearFeatureFlag('from config 3')
+  Bugsnag.addFeatureFlags(featureFlags)
+  Bugsnag.clearFeatureFlag('from config 3')
```
djskinner, line 148: This was failing due to the context loss from doing `bodyParser.urlencoded()` with the older version of the dependency. Bumping fixes it. Could also have used `middleware.runInContext` to restore it

test/node/features/unhandled_errors.feature
```diff
@@ -55,50 +55,6 @@ Scenario: not reporting unhandledRejections when autoDetectErrors is off
   And I wait for 1 second
   Then I should receive no requests
 
-Scenario: using contextualize to add context to an error
```
djskinner, line 58: moved to the new contextualize suite

imjoehaines (approved):

PR #1934, opened 2023-03-07
Update UPGRADING.md for v8

djskinner:


UPGRADING.md
```diff
@@ -1,6 +1,51 @@
...
+
+### node
+
+This version contains major improvements to the node notifier, including making the top-level `Bugsnag` static interface context-aware and enabling breadcrumbs for node projects.
+
+The minimum supported node version is now v12.17.0.
+
+#### context-aware `Bugsnag` calls
+
+When using `plugin-express`, `plugin-koa`, `plugin-restify`, or `plugin-contextualize` a clone of the top-level Bugsnag client is made so that any subsequent changes made to the client (such as attaching metadata) only affect the scope of a particular context (a request in the case of the web server plugins, and a function call in the case of `plugin-contextualize`).
```
gareththackeray, line 14: My feeling is that the "request context" is the overwhelming majority use case.  So might be better to rephrase as
```suggestion
When using `plugin-express`, `plugin-koa`, `plugin-restify`, or `plugin-contextualize` a clone of the top-level Bugsnag client is made so that any subsequent changes made to the client (such as attaching metadata) only affect the scope of the current web request. (Or a function call when using `plugin-contextualize`.)
```

UPGRADING.md
```diff
@@ -1,6 +1,51 @@
...
+
+Prior to v8, calls made to the top-level `Bugsnag` static interface were not aware of this context so users had to ensure they were calling methods on the correct client instance, i.e. the cloned client that was made available on `req.bugsnag` (or `ctx.bugsnag` for koa). This wasn't ideal because if you wanted to interact with Bugsnag in some function deep in a call stack you would have to pass `req.bugsnag` all the way down, as calling `Bugsnag.notify` would not have contained the request metadata gathered by the plugin. With version 8 of the notifier top-level calls to `Bugsnag` are now context-aware. This means you can call `Bugsnag.notify` (or `Bugsnag.leaveBreadcrumb` etc.), and, if it was called within a context, the call will be forwarded to the correct cloned version of that client (i.e. for the particular request from which the call originated).
```
gareththackeray, line 16: Am I silly to think that "v8" could be confused for the JS engine?  We could say `bugsnag-js` v8.

UPGRADING.md
```diff
@@ -1,6 +1,51 @@
...
+    await next()
+  }
+})
+```
+
+Note: `req.bugsnag` (and `ctx.bugsnag` in koa) is still present in v8.
+
+#### breadcrumb support
+
+Breadcrumb support has been enabled for node. This means you can call `Bugsnag.leaveBreacrumb` to attach short log statements to each error report to help diagnose what events led to the error.
```
gareththackeray, line 45:
```suggestion
Breadcrumb support has been enabled for node. This means you can call `Bugsnag.leaveBreadcrumb` to attach short log statements to each error report to help diagnose what events led to the error.  
```

UPGRADING.md
djskinner: I agree the clarification is worth it

UPGRADING.md
```diff
@@ -1,6 +1,51 @@
...
+-    ctx.bugsnag.notify(new Error('handled'))
++    Bugsnag.notify(new Error('handled'))
+    await next()
+  } else {
+    await next()
+  }
+})
+```
+
+Note: `req.bugsnag` (and `ctx.bugsnag` in koa) is still present in version 8 of `bugsnag-js`.
```
djskinner, line 41: I wonder if we want to say something:
- "you can continue using req.bugsnag as before"
- there are situations where context is lost and you may need to use `req.bugsnag`. See the context-loss section of the docs

UPGRADING.md
```diff
@@ -1,6 +1,51 @@
...
+
+#### breadcrumb support
+
+Breadcrumb support has been enabled for node. This means you can call `Bugsnag.leaveBreadcrumb` to attach short log statements to each error report to help diagnose what events led to the error.  
+
+Currently no breadcrumbs are automatically collected in node.
+
 ## `bugsnag-react-native@*` to `@bugsnag/react-native@7.3`
```
tomlongridge, line 49: We should add a note about `contextualise` now only being fired if configured to do so with the autoDetectErrors and enabledErrorTypes configuration options.

UPGRADING.md
```diff
@@ -9,6 +54,10 @@ The previous Bugsnag React Native SDK – [`bugsnag-react-native`](https://githu
 
 See the [React Native upgrade guide](/packages/react-native/UPGRADING.md) for specific instructions on how to upgrade from `bugsnag-react-native` to `@bugsnag/react-native`.
 
+#### plugin-contextualize
+
+Unhandled errors that occur within a contextualize context now respect the `autoDetectErrors`` and `enabledErrorTypes`` configuration options. Previously unhandled errors would have been caught regardless of the configuration.
```
tomlongridge, line 59:
```suggestion
Unhandled errors that occur within a contextualize context now respect the `autoDetectErrors` and `enabledErrorTypes` configuration options. Previously unhandled errors would have been caught regardless of the configuration.
```

tomlongridge, line 59: Is the change also that we previously _handled_ the error, so the thread didn't crash?

UPGRADING.md
tomlongridge:
```suggestion
Notes

* `req.bugsnag` (and `ctx.bugsnag` in koa) is still present in version 8 of `bugsnag-js`, so you can continue using these as before.
* There are rare situations on Express servers when this contextual storage can get lost, causing the data stored to become server-scoped and so affect all threads that are being executed. See our [online docs](https://docs.bugsnag.com/platforms/javascript/express/node-async/#context-loss-in-express-servers) for full details.
```

UPGRADING.md
djskinner: That also applies to unhandled errors in requests so will need mentioning elsewhere as well

djskinner: I've discussed this in two places - what do you think to the wording?

tomlongridge (approved):

PR #1935, opened 2023-03-08
Use android.namespace to support projects using AGP 8

lemnik:
## Goal
Support for ReactNative projects using AGP, which moved the app namespace out of the `AndroidManifest.xml` and into `build.gradle`.

## Changelog
Added a conditional definition of the project `namespace` if the Android Gradle Plugin supports it. This allows us to remain backwards compatible with older versions of AGP.

## Testing
Tested manually with an AGP 8 compatible project, relied on existing tests for backwards compatibility testing.

### DIFF
--- a/packages/react-native/android/build.gradle
+++ b/packages/react-native/android/build.gradle
@@ -24,6 +24,10 @@ android {
     buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
 
+    if (android.hasProperty('namespace')) {
+        namespace 'com.bugsnag.reactnative'
+    }
+
     defaultConfig {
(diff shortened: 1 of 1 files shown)


packages/react-native/android/build.gradle
```diff
@@ -24,6 +24,10 @@ android {
     buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
 
+    if (android.hasProperty('namespace')) {
+        namespace 'com.bugsnag.reactnative'
```
tomlongridge, line 28: So there are no issues with it still being defined in `AndroidManifest.xml` as well?

lemnik: It emits a warning when defined in both, but doesn't fail the build.

tomlongridge (approved):

PR #1938, opened 2023-03-09
[PLAT-9731] Fix XHR delivery postReportCallback

gingerbenw:
## Goal

To ensure delivery errors are passed to the post report callback when using the `delivery-xml-http-request` plugin

## Design

Follow the same implementation as `delivery-x-domain-request`

## Testing

Added new unit test and update existing test to verify errors are handled correctly

### DIFF
--- a/packages/delivery-xml-http-request/delivery.js
+++ b/packages/delivery-xml-http-request/delivery.js
@@ -37,9 +37,20 @@ module.exports = (client, win = window) => ({
     try {
       const url = client._config.endpoints.sessions
       const req = new win.XMLHttpRequest()
+
       req.onreadystatechange = function () {
-        if (req.readyState === win.XMLHttpRequest.DONE) cb(null)
+        if (req.readyState === win.XMLHttpRequest.DONE) {
+          const status = req.status
+          if (status === 0 || status >= 400) {
+            const err = new Error(`Request failed with status ${status}`)
+            client._logger.error('Session failed to send…', err)
+            cb(err)
+          } else {
+            cb(null)
+          }
+        }
       }
+
       req.open('POST', url)
       req.setRequestHeader('Content-Type', 'application/json')
       req.setRequestHeader('Bugsnag-Api-Key', client._config.apiKey)
--- a/packages/delivery-xml-http-request/delivery.js
+++ b/packages/delivery-xml-http-request/delivery.js
@@ -1,5 +1,14 @@
 const payload = require('@bugsnag/core/lib/json-payload')
 
+function deliveryFailed (client, body, cb) {
+  const err = new Error('Event failed to send')
+  client._logger.error('Event failed to send…', err)
+  if (body.length > 10e5) {
+    client._logger.warn(`Event oversized (${(body.length / 10e5).toFixed(2)} MB)`)
+  }
+  cb(err)
+}
+
 module.exports = (client, win = window) => ({
   sendEvent: (event, cb = () => {}) => {
     try {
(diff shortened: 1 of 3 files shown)


packages/delivery-xml-http-request/delivery.js
```diff
@@ -11,20 +11,29 @@ module.exports = (client, win = window) => ({
...
       req.setRequestHeader('Bugsnag-Api-Key', event.apiKey || client._config.apiKey)
       req.setRequestHeader('Bugsnag-Payload-Version', '4')
       req.setRequestHeader('Bugsnag-Sent-At', (new Date()).toISOString())
-      req.send(body)
+      try {
+        req.send(body)
+      } catch (err) {
+        client._logger.error(err)
+        cb(err)
+      }
```
imjoehaines, line 36: I don't think this can throw — it's an asynchronous request, so this won't catch errors that are raised during the request and [MDN says `send` can only throw 2 errors](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send#exceptions):

- `InvalidStateError` if `send` has already been invoked for the request — this isn't relevant as we can't call `send` multiple times with this code
- `NetworkError` if the resource type to be fetched is a Blob, and the method is not GET — we're not using Blobs and are sending POST requests anyway

So I think this try/catch is unnecessary, unless you found it could throw errors in some other case?

gingerbenw: I think it only catches the implementation used in the e2e test (same as x-domain), which doesn't really feel right. I would probably prefer to add an onerror handler or event listener... I'll revisit this, particularly as it's one of our main delivery systems for browsers.

packages/delivery-xml-http-request/delivery.js
```diff
@@ -1,5 +1,14 @@
 const payload = require('@bugsnag/core/lib/json-payload')
 
+function deliveryFailed (client, body, cb) {
+  const err = new Error('Event failed to send')
```
imjoehaines, line 4: Can we add some useful information to the message here, e.g. the status code?

Maybe something like "Request failed with status ${status}"

packages/delivery-xml-http-request/delivery.js
```diff
@@ -11,14 +20,17 @@ module.exports = (client, win = window) => ({
...
+          } else {
+            cb(null)
           }
-          cb(null)
         }
       }
+
+      req.onerror = function () {
+        deliveryFailed(client, body, cb)
+      }
```
imjoehaines, line 32: I don't think we actually need this and it would cause the callback to be called twice (once here & once in the `readystatechange` listener)

The very first thing that happens when a request errors is for the `readyState` to be set to `done`:

> The request error steps for an [XMLHttpRequest](https://xhr.spec.whatwg.org/#xmlhttprequest) object xhr, event, and optionally exception are:
> 1. Set xhr’s [state](https://xhr.spec.whatwg.org/#concept-xmlhttprequest-state) to done.
> &hellip;
>
> https://xhr.spec.whatwg.org/#request-error-steps

Testing this locally, I couldn't find a scenario where `onerror` fires but `readystatechange` doesn't

`onerror` would be good if it provided access to the exception, but unfortunately that's only possible when it's a synchronous request

So I think we can safely remove this

```suggestion
```

imjoehaines (changes requested):

gingerbenw:
latest changes to the xhr mocks are a bit more sweeping, but a lot more similar to the real thing

packages/delivery-xml-http-request/delivery.js
```diff
@@ -11,14 +11,18 @@ module.exports = (client, win = window) => ({
         if (req.readyState === win.XMLHttpRequest.DONE) {
           const status = req.status
           if (status === 0 || status >= 400) {
-            client._logger.error('Event failed to send…')
+            const err = new Error(`Request failed with status ${status}`)
```
imjoehaines, line 14: Maybe worth applying this to `sendSession` as well — it doesn't look like we currently use/expose the `cb` parameter there but if that changes in the future then it'd be nice for it to not have the same bug

imjoehaines (approved):

PR #1939, opened 2023-03-14
[PLAT-9555] Update release script

gingerbenw:
## Goal

Previously, the release script would increment version numbers of changed packages after building, meaning that production bundles including the package version would be one version out of date on every release, skewing the internal metrics for version usage.


bin/release.sh
```diff
@@ -36,24 +50,17 @@ npx lerna run build \
   --ignore @bugsnag/plugin-electron-app \
   --ignore @bugsnag/plugin-electron-client-state-persistence
 
+# publish
+if [ -z RETRY_PUBLISH ]; then
+  npx lerna publish from-package
+else
+  npx lerna publish from-git
```
gingerbenw, line 57: Does this command push git tags to the remote? 

the [documentation](https://github.com/lerna/lerna/tree/main/libs/commands/publish#bump-from-git) suggests this is the command to use after versioning manually.

> This is useful in CI scenarios where you wish to manually increment versions, but have the package contents themselves consistently published by an automated process.

bin/release.sh
```diff
@@ -36,24 +50,17 @@ npx lerna run build \
   --ignore @bugsnag/plugin-electron-app \
   --ignore @bugsnag/plugin-electron-client-state-persistence
 
+# push latest git tag
+TAG=$(git describe --tags --abbrev=0)
+git push origin $TAG
+
+# publish
+npx lerna publish from-package
```
imjoehaines, line 58: I think we want `from-git` instead of `from-package`?

> bump from-git
>
> &hellip; This will identify packages tagged by lerna version and publish them to npm. This is useful in CI scenarios where you wish to manually increment versions, but have the package contents themselves consistently published by an automated process.

`from-package` still sounds best for `RETRY_PUBLISH` though

bin/release.sh
```diff
@@ -26,6 +26,20 @@ cd /app/bugsnag-js
...
 
+# increment package version numbers
+if [ ! RETRY_PUBLISH ]; then
+  case $VERSION in
+    "prerelease" | "prepatch" | "preminor" | "premajor")
+      npx lerna version "$VERSION" --dist-tag next --no-push
+      ;;
+
+    *)
+      npx lerna version "$VERSION" --no-push
```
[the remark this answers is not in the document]
gingerbenw: this is to prevent git tags being pushed until after we've built as per the steps in the ticket. that being said - we could allow tagging during the version and remove the later step to push the tags?

bin/release.sh
gingerbenw: that's what I initially thought - but I saw in a recent nx walkthrough of lerna they used it this way... let's stick with the documentation though

bin/release.sh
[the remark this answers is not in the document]
imjoehaines: Ah yeah — I figured since we've already built on CI it would always work but I guess there's no harm in being paranoid in case something changes between CI running and this script running

bin/release.sh
```diff
@@ -26,6 +26,20 @@ cd /app/bugsnag-js
 npm ci
 npm run bootstrap -- --ci
 
+# increment package version numbers
+if [ ! RETRY_PUBLISH ]; then
```
imjoehaines, line 30: [Spellcheck](https://www.shellcheck.net) doesn't like this:

> ^-- [SC2078](https://www.shellcheck.net/wiki/SC2078) (error): This expression is constant. Did you forget a $ somewhere?

```suggestion
if [ -z "$RETRY_PUBLISH" ]; then
```

bin/release.sh
```diff
@@ -36,24 +50,21 @@ npx lerna run build \
   --ignore @bugsnag/plugin-electron-app \
   --ignore @bugsnag/plugin-electron-client-state-persistence
 
-# check if CDN packages changed – if they didn't we don't need to upload to the CDN
-BROWSER_PACKAGE_CHANGED=$(npx lerna changed --parseable | grep -c packages/js$ || test $? = 1;)
-WORKER_PACKAGE_CHANGED=$(npx lerna changed --parseable | grep -c packages/web-worker$ || test $? = 1;)
+# push latest git tag
+TAG=$(git describe --tags --abbrev=0)
+git push origin $TAG
```
imjoehaines, line 55:
```suggestion
git push origin "$TAG"
```

imjoehaines: Do we need to push 1 tag specifically, or would `git push --tags` be OK?

bin/release.sh
```diff
@@ -36,24 +50,21 @@ npx lerna run build \
...
 
-if [ -v RETRY_PUBLISH ]; then
+# publish
+if [ ! RETRY_PUBLISH ]; then
```
imjoehaines, line 58:
```suggestion
if [ -z "$RETRY_PUBLISH" ]; then
```

bin/release.sh
gingerbenw: I think push --tags would be okay actually, as it's from a fresh docker container

yousif-bugsnag (approved):

PR #1947, opened 2023-04-14
Adding a Gatsby example app

hannah-smartbear:
Creating a Gatsby example app with BugSnag integrated. We can then direct any customers setting BugSnag up in their app to the example app, since we don't have official support for Gatsby. Similar to what we do with Next.js.

Created using Gatsby's quick start instructions (npm init Gatsby)


tomlongridge:
This has now been done in #2054

Issue #1945, opened 2023-04-06
interaction-breadcrumbs.js click handler chokes on whitespace


alcohol:
### Describe the bug

Bugsnagjs has an issue with a page that contains somewhat questionable html markup. The page becomes completely frozen and chrome offers to kill it once I 'click' anywhere.

```
[Violation] 'click' handler took 85672ms bugsnag.min.js:1
```

It looks like in `interaction-breadcrumbs.js` the following value `let text = el.textContent || el.innerText || ''` contains an insane amount of whitespace, which is causing the `text = text.replace(/^\s+|\s+$/g, '') // trim whitespace` to basically take up an enormous amount of time. Since the next line is only taking the first 140 characters of this (`return truncate(text, 140)`), I reckon this could probably be improved upon somehow?

Additional details:

```javascript
el.textContent.length
// 438228

let txt = el.textContent
console.time('replace'); 
txt = txt.replace(/^\s+|\s+$/g, ''); 
console.timeEnd('replace');
// replace: 78925.93701171875 ms
```

alcohol:
Cheers :-)

alcohol:
Out of curiousity, how often is your CDN hosted version updated? We're still receiving 7.19.0 at https://d2wy8f7a9ursnm.cloudfront.net/v7/bugsnag.min.js

imjoehaines:
> Out of curiousity, how often is your CDN hosted version updated? We're still receiving 7.19.0 at https://d2wy8f7a9ursnm.cloudfront.net/v7/bugsnag.min.js

Hey @alcohol, thanks for pointing this out! `v7.20.2` is now on the CDN

PR #1948, opened 2023-04-14
Interaction breadcrumb hanging

imjoehaines:
## Goal

Fixes an issue where the interaction breadcrumb plugin could hang when an element had a huge amount of whitespace in between non-whitespace characters, e.g.:

```
<button>[whitespace] a [whitespace] b [whitespace]</button>
```

This would completely freeze the page until the regex finished, which could take multiple seconds

See #1945

### DIFF
--- a/packages/plugin-interaction-breadcrumbs/interaction-breadcrumbs.js
+++ b/packages/plugin-interaction-breadcrumbs/interaction-breadcrumbs.js
@@ -21,12 +21,23 @@ module.exports = (win = window) => ({
   }
 })
 
-// extract text content from a element
-const getNodeText = el => {
+const trimStart = /^\s+/
+const trimEnd = /(^|[^\s])\s+$/
+
+function getNodeText (el) {
   let text = el.textContent || el.innerText || ''
-  if (!text && (el.type === 'submit' || el.type === 'button')) text = el.value
-  text = text.replace(/^\s+|\s+$/g, '') // trim whitespace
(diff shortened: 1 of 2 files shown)


packages/plugin-interaction-breadcrumbs/interaction-breadcrumbs.js
```diff
@@ -53,8 +82,36 @@ function getNodeSelector (el, win) {
...
+    case '\u2007': // figure space
+    case '\u2008': // punctuation space
+    case '\u2009': // thin space
+    case '\u200a': // hair space
+    case '\u2028': // line separator
+    case '\u2029': // paragraph separator
+    case '\u202f': // narrow no-break space
+    case '\u205f': // medium mathematical space
+    case '\u3000': // ideographic space
+    case '\ufeff': // zero width no-break space
```
imjoehaines, line 111: This matches the previous behaviour[^1], but is maybe excessive — we could save a lot of bytes by not counting more esoteric whitespace here. I think the only common values would be ` `, `\t`, `\n` & `\r` and especially only those as repeating characters — i.e. I'd expect multiple ` ` or `\t` for indentation and `\n` & `\r` for new lines, but I wouldn't expect many repeated hair spaces

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#white_space

gingerbenw (approved):

PR #1950, opened 2023-04-21
Update BitBar tests to use the latest version of Edge

gingerbenw:
## Goal

Edge version 106 has been removed from the BitBar capabilities creator, this change updates the target version to the latest supported version of Edge in BitBar

### DIFF
--- a/.buildkite/browser-pipeline.yml
+++ b/.buildkite/browser-pipeline.yml
@@ -90,7 +90,7 @@ steps:
         concurrency: 5
         concurrency_group: "bitbar-web"
 
-      - label: ":edge: v106 Browser tests"
+      - label: ":edge: v107 Browser tests"
         depends_on: "browser-maze-runner"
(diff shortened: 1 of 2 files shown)


.buildkite/browser-pipeline.yml
```diff
@@ -90,7 +90,7 @@ steps:
         concurrency: 5
         concurrency_group: "bitbar-web"
 
-      - label: ":edge: v106 Browser tests"
+      - label: ":edge: Browser tests"
```
twometresteve, line 93: `:edge:` will be transformed into the Edge icon by Buildkite, so I don't know if `:edge: Latest Edge browser tests` would be clearly to read in the Buidlkite UI?  It's up to you, I would just make the wording consistent for each such job in the pipeline (i.e. Firefox/Chrome as well).

twometresteve (approved):

PR #1957, opened 2023-05-04
Adapt Android RN test fixture to work with BitBar

Cawllec:
## Goal

This is a required change for Android test fixtures to run using BitBar as a device farm under our current setup.  Cocoa was initially in scope, but is being omitted as further work is required.

## Design

The logic to read the files is pulled directly from equivalent test fixtures used in the Android and Cocoa repos.  Instead of passing maze-runner addresses up to the top RN layer, instead the fixture will test to see if a specific notify/session endpoint has been input when starting BugSnag, only attempting to read from the config file if one hasn't been set.

This keeps all logic in the native layer, requiring minimal changes in JS or interaction between the code layers.

## Testing

The tests will be independently updated to target BitBar as a device farm provider, although I intend to move one of the current test runs over as evidence of the change working successfully.  In addition, manual testing will be performed to ensure the real dashboard and user-input options for BugSnag endpoints work as expected.

### DIFF
--- a/test/react-native/features/fixtures/ios-module/BugsnagModule.m
+++ b/test/react-native/features/fixtures/ios-module/BugsnagModule.m
@@ -10,6 +10,7 @@
 
 #import <Bugsnag/Bugsnag.h>
 #import "BugsnagModule.h"
+#import "ConfigFileReader-Bridging-Header.h"
 #import "Scenario.h"
 
 @implementation BugsnagModule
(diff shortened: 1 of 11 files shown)


test/react-native/features/fixtures/ios-module/BugsnagModule.m
```diff
@@ -55,10 +55,21 @@ @implementation BugsnagModule
...
-  NSString *notifyEndpoint = endpointsIn[@"notify"];
-  NSString *sessionsEndpoint = endpointsIn[@"sessions"];
+  NSString *notifyEndpoint;
+  NSString *sessionsEndpoint;
+  if (options[@"endpoints"] != nil && options[@"endpoints"][@"notify"] != nil && options[@"endpoints"][@"sessions"] != nil) {
+    NSDictionary *endpointsIn = options[@"endpoints"];
+    notifyEndpoint = endpointsIn[@"notify"];
+    sessionsEndpoint = endpointsIn[@"sessions"];
+  } else {
+    NSString *baseAddress = @"bs-local.com:9339";
```
Cawllec, line 65: This will be replaced with the file loading logic in the below swift file once it can be compiled correctly

gingerbenw (approved):

PR #1967, opened 2023-06-01
Move React-native CLI android tests to using BitBar.

Cawllec:
## Goal

To switch react-native-cli tests running on Android to use BitBar as a device farm provider instead of BrowserStack.

### Changes

- Docker-compose definitions and pipeline files have been updated in line with Bugsnag-Android to enable the test harness to provide the test fixture with its remote address.
- New methods have been added to the native side of the tests to:
  - Load the maze-runner added config file and acquire the maze-runner endpoint from it
  - Convert the original configuration (as provided by the CLI script) into metadata and attach it to the Bugsnag payloads
- The build script has been updated to find and replace the `Bugsnag.start` with default arguments (as added by the CLI script) with a line that calls a new method, creating the custom configuration with the above attributes.

### DIFF
--- a/test/react-native-cli/features/fixtures/rn0_69/android/app/src/main/java/com/rn0_69/TestUtils.kt
+++ b/test/react-native-cli/features/fixtures/rn0_69/android/app/src/main/java/com/rn0_69/TestUtils.kt
@@ -16,7 +16,7 @@ const val ENDPOINT_SESSIONS = "$BUGSNAG_NS.ENDPOINT_SESSIONS"
 class TestUtils {
 
     fun getMazeRunnerAddress(context: Context): String {
-        val externalFilesDir = File("/sdcard/Android/data/com.rn0_69/files")
+        val externalFilesDir = context.getExternalFilesDir(null)
         val configFile = File(externalFilesDir, "fixture_config.json")
         var mazeAddress: String? = null
         Log.i("Bugsnag", "Attempting to read Maze Runner address from config file ${configFile.path}")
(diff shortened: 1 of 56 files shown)


test/react-native-cli/features/steps/steps.rb
```diff
@@ -209,11 +209,7 @@ def parse_package_json
 end
 
 def get_android_main_application_path(current_fixture)
-  if current_fixture.include? 'expo_ejected'
-    "android/app/src/main/java/com/bugsnag/#{current_fixture}/MainApplication.java"
-  else
-    "android/app/src/main/java/com/#{current_fixture}/MainApplication.java"
-  end
+  "android/app/src/main/java/com/#{current_fixture}/MainApplication.java"
```
Cawllec, line 212: This is no longer required since the `rn0_63_expo_ejected` fixture has the same format as the rest of the fixtures.

Cawllec:
I've refactored the commits so each should cover one fixture/set of code changes.  I'd suggest reviewing commit by commit if possible.

test/react-native-cli/features/fixtures/rn0_69/android/app/src/main/java/com/rn0_69/TestUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+const val CONFIG_FILE_TIMEOUT = 15000
+const val BUGSNAG_NS = "com.bugsnag.android"
+const val API_KEY = "$BUGSNAG_NS.API_KEY"
+const val ENDPOINT_NOTIFY = "$BUGSNAG_NS.ENDPOINT_NOTIFY"
+const val ENDPOINT_SESSIONS = "$BUGSNAG_NS.ENDPOINT_SESSIONS"
+
+class TestUtils {
+
+    fun getMazeRunnerAddress(context: Context): String {
+        val externalFilesDir = File("/sdcard/Android/data/com.rn0_69/files")
```
twometresteve, line 19: Should perhaps be like the other fixtures?
```
val externalFilesDir = context.getExternalFilesDir(null)
```

twometresteve (approved):
LGTM pending CI.

PR #1969, opened 2023-06-08
Move latest browser tests to BitBar

twometresteve:
## Goal

Move latest browser tests to BitBar, using public IP addresses.

## Changeset

- Buildkite steps for running tests on the latest desktop browsers moved to BitBar.
- Docker Compose file refactored to reduce duplication across services.
- Separate Docker images built for BB and BS, as the Docker Compose service for BB needs to use service ports.
- Docker Compose Buildkite plugin updated across the board
- Manually implemented web service for hosting the test fixture replaced with the document server built into Maze Runner.

## Testing

Covered by CI.


gingerbenw (approved):
Looks good to me!

PR #1973, opened 2023-06-14
Support ReactNative New architecture

lemnik:
## Goal
Support apps using ReactNative [New Architecture](https://reactnative.dev/docs/the-new-architecture/landing-page).

## Design
The Bugsnag native components have been adapted to be usable as both Native Modules or Turbo Modules, depending on the configuration of the app being built. A Turbo Module `Spec` was added to define the functions that can be triggered by the existing Bugsnag ReactNative JS code.

When loading the Bugsnag ReactNative `NativeClient` we check whether Turbo Modules are enabled and use the `TurboModulesRegistry` to access the native code if they are.

### Android
For Android this involves splitting the Java code into 3 components:
- The common code, which was extracted from the existing module (and resides in the `main` sourceset)
- A Turbo Module facade - which implements the `codegen` class and forwards the method call to the common module
- A Native Module facade - which includes the `@ReactMethod` annotations and delegates to the common module

### iOS
The iOS TurboModule is enabled by using conditional compilation flags (`#ifdef`).

## Testing
A new test fixture was added for ReactNative 0.71 with New Architecture enabled.


packages/react-native/android/build.gradle
```diff
@@ -1,20 +1,28 @@
...
+//        mavenCentral()
+//        gradlePluginPortal()
+//        maven { // add this to use snapshots
+//            url "https://oss.sonatype.org/content/repositories/snapshots/"
+//        }
+//    }
+//    dependencies {
+//        classpath "com.android.tools.build:gradle:${project.ext.androidToolsVersion}"
+//    }
+//}
```
yousif-bugsnag, line 16: not sure I fully understand why this is commented out - can it be removed?

yousif-bugsnag:
changes LGTM, but the CI pipeline now needs updating as we've moved all the RN tests over to bitbar now

yousif-bugsnag (approved):

PR #1979, opened 2023-06-21
[PLAT-10144] Update React Native CLI to install BugSnag CLI

joshedney:
## Goal

Update the React Native CLI  to install BugSnag CLI instead of the Source Maps package for the Android Integration. 

The RN 0.67 apk (Hermes) CLI build/test have been disabled due to issues with Hermes replacing the mapping file with its own Hermes binary file. This should be covered via the BAGP tests.

## Changeset

Update the `AutomateSymbolicationCommand.ts` to install the BugSnag CLI package for the Android integration.

## Testing

Covered by CI


packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
```diff
@@ -38,15 +38,16 @@ export default async function run (projectRoot: string, urls: OnPremiseUrls): Pr
     const { androidIntegration } = await prompts({
       type: 'confirm',
       name: 'androidIntegration',
-      message: 'Do you want to automatically upload JavaScript source maps as part of the Gradle build?',
+      message: 'Do you want to install the BugSnag CLI to allow you to upload JavaScript source maps?',
```
tomlongridge, line 41: I think that's probably TMI for the user at this point. All they probably care about it the outcome:

```suggestion
      message: 'Do you want to automatically upload JavaScript source maps for Android builds?',
```

packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
```diff
@@ -38,15 +38,16 @@ export default async function run (projectRoot: string, urls: OnPremiseUrls): Pr
...
       initial: true
     }, { onCancel })
 
     if (androidIntegration) {
       await enableReactNativeMappings(projectRoot, urls[UrlType.UPLOAD], urls[UrlType.BUILD], logger)
```
tomlongridge, line 46: In fact, I think we'd be better removing this task entirely as `uploadReactNativeMappings` defaults to false: there's no need to meddle with their Gradle files at all in a fresh project.

Better again would be to check for the presence of this setting and warn the user to disable it.

@lemnik – thoughts?

.buildkite/react-native-cli-pipeline.yml
```diff
@@ -339,20 +339,20 @@ steps:
...
-        env:
-          DEBUG: true
-          REACT_NATIVE_VERSION: rn0_67_hermes
-        plugins:
-          - docker-compose#v4.12.0:
-              run: react-native-cli-android-builder
-              command: ["features/build-app-tests/build-android-app.feature"]
-        artifact_paths:
-          - build/rn0_67_hermes.apk
+#      - label: ":android: Init and build RN 0.67 apk (Hermes)"
```
tomlongridge, line 342: > The RN 0.67 apk (Hermes) CLI build/test have been disabled due to issues with Hermes replacing the mapping file with its own Hermes binary file. This should be covered via the BAGP tests.

So are we effectively saying that we don't support Hermes on RN < 0.68/69?

packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
tomlongridge: Agreed in meeting to produce a warning here if `uploadReactNativeMappings` is found and set to `true` in the Gradle file.

.buildkite/react-native-cli-pipeline.yml
tomlongridge: Agreed in meeting to add an error that advises people on < RN 0.68:

> You are running a version of React Native that we cannot automatically integrate with due to known issues with the build when Hermes is enabled. If you cannot upgrade to a later version of React Native (version 0.68 or above), you can use an older version of this CLI (version 7.20.x or earlier) or follow the manual integration instructions in our online docs: https://docs.bugsnag.com/platforms/react-native/react-native/manual-setup/

.buildkite/react-native-cli-pipeline.yml
```diff
@@ -708,26 +708,26 @@ steps:
...
-            command:
-              - --app=build/rn0_67_hermes.apk
-              - --farm=bs
-              - --device=ANDROID_12_0
-              - --a11y-locator
-              - features/run-app-tests
-        concurrency: 5
-        concurrency_group: "browserstack-app"
-        concurrency_method: eager
+#      - label: ":runner: RN 0.67 Android end-to-end tests (Hermes)"
```
tomlongridge, line 711: Remove Hermes tests for 0.67 as this is officially not supported now.

(Also agreed to introduce a Hermes enabled 0.69 test.)

packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
joshedney: @tomlongridge is this still the wording that we want to go with ?

PR #1998, opened 2023-08-01
[PLAT-10694] Fix inline script metadata in Safari 16

gingerbenw:
## Goal

Update inline script url comparison to ensure metadata is added in Safari 16.5 where the value of `stack.frame` no longer includes a query string

## Design

Remove query string from url for inline script detection


gingerbenw:
Raised PLAT-10754 to cover updating the list of globals, as it drastically affects the scope of this change.

yousif-bugsnag:
Can you update the PR title and description with a bit more info for future travellers?

yousif-bugsnag (approved):

PR #2003, opened 2023-08-11
startSession should no longer clone the client

djskinner:
## Goal

Update startSession so that it starts a session in the current client, rather than cloning the client. The method should behave like the other client methods, so that calling `Bugsnag.notify` or `req.bugsnag.notify` afterwards would be in the new session.

### DIFF
--- a/packages/plugin-express/src/express.js
+++ b/packages/plugin-express/src/express.js
@@ -13,9 +13,11 @@ module.exports = {
   name: 'express',
   load: client => {
     const requestHandler = (req, res, next) => {
-      // Get a client to be scoped to this request. If sessions are enabled, use the
-      // resumeSession() call to get a session client, otherwise, clone the existing client.
-      const requestClient = client._config.autoTrackSessions ? client.resumeSession() : clone(client)
+      // clone the client to be scoped to this request. If sessions are enabled, start one
+      const requestClient = clone(client)
+      if (requestClient._config.autoTrackSessions) {
+        requestClient.startSession()
+      }
 
       // attach it to the request
       req.bugsnag = requestClient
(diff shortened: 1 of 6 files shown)


packages/plugin-server-session/test/session.test.ts
```diff
@@ -130,28 +106,28 @@ describe('plugin: server sessions', () => {
 
     const c = new Client({ apiKey: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' }, undefined, [plugin])
 
-    // start a session and get its id
-    const sessionClient = c.startSession()
-    const sid0 = (sessionClient._session as Session).id
+    // start a session and get its idc.startSession()
```
imjoehaines, line 109:
```suggestion
    // start a session and get its id
```

packages/plugin-express/src/express.js
```diff
@@ -13,9 +13,11 @@ module.exports = {
   name: 'express',
   load: client => {
     const requestHandler = (req, res, next) => {
-      // Get a client to be scoped to this request. If sessions are enabled, use the
-      // resumeSession() call to get a session client, otherwise, clone the existing client.
-      const requestClient = client._config.autoTrackSessions ? client.resumeSession() : clone(client)
+      // clone the client to be scoped to this request. If sessions are enabled, start one
```
imjoehaines, line 16: Would we need to clone the client if we used the stored value? i.e.:

```js
if (client._config.autoTrackSessions) {
  client._clientContext.getStore().startSession()
  // or even:
  Bugsnag.startSession()
}
```

djskinner: What about other things like metadata addOnError and breadcrumbs? If we don't clone the client then wouldn't add these things added during the request then be present on all subsequent requests?

imjoehaines: I thought those things would only apply in the current context after the other changes

I guess you'd also need:

```js
req.bugsnag = client._clientContext.getStore()
```

djskinner: But it's the cloned client we put into the store and its that cloned client we retrieve whenever the context is accessed

imjoehaines: Ohh I see 🤦‍♀️ that makes sense!

PR #2013, opened 2023-09-19
Run Lerna bootstrap after copying packages into Docker containers

twometresteve:
## Goal

Fix the broken Docker image build steps.

## Design

It seemed to start failing after an update to our Buildkite Elastic CI Stack for AWS:
```
1.506 lerna ERR! lerna EACCES: permission denied, rename '/app/packages/plugin-react-navigation/package.json' -> '/app/packages/plugin-react-navigation/package.json.lerna_backup'
```

I assume that a new version of Docker meant that file paths not existing in the container are not treated different.  If Lerna is renaming files like then then it seems reasonable that the files should exist in the first place in the contrainer.

## Changeset

Copying of `packages` into the containers moved to before the `lerna bootstrap`.

## Testing

Covered by CI.


gingerbenw (approved):
👍🏻

PR #2022, opened 2023-10-10
Fix Electron Linux crash

lemnik:
## Goal
Fix an Electorn crash on Linux that can be triggered on startup, when we attempt to access the `electron.powerManager` before it is fully accessible.

Accessing the powerManager before `app.whenReady` can cause Electron to [access the Chromium Bluez stack](https://github.com/electron/electron/blob/86df4db6f10afbe5872f8ecbcbfe6191374fc1f7/shell/browser/lib/power_observer_linux.cc#L37C3-L37C3) (in order to reach the dbus client) before the Bluez code has been properly initialized. This in turn causes an [assertion in the Chromium Bluez integration](https://chromium.googlesource.com/chromium/src/+/3e60c04292e92cc617937515f0c9a9df826bcc0f/device/bluetooth/dbus/bluez_dbus_thread_manager.cc#68) to fail, triggering an [immediate crash](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/base/immediate_crash.h).

## Testing
Updated CI to cover more versions of Electron and Linux, and tested both the original crash and fix manually.

This PR removes two of the unit tests which are problematic on Linux as they used complete Electron stacks which (depending on the distro) caused Jest to either not run the test at all, or hang while testing. Future tasks will be raised to fix these tests.


twometresteve (commented):
Looks good from my perspective.

packages/plugin-electron-device/device.js
```diff
@@ -45,15 +45,51 @@ module.exports = (app, screen, process, filestore, NativeClient, powerMonitor) =
...
+        {},
+        event.device,
+        device,
+        {
+          freeMemory: kibibytesToBytes(process.getSystemMemoryInfo().free),
+          time: new Date()
+        }
+      )
+
+      if (powerMonitor && powerMonitor.getSystemIdleTime) {
```
imjoehaines, line 86: Why is this if necessary now?

packages/plugin-electron-device/device.js
```diff
@@ -45,15 +45,51 @@ module.exports = (app, screen, process, filestore, NativeClient, powerMonitor) =
...
+        event.addMetadata('device', { idleTime: powerMonitor.getSystemIdleTime() })
+      }
+
+      setDefaultUserId(event)
```
imjoehaines, line 90: I don't think this should be inside the `whenReady` as it'll mean events that are triggered before the app is ready will be missing a user ID

lemnik: This isn't part of the `whenReady`, that ends on line 73 and the next one picks up on line 93.

packages/plugin-electron-device/device.js
lemnik: It probably isn't. I'll remove it.

PR #2025, opened 2023-10-13
Fix Electron unhandled secondary errors during delivery

### LINKED ISSUE #2023: @bugsnag/electron: infinite requests loop to notify.bugsnag.com
👋 Thanks for working on this project.

### Environment
Bugsnag version: "@bugsnag/electron": "^7.21.0" (latest at the time of writing)

### Describe the bug

We are getting an infinite loop of requests hitting `notify.bugsnag.com` (default Bugsnag endpoints configuration) while using `@bugsnag/electron`

https://github.com/bugsnag/bugsnag-js/blob/86e96f9ad275f870dff28d3139bd8f07f387ec42/packages/electron/src/config/common.js#L26-L28

After digging into the code, a possible cause is:

- Bugsnag tried to [send](https://github.com/bugsnag/bugsnag-js/blob/86e96f9ad275f870dff28d3139bd8f07f387ec42/packages/delivery-electron/delivery.js#L151) a session request on app startup (which might help calculate [stability score](https://www.bugsnag.com/product/stability-score/) or something). Nothing could stop this request, even by specifying `autoTrackSessions: false` (documentation [here](https://docs.bugsnag.com/platforms/electron/configuration-options/#autotracksessions))
- When we have a corporate proxy that requires authentication/or DNS blocker (pi-hole/Adguard... which automatically blocks Bugsnag), this session request failed with 407 (Proxy Authentication Required)

Screenshot 2023-10-11 at 09 59 56

- This failure somehow triggered an `uncaughtException` that was handled in [this handler plugin](https://github.com/bugsnag/bugsnag-js/blob/86e96f9ad275f870dff28d3139bd8f07f387ec42/packages/plugin-node-uncaught-exception/uncaught-exception.js#L12). This plugin comes builtin with `@bugsnag/electron`

https://github.com/bugsnag/bugsnag-js/blob/86e96f9ad275f870dff28d3139bd8f07f387ec42/packages/plugin-node-uncaught-exception/uncaught-exception.js#L12

which tried to notify Bugsnag endpoint `notify.bugsnag.com`. Like session request, notify request also failed with 407, which then triggered another `uncaughtException` which notified `notify.bugsnag.com` which then failed => infinite loop!

This happens in real world of corporates with proxy setup, that consumes CPU cycles of our machines. That happened to our end users' machines too.

### Steps to reproduce
1. Use `@bugsnag/electron` in a dummy Electron project
2. Setup a local proxy, for the sake of simplicity, I use https://mitmproxy.org/, detail of setup could be found in this [blog](https://earthly.dev/blog/mitmproxy/)
3. Run `mitmproxy --proxyauth foo:bar --showhost`
4. Configure your OS system to use `127.0.0.1:8080` as proxy (default address of mitm)
5. Start Electron application and see the infinite Bugsnag loop in the console

Below is real log from our corporate proxy

bugsnag

For now we put a dirty patch in @bugsnag/delivery-electron before this call [here](https://github.com/bugsnag/bugsnag-js/blob/86e96f9ad275f870dff28d3139bd8f07f387ec42/packages/delivery-electron/delivery.js#L96-L98) to avoid calling it infinitely in case we hit 407. It's clearly not a long term viable solution.

Another possibility would be disabling [enablederrortypes](https://docs.bugsnag.com/platforms/electron/configuration-options/#enablederrortypes) to avoid this infinite loop, but that would defeat the purpose of Bugsnag as bug tracking tool.

Please let me know if anything I can help to fix it.

TIA for feedbacks!

yousif-bugsnag:
## Goal

Fixes #2023  

In some cases (e.g. proxy authentication errors) Electron net will emit error events on both the request and response after the response is received.

If these are not handled this results in further uncaught exceptions, causing a potentially infinite loop of failed requests - if the error response has already been handled these secondary errors should be ignored.

### DIFF
--- a/packages/delivery-electron/delivery.js
+++ b/packages/delivery-electron/delivery.js
@@ -6,28 +6,46 @@ const NetworkStatus = require('@bugsnag/electron-network-status')
 
 const delivery = (client, filestore, net, app) => {
   const send = (opts, body, cb) => {
-    const errorHandler = err => {
-      err.isRetryable = true
-      cb(err)
+    // an error can still happen on both the request and response even after a response is received,
+    // so we store a reference to the response error to ensure this is only handled once
+    let responseError
+
+    const requestErrorHandler = requestError => {
+      if (responseError) return
+
+      requestError.isRetryable = true
+      cb(requestError)
+    }
+
+    const responseErrorHandler = (response, err) => {
+      if (responseError) return
+
+      responseError = err
+      responseError.isRetryable = isRetryable(response.statusCode)
+      // do not retry oversized payloads regardless of status code
+      if (body.length > 10e5) {
+        client._logger.warn(`Discarding over-sized event (${(body.length / 10e5).toFixed(2)} MB) after failed delivery`)
+        responseError.isRetryable = false
+      }
+
+      cb(responseError)
     }
 
     const req = net.request(opts, response => {
-      req.removeListener('error', errorHandler)
+      // handle errors on the response stream
+      response.on('error', err => {
+        responseErrorHandler(response, err)
+      })
+
       if (isOk(response)) {
         cb(null)
       } else {
         const err = new Error(`Bad status code from API: ${response.statusCode}`)
-        err.isRetryable = isRetryable(response.statusCode)
-        // do not retry oversized payloads regardless of status code
-        if (body.length > 10e5) {
-          client._logger.warn(`Discarding over-sized event (${(body.length / 10e5).toFixed(2)} MB) after failed delivery`)
-          err.isRetryable = false
-        }
-        cb(err)
+        responseErrorHandler(response, err)
       }
     })
 
-    req.on('error', errorHandler)
+    req.on('error', requestErrorHandler)
 
     try {
       req.write(body)
--- a/packages/delivery-electron/delivery.js
+++ b/packages/delivery-electron/delivery.js
@@ -5,47 +5,43 @@ const PayloadDeliveryLoop = require('./payload-loop')
 const NetworkStatus = require('@bugsnag/electron-network-status')
 
 const delivery = (client, filestore, net, app) => {
+  const noop = () => {}
+
   const send = (opts, body, cb) => {
-    // an error can still happen on both the request and response even after a response is received,
-    // so we store a reference to the response error to ensure this is only handled once
-    let responseError
+    let errorCallback = (error, response) => {
+      // an error can still happen on both the request and response even after a response is received,
+      // so we noop on subsequent calls to ensure this is only handled once
+      errorCallback = noop
 
-    const requestErrorHandler = requestError => {
-      if (responseError) return
-
-      requestError.isRetryable = true
-      cb(requestError)
-    }
-
-    const responseErrorHandler = (response, err) => {
-      if (responseError) return
-
-      responseError = err
-      responseError.isRetryable = isRetryable(response.statusCode)
-      // do not retry oversized payloads regardless of status code
-      if (body.length > 10e5) {
-        client._logger.warn(`Discarding over-sized event (${(body.length / 10e5).toFixed(2)} MB) after failed delivery`)
-        responseError.isRetryable = false
+      if (response) {
+        error.isRetryable = isRetryable(response.statusCode)
+        // do not retry oversized payloads regardless of status code
+        if (body.length > 10e5) {
+          client._logger.warn(`Discarding over-sized event (${(body.length / 10e5).toFixed(2)} MB) after failed delivery`)
+          error.isRetryable = false
+        }
+      } else {
+        error.isRetryable = true
       }
 
-      cb(responseError)
+      cb(error)
     }
 
     const req = net.request(opts, response => {
       // handle errors on the response stream
       response.on('error', err => {
-        responseErrorHandler(response, err)
+        errorCallback(err, response)
       })
 
       if (isOk(response)) {
         cb(null)
       } else {
         const err = new Error(`Bad status code from API: ${response.statusCode}`)
-        responseErrorHandler(response, err)
+        errorCallback(err, response)
       }
     })
 
-    req.on('error', requestErrorHandler)
+    req.on('error', err => errorCallback(err))
 
     try {
       req.write(body)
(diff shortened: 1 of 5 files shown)


packages/delivery-electron/delivery.js
```diff
@@ -13,6 +13,14 @@ const delivery = (client, filestore, net, app) => {
 
     const req = net.request(opts, response => {
       req.removeListener('error', errorHandler)
+
+      // in some cases (e.g. proxy authentication errors) Electron net
+      // will emit an error event on the response stream
+      response.on('error', (err) => {
+        err.isRetryable = isRetryable(response.statusCode)
+        cb(err)
```
imjoehaines, line 21: Will this not call `cb` twice? Once on line 25 & once here

yousif-bugsnag: Well yes, actually it's the `cb` call on line 34 because the response isn't successful, but it's even worse than that! 

It looks like Electron actually emits errors on both the response _and_ the request (as well as returning the response) - so just handling the response error isn't enough. 

We do have an error handler on the request, but this is removed at the start of the response callback on line 15, presumably to prevent the request error handler being called if there's a response. 

The problem is that without that error handler, the request error will also fall through to the uncaught exception handler, and so we still end up in the same infinite request loop even if we handle the response error.

I think we'll need to remove line 15 and keep both of these error handlers - maybe we should have some kind of `errorHandled` flag which we set before calling `cb`, so that a failed request is only handled once by one of these 3 handlers?

yousif-bugsnag: Updated - if the response error has already been handled any subsequent errors on the request/response will be ignored

imjoehaines (commented):
I wonder if swapping out `cb` is a little cleaner? e.g.

```js
const noop = () => {}

const send = (opts, body, cb) => {
  let callback = (error) => {
    callback = noop
    cb(error)
  }

  // ...
}
```

packages/delivery-electron/test/delivery.test-main.ts
```diff
@@ -337,6 +337,60 @@ describe('delivery: electron', () => {
...
+    } as unknown as EventDeliveryPayload
+
+    const err = new Error('Error after response received')
+
+    // create a mock response object that emits an error event
+    const response = {
+      statusCode: 407,
+      statusMessage: STATUS_CODES[407],
+      on (_event: any, cb: any) {
+        if (_event === 'error') {
```
imjoehaines, line 352:
```suggestion
      on (event: any, cb: any) {
        if (event === 'error') {
```

packages/delivery-electron/test/delivery.test-main.ts
```diff
@@ -337,6 +337,60 @@ describe('delivery: electron', () => {
...
+      }
+    }
+
+    let requestErrorCallback: any = () => {}
+
+    // create a mock net instance that will return a response and also emit errors
+    const net = {
+      request: (opts: any, responseCallback: any) => ({
+        on (_event: any, cb: any) {
+          if (_event === 'error') {
```
imjoehaines, line 364:
```suggestion
        on (event: any, cb: any) {
          if (event === 'error') {
```

packages/delivery-electron/test/delivery.test-main.ts
```diff
@@ -337,6 +337,60 @@ describe('delivery: electron', () => {
...
+
+    const logger = { error: jest.fn(), info: jest.fn() }
+    const config = {
+      apiKey: 'aaaaaaaa',
+      endpoints: { notify: 'http://localhost:9999/events/' },
+      redactedKeys: []
+    }
+
+    delivery(filestore, net, app)(makeClient(config, logger)).sendEvent(payload, (err: any) => {
+      expect(err).toBe(err)
```
imjoehaines, line 384: This asserts the same variable is itself which is always true

trungutt:
👋 Would we need to have something similar in `@bugsnag/plugin-electron-deliver-minidumps` as well?

packages/delivery-electron/delivery.js
```diff
@@ -6,28 +6,46 @@ const NetworkStatus = require('@bugsnag/electron-network-status')
...
-      err.isRetryable = true
-      cb(err)
+    // an error can still happen on both the request and response even after a response is received,
+    // so we store a reference to the response error to ensure this is only handled once
+    let responseError
+
+    const requestErrorHandler = requestError => {
+      if (responseError) return
+
+      requestError.isRetryable = true
```
trungutt, line 16: **thought:** I didn't have a good overall understanding of bugsnag codebase, but I'd suggest (a very aggressive approach) `isRetryable = false` for any network errors.

In terms of Desktop applications, we don't know (or even cannot predict) how users setup their machines locally. Things like firewalls, enterprise proxies, dns blockers (add blockers) or unexpected stuff would make the request to Bugsnag fail. We should be really careful when it comes to retryable request, to avoid any possible infinite loop

yousif-bugsnag:
> I wonder if swapping out `cb` is a little cleaner? e.g.
> 
> ```js
> const noop = () => {}
> 
> const send = (opts, body, cb) => {
>   let callback = (error) => {
>     callback = noop
>     cb(error)
>   }
> 
>   // ...
> }
> ```

yeah that's probably a little cleaner, but we'll need to funnel both response and request errors through the same callback, e.g.
```js
const noop = () => {}

const send = (opts, body, cb) => {
  let callback = (error, response) => {
    callback = noop
	
  if (response) {
  	error.isRetryable = isRetryable(response.statusCode)
	// ...
  } else {
  	error.isRetryable = true
  }

    cb(error)
  }

  // ...
}
```

packages/delivery-electron/delivery.js
yousif-bugsnag: Thanks for the comment @trungutt, and for raising this issue initially!

The `isRetryable` flag is there to determine whether events that fail to send should be persisted and retried later or whether they should be discarded completely.

The potential for an infinite request loop isn't related to whether or not an error is marked as retryable - it happens because the error events emitted by Electron net on the request object and response stream after a response is received were not being handled, so these would then fall through to the uncaught exception handler which would then try and notify that error as a separate, duplicate error regardless of whether or not the initial failure was retryable. Now that these error events are being handled properly, this is no longer the case.

imjoehaines (approved):

PR #2034, opened 2023-10-27
Updated Next.js to v13

toniewypada:
Update Next.JS app to v13
Tested all scenarios which should work in local environment, all worked fine.

## Changeset

Dependencies are updated to the current version 

## Testing

Test scenarios are included in Next.JS example readme

### DIFF
--- a/examples/js/nextjs/package-lock.json
+++ b/examples/js/nextjs/package-lock.json
@@ -10,9 +10,10 @@
       "dependencies": {
         "@bugsnag/js": "^7.9.2",
         "@bugsnag/plugin-aws-lambda": "^7.9.2",
-        "next": "^10.0.8 || 11.x",
-        "react": "^17.0.2",
-        "react-dom": "^17.0.2"
+        "eslint-config-next": "^13.4.4",
+        "next": "^13.4.4",
+        "react": "^18.2.0",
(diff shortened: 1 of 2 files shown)


mclack:
I've tried all scenarios with the update, and noticed some unexpected behaviour. The [README](https://github.com/bugsnag/bugsnag-js/tree/next/examples/js/nextjs#:~:text=API%20scenario%202,scenario%201%20fails) says we expect API scenario 2 and Client scenario 1 to fail, but I saw [Client 1 works](https://app.bugsnag.com/bugsnag-test-projects/milo-react/errors/653bb59432020c0007c71e6f?filters[error.status]=open&filters[event.since]=30d&event_id=653bb68000c7207da1c20000) and [API 2 works](https://app.bugsnag.com/bugsnag-test-projects/milo-react/errors/653bb3afd5236800099ecc35?filters[error.status]=open&filters[event.since]=30d&event_id=653bb3af00c72254c0d90000). However, I was unable to get anything for API scenario 1.

I think it might be worth quickly checking all scenarios again with Next.js v13.5.4, and making sure the behaviour you see matches up with what we say in the README. Is this a direct result of changing the version of Next, or do you see the same behaviour in the old example app?

PR #2036, opened 2023-11-02
bump peer dependencies for v8

djskinner:
When lerna create a new version it does not update peer dependencies. So this PR does it manually.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -30,7 +30,6 @@
   "author": "Bugsnag",
   "license": "MIT",
   "devDependencies": {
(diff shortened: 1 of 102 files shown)


packages/browser/package.json
```diff
@@ -30,7 +30,6 @@
   "author": "Bugsnag",
   "license": "MIT",
   "devDependencies": {
-    "@bugsnag/core": "^7.0.1",
```
djskinner, line 33: already listed in `dependencies` so shouldn't be needed here

gingerbenw (approved):

PR #2040, opened 2023-11-15
fix(electron): Do not sync to NativeClient when `autoDetectErrors` or `nativeCrashes` are disabled

yousif-bugsnag:
## Goal

If either of `enabledErrorTypes.nativeCrashes` or `autoDetectErrors` are disabled, we don't install the sync layer (`NativeClient.install`) but still call all the methods to sync state to the native client. 

Currently this logs an error message: `Sync layer is not installed, first call install()`, which isn't very useful as users can't call `NativeClient.install`

This PR updates each of the plugins that sync to the native layer to ensure that they do not sync to the native layer if those config options are disabled. This affects:
- `plugin-electron-client-state-persistence`
- `plugin-electron-app`
- `plugin-electron-device`
- `plugin-electron-session`

## Testing

Added unit tests to assert that the native sync methods are not called when the relevant config options are disabled.

Also updated the `makeClientForPlugin` function in `electron-test-helpers` to support creating a client with an array of plugins - this allows the `client-state-persistence` unit tests to make use of the test helper since the plugin depends on the `client-state-manager` plugin being loaded.

### DIFF
--- a/packages/electron-test-helpers/src/client.ts
+++ b/packages/electron-test-helpers/src/client.ts
@@ -11,8 +11,8 @@ interface ClientTestHelpers {
 export function makeClientForPlugin ({
   config = {},
   schema = {},
-  plugin = undefined
-}: { config?: object, schema?: object, plugin?: Plugin } = {}): ClientTestHelpers {
+  plugins = []
+}: { config?: object, schema?: object, plugins?: Plugin[] } = {}): ClientTestHelpers {
   const client = new Client(
     {
       apiKey: 'abcabcabcabcabcabcabc1234567890f',
@@ -20,7 +20,7 @@ export function makeClientForPlugin ({
       ...config
     },
     { ...defaultSchema, ...schema },
-    plugin !== undefined ? [plugin] : []
+    plugins
   )
 
   let lastSession: SessionPayload
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,12 @@
 
 ### Changed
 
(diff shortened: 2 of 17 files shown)


imjoehaines (commented):
I think it might be safer to check for `enabledErrorTypes.nativeCrashes` before trying to sync to the native client rather than removing this error

packages/electron-test-helpers/src/client.ts
```diff
@@ -11,16 +11,16 @@ interface ClientTestHelpers {
 export function makeClientForPlugin ({
   config = {},
   schema = {},
-  plugin = undefined
-}: { config?: object, schema?: object, plugin?: Plugin } = {}): ClientTestHelpers {
+  plugins = undefined
```
gingerbenw, line 14: could we just default this to an empty array and remove the logic on line 23?

gingerbenw (approved):
couple of small comments, but approved based on unit tests as I don't understand enough about `Session.prototype._track` or `client._sessionDelegate` and why you would use one over the other.

packages/electron-test-helpers/src/client.ts
yousif-bugsnag: yes we could! updated 👍🏾

yousif-bugsnag:
> couple of small comments, but approved based on unit tests as I don't understand enough about `Session.prototype._track` or `client._sessionDelegate` and why you would use one over the other.

`Session.prototype._track` just updates the handled/unhandled event count for the session, whereas `client._sessionDelegate` is responsible for starting/pausing/resuming sessions

PR #2065, opened 2024-01-09
Set appActivity and appPackage Appium capabilities to avoid RN test flakes 

twometresteve:
## Goal

Set appActivity and appPackage Appium capabilities, using Maze Runner options, to avoid test flakes when trying to start the Android React Native test fixtures.

## Design

The main change here should resolve a common test failure when starting Android apps with Appium 2.  After making the initial change, there was some confusion resulting from each test fixture having a difference package name.  So I've also taken the opportunity to make them all consistent, as `com.reactnative`.

## Changeset

A number of files in the diff are showing as deleted and created, but all I've done is change their package, resulting in a file move that GitHub somehow hasn't recognised.

## Testing

Covered by a full CI run.


Cawllec (approved):
LGTM

PR #2073, opened 2024-01-24
[PLAT-11399] Bugsnag CLI iOS x React Native CLI

joshedney:
## Goal

Update the React Native CLI  to install BugSnag CLI instead of the Source Maps package for the iOS/Xcode Integration. 

If we detect that RN version is >= 0.69.0 we will set up the `.xcode.env` file to output the file source map files to our happy path `ios/build/sourcemaps/`.

If we detect that the RN version is < 0.69.0 we set up the extra arguments within the xcode build phase due to it not supporting the `.xcode.env` file.

Both options ensure the source map file is located where our `bugsnag-cli` expects it. 

Add all 5 upload tasks and the create build task to the `package.json`

- bugsnag:create-build        - Creates a new build
- bugsnag:upload-android      - Uploads Android source maps
- bugsnag:upload-rn-android   - Uploads React Native Android source maps
- bugsnag:upload-dsym         - Uploads iOS dSYMs
- bugsnag:upload-rn-ios       - Uploads React Native iOS source maps
- bugsnag:upload              - Runs all of the above tasks

## Changeset

Update the `AutomateSymbolicationCommand.ts` to install the BugSnag CLI package for the iOS integration.

## Testing

Covered by CI


packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
```diff
@@ -26,45 +26,25 @@ const HERMES_INSTRUCTIONS = `You are running a version of React Native that we c
 
 `
 
-const BUGSNAG_CLI_INSTRUCTIONS = `bugsnag:upload-android task added to your package.json - run this task to upload Android source maps after a build.
+const BUGSNAG_CLI_INSTRUCTIONS = `The following tasks have been added to your package.json and can be run after a build to upload source maps to BugSnag:
+
+    bugsnag:upload-android
+    bugsnag:upload-ios
```
tomlongridge, line 32: Needs an update with the other commands - maybe also a few words against each to explain them?

packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
```diff
@@ -79,23 +59,33 @@ export default async function run (projectRoot: string, urls: OnPremiseUrls): Pr
...
+        logger.info('Modifying the Xcode project')
+        await updateXcodeProject(projectRoot, urls[UrlType.UPLOAD], reactNativeVersion as string, logger)
+
+        await prompts({
+          type: 'text',
+          name: 'dsymUploadInstructions',
+          message: DSYM_INSTRUCTIONS,
+          initial: 'Hit enter to continue …'
+        }, { onCancel })
+      }
```
tomlongridge, line 88: I think we agreed that we only want to keep the part that enables source maps and we're not going to add the call to upload to the build - to keep it consistent and understandable.

We probably still want a prompt here that we're about to modify the Xcode project, but not for "automatic upload" and this functionality can be taken out of the Xcode file.

packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
```diff
@@ -171,21 +137,31 @@ async function writeToPackageJson (packageJsonPath: string, uploadUrl?: string,
...
     packageJson.scripts = {
       ...packageJson.scripts,
-      'bugsnag:create-build': buildCommand,
-      'bugsnag:upload-android': uploadCommand
+      'bugsnag:create-build': createBuildCommand,
+      'bugsnag:upload-android': androidUploadCommand,
+      'bugsnag:upload-rn-android': rnAndroidUploadCommand,
+      'bugsnag:upload-dsym': dsymUploadCommand,
+      'bugsnag:upload-rn-ios': rnIosUploadCommand,
+      'bugsnag:upload': 'npm run bugsnag:upload-android && npm run bugsnag:upload-rn-android && npm run bugsnag:upload-rn-ios && npm run bugsnag:upload-dsym'
```
tomlongridge, line 164: As discussed IRL, let's call the CLI directly here to avoid yarn compatibility issues.

packages/react-native-cli/src/commands/AutomateSymbolicationCommand.ts
joshedney: Bundled as part of this is setting where the sourcemaps are exported (our happy path) did we want a prompt for that instead or just update with out asking?

tomlongridge (approved):

PR #2083, opened 2024-02-09
[PLAT-11445] Fix react navigation test flakes 

gingerbenw:
## Goal

Ensure `getPlugin` is not called before bugsnag has started

## Changeset

Updated test fixture to await `Bugsnag.isStarted()` to be true before calling `getPlugin`


lemnik (approved):
LGTM

twometresteve (approved):

PR #2094, opened 2024-03-08
Improved performance of click event breadcrumbs

AnastasiiaSvietlova:
## Goal

Reduce time it takes gather innerText of a large element

## Changeset

Updated RegExp for trimming whitespaces in breadcrumbs for UI click

## Testing

Covered by existing unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## TBD
-
-### Changed
-
-- (plugin-interaction-breadcrumbs) Improved performance of click event breadcrumbs [#2094](https://github.com/bugsnag/bugsnag-js/pull/2094)
-
 ## v7.22.6 (2024-03-05)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


packages/plugin-interaction-breadcrumbs/interaction-breadcrumbs.js
```diff
@@ -21,8 +21,7 @@ module.exports = (win = window) => ({
   }
 })
 
-const trimStart = /^\s+/
-const trimEnd = /(^|[^\s])\s+$/
+const trim = /^\s*([^\s].{0,139}[^\s])?\s*/s
```
gingerbenw, line 24: @lemnik looks like we've got an issue with this regex - perhaps the `s` at the end? it was modified slightly from the suggestion as it didn't remove the trailing whitespace.

lemnik: Yeah, that's certainly what the error indicated, although the `/s` flag is the [dotAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) flag. Maybe it's a version compatibility thing and we need to use `[.\n]` instead of `.` (assuming `dotAll` is not available on the target browser)?

AnastasiiaSvietlova: I replaced `.` with `[.\n]` and tests failed locally.
 For example:
Screenshot 2024-03-11 at 16 32 24

CHANGELOG.md
tomlongridge, line 1: Looks like your linter has applied some formatting to older entries - can you revert so that only the new entry is added?

packages/plugin-interaction-breadcrumbs/interaction-breadcrumbs.js
AnastasiiaSvietlova: It worked! Thanks!

gingerbenw (approved):

PR #2103, opened 2024-03-27
v8: fix crash when calling Bugsnag methods before start

djskinner:


packages/node/src/notifier.js
```diff
@@ -79,7 +79,7 @@ Object.keys(Client.prototype).forEach((m) => {
   if (/^_/.test(m)) return
   Bugsnag[m] = function () {
     // if we are in an async context, use the client from that context
-    const client = Bugsnag._client._clientContext.getStore() || Bugsnag._client
+    const client = Bugsnag._client?._clientContext?.getStore() || Bugsnag._client
```
gingerbenw, line 82: if we're still supporting IE11 then we'll have to drop the optional chaining I'm afraid

@tomlongridge does v8 drop IE11 support?

djskinner: this code only runs on node

djskinner: though the node e2e tests are failing so maybe it was only added in node 20. will revert to the tried and tested ways then

gingerbenw (approved):
seems good to me - but not sure why these lambda tests are failing? 🤔

djskinner:
> but not sure why these lambda tests are failing? 🤔

Yeah, me neither, but it's not related to this PR as it can be seen on the integration/v8 branch itself. I am looking into this today.

PR #2106, opened 2024-03-28
add node 20 to e2e test pipeline

djskinner:


test/node/features/fixtures/restify/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "bugsnag-test",
   "dependencies": {
-    "restify": "^7.2.1",
+    "restify": "^11.1.0",
```
djskinner, line 4: works on node 18 and 20 but no longer on node 12

test/node/features/fixtures/restify/scenarios/app-disabled.js
```diff
@@ -29,15 +29,15 @@ server.use(function (req, res, next) {
   next()
 })
 
-server.get('/', function (req, res) {
+server.get('/', function (req, res, next) {
```
djskinner, line 32: Fix these runtime errors:

```
AssertionError [ERR_ASSERTION]: Handler [handler-1 on GET /sync/:message] is missing a third argument (the "next" callback) but is not an async function. Middleware handlers can be either async/await or callback-based.Callback-based (non-async) handlers should accept three arguments: (req, res, next). Async handler functions should accept maximum of 2 arguments: (req, res).
```

test/node/features/restify.feature
```diff
@@ -1,4 +1,4 @@
-@skip_node_18
+@skip_node_12
```
djskinner, line 1: new restify version doesn't work on 12 because of a syntax error:

```
/Users/dan.skinner/bugsnag-js/test/node/features/fixtures/restify/node_modules/sonic-boom/index.js:393
    cb?.()
```

test/node/features/fixtures/webpack/webpack.config.js
```diff
@@ -8,7 +8,13 @@ module.exports = {
...
     filename: 'index.bundle.js',
     library: '',
-    libraryTarget: 'commonjs'
+    libraryTarget: 'commonjs',
+  },
+  // The version of terser shipping with webpack 4 does not work in node 20
+  // without the `--openssl-legacy-provider` flag. So we disable minification
+  // in order to continue being able to test webpack 4 on node 20.
+  optimization: {
+    minimize: false,
```
djskinner, line 17: any version of webpack 4 on node 20 fails with `Error message "error:0308010C:digital envelope routines::unsupported"` to do with changes to lib ssl version being used by an older version of the terser plugin that comes with webpack. To work around this just disable minification as it's not necessary for our testing

test/node/features/restify.feature
```diff
@@ -1,4 +1,4 @@
-@skip_node_18
```
djskinner, line 1: 18 was failing for the same reason as 20 (restify needed bumping) and it had been disabled a while back with the intention of coming back and fixing it, but that was forgotten about until now

gingerbenw (approved):
looks good to me - while we're skipping node 12 in a few places, is this a good time to raise whether we actually should still support node 12 @tomlongridge ? I believe it went out of support in April 2022

PR #2107, opened 2024-04-03
automatic console breadcrumbs for node (bugsnag-js v8)

djskinner:
## Goal

- Support for breadcrumbs on bugsnag-js v8 is already merged into the integration branch
- Adding automatic breadcrumbs on console calls makes sense to improve error debugging capabilities

## Design

- Re-use the existing plugin to benefit from DRY.
- Call the `leaveBreadcrumb` method on the client clone stored in the async context, if it exists
- The downside to this approach is a slight increase in the browser bundle size

## Changeset

- plugin-console-breadcrumbs included in the node notifier
- behaviour of plugin-console-breadcrumbs in unchanged in the browser

## Testing

- [x] unit tests
- [x] e2e tests
- [x] manual test against a test project using an alpha release

Screenshot 2024-04-03 at 17 41 07


test/node/features/fixtures/contextualize/scenarios/contextualize.js
```diff
@@ -11,6 +11,7 @@ Bugsnag.start({
 var contextualize = Bugsnag.getPlugin('contextualize')
 
 contextualize(function () {
+  console.log('manual notify')
```
gingerbenw, line 14: I know it's only a test - but are these logs being used?

gingerbenw (approved):

packages/plugin-console-breadcrumbs/console-breadcrumbs.js
```diff
@@ -13,7 +13,9 @@ exports.load = (client) => {
   map(CONSOLE_LOG_METHODS, method => {
     const original = console[method]
     console[method] = (...args) => {
-      client.leaveBreadcrumb('Console output', reduce(args, (accum, arg, i) => {
+      // if we are in an async context, use the client from that context
+      const c = client._clientContext && client._clientContext.getStore() ? client._clientContext.getStore() : client
```
imjoehaines, line 17: Would this be something we have to do in every plugin used in the node notifier now? If so it seems like we should make this automatic so that the plugins themselves don't need to change, otherwise we're inviting bugs where things are using the wrong client

test/node/features/fixtures/contextualize/scenarios/contextualize.js
djskinner: yes, ensuring that a breadcrumb is created on console.log

djskinner: see `test/node/features/contextualize.feature`

packages/node/src/notifier.js
```diff
@@ -53,6 +53,22 @@ const Bugsnag = {
 
     const bugsnag = new Client(opts, schema, internalPlugins, { name, version, url })
 
+    Object.keys(Client.prototype).forEach((m) => {
+      if (/^_/.test(m)) return
```
imjoehaines, line 63: I'm not sure that this is right — wouldn't we want to proxy the call even if it's a "private" method?

packages/core/client.js
```diff
@@ -186,7 +186,6 @@ class Client {
     // stuff like __proto__ etc. (only store the result if the plugin had a
     // name)
     if (plugin.name) this._plugins[`~${plugin.name}~`] = result
-    return this
```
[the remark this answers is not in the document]
djskinner: It doesn't but I thought it worth removing since we're doing a major version change and it's not used

packages/node/src/notifier.js
djskinner: Yes, this needs to apply to all method calls so that it's consistently applied to the same client e.g. if a public method then calls a private method they need directing to the same place

packages/node/test/notifier.test.ts
```diff
@@ -309,4 +309,29 @@ describe('node notifier', () => {
...
+        const contextualize = Bugsnag.getPlugin('contextualize')
+
+        contextualize(() => {
+          console.log('test')
+          // @ts-ignore
+          expect(Bugsnag._client._clientContext.getStore()._breadcrumbs[0].message).toBe('Console output')
+        })
+
+        // @ts-ignore
+        expect(Bugsnag._client._breadcrumbs.length).toBe(0)
```
imjoehaines, line 333:
```suggestion
        expect(Bugsnag._client._breadcrumbs).toHaveLength(0)
```

imjoehaines (approved):

Issue #2101, opened 2024-03-25
build failed when enable react native new arch


iwater:
<img width="573" alt="截屏2024-03-25 10 32 12" src="https://github.com/bugsnag/bugsnag-js/assets/108937/a4773493-5e06-4e72-ad53-5fd2ccb29a64">

### Describe the bug
build failed when enable react native new arch

### Steps to reproduce
1. init a new react-native app
2. enable new arch
3. install @bugsnag/react-native
4. build & See error

### Environment
* Bugsnag version: 7.22.6
* ReactNative Version: 0.73.6

clr182:
Hi @iwater 

Thank you for reaching out. 

We not have an item on our backlog to investigate this further when priorities allow. Once we have more information we will be sure to reach out and share an update! 

Kind regards,
Christian

toniewypada:
Hello 👋🏻, 
We’ve released new version of our JS notifier, [v7.22.7](https://github.com/bugsnag/bugsnag-js/pull/2121), which contains fix to the build issue, you’ve reported. The update should move BugsnagReactNative from objective-c to objective c++ while building React Native app.
If the issue still persists after updating the React Native SDK, please let us know and we’ll investigate further.

PR #2113, opened 2024-04-08
[PLAT-11900] Move BugsnagReactNative from objective-c to objective c++

yousif-bugsnag:
## Goal

Fixes an iOS build error in React Native when New Architecture is enabled (reported in #2101).

A fix introduced in React Native 0.73.5 (and backported to 0.72.12) ensures that the `RCT_NEW_ARCH_ENABLED` flag is now passed down to libraries where it was not previously (https://github.com/facebook/react-native/pull/43088). This means that anything imports the generated spec header must be Objective-C++.

## Design

Moved `BugsnagReactNative` from Objective-C to Objective-C++.

## Testing

Upgraded the React Native 0.72 test fixture to 0.72.12 which contains the aforementioned fix.


lemnik (approved):
LGTM

PR #2143, opened 2024-05-17
[PLAT-12017] lodash.template | CVE-2021-23337

gingerbenw:
## Goal

Manually override `@lerna/conventional-commits` to remove vulnerable `lodash.template` package, resolving CVE-2021-23337

## Design

Ideally, we should update the major version of lerna to resolve this and additional security vulnerabilities, but this requires additional work around CI pipelines and is out of scope for this PR.

### DIFF
--- a/examples/electron/electron-bundled/package.json
+++ b/examples/electron/electron-bundled/package.json
@@ -27,7 +27,7 @@
     "@electron-forge/plugin-fuses": "^7.4.0",
     "@electron-forge/plugin-webpack": "^7.4.0",
     "@electron/fuses": "^1.8.0",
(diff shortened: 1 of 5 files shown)


djskinner (approved):
has this completely resolved the alert? I can't remember for sure but I thought CodeQL would report a dependabot alert as resolved back to the PR

gingerbenw:
> has this completely resolved the alert? I can't remember for sure but I thought CodeQL would report a dependabot alert as resolved back to the PR

I've never seen that before, actually! it might be possible to do with dependabot but I don't think we've got it configured that way

PR #2148, opened 2024-05-29
Discard duplicate JS exceptions in iOS New Architecture

yousif-bugsnag:
## Goal

In the Old Architecture React Native raises an Objective-C `RCTFatalException` when an unhandled JS exception occurs. These are discarded in the native notifier because BugSnag's JS layer also catches these JS exceptions.

However in 0.73+, with New Architecture enabled, these exceptions are caught and rethrown as a cpp exception, so the discard pattern no longer applies and a duplicate native event is reported.

This PR adds an additional `onSendError` block (when the New Architecture is enabled) to also discard these new exception types 

## Design

Because this new exception type `facebook::jsi::JSError` is also thrown for any exceptions that occur in Turbo Module methods, we're not able to discard unhandled JS exceptions using a `discardClasses` pattern.

Instead an onSendError callback is used to discard based on the error message. For unhandled JS exceptions the error message is always prepended with `Exception in HostFunction: Unhandled JS Exception`

## Changeset

Added a new `onSendError` block to the native config in `BugsnagReactNativePlugin.m` if the New Architecture is enabled.

The iOS New Architecture tests for 0.73 have been reinstated - as part of this all scenarios that throw native exceptions have been split into separate tests for old arch and new arch to account for the new exception handling behaviour in iOS Turbo Modules

## Testing

Covered by a full CI run

### DIFF
--- a/packages/react-native/ios/BugsnagReactNative/BugsnagReactNativePlugin.m
+++ b/packages/react-native/ios/BugsnagReactNative/BugsnagReactNativePlugin.m
@@ -12,7 +12,7 @@
 
 - (void)load:(BugsnagClient *)client {
     //
-    // React Native catches JS exceptions and calls RCTFatal() to raise an Objective-C exception in response.
+    // When using the old architecture, React Native catches JS exceptions and calls RCTFatal() to raise an Objective-C exception in response.
     // These need to be ignored because Bugsnag's JS layer also catches JS exceptions, via a different mechanism.
     //
     // RCTFatal() sets the exception name to "RCTFatalException: ${error.localizedDescription}"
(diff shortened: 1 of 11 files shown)


packages/react-native/ios/BugsnagReactNative/BugsnagReactNativePlugin.m
```diff
@@ -12,7 +12,7 @@ @implementation BugsnagReactNativePlugin
 
 - (void)load:(BugsnagClient *)client {
     //
-    // React Native catches JS exceptions and calls RCTFatal() to raise an Objective-C exception in response.
+    // In Old Architecture, React Native catches JS exceptions and calls RCTFatal() to raise an Objective-C exception in response.
```
gingerbenw, line 15:
```suggestion
    // When using the old architecture, React Native catches JS exceptions and calls RCTFatal() to raise an Objective-C exception in response.
```

packages/react-native/ios/BugsnagReactNative/BugsnagReactNativePlugin.m
```diff
@@ -29,6 +29,19 @@ - (void)load:(BugsnagClient *)client {
     NSMutableSet *discardClasses = [client.configuration.discardClasses mutableCopy] ?: [NSMutableSet set];
     [discardClasses addObject:[NSRegularExpression regularExpressionWithPattern:discardPattern options:0 error:nil]];
     client.configuration.discardClasses = discardClasses;
+
+    // In the New Architecture on 0.73+, these exceptions are caught and rethrown as facebook::jsi::JSError exceptions,
```
gingerbenw, line 33:
```suggestion
    // When using new architecture on 0.73+, these exceptions are caught and re-thrown as facebook::jsi::JSError exceptions,
```

gingerbenw (approved):
Looks good to me - can I also say the PR description is great, loads of detail! 👍🏻

PR #2152, opened 2024-06-05
Use synchronous native module calls when New Architecture is enabled

yousif-bugsnag:
## Goal

In RN 0.74 New Architecture (bridgeless mode), once an exception is thrown, async promises are no longer being resolved. This means JS exceptions are not being reported, because post-error communication with the native layer to capture native state and send an exception for delivery is not able to run.

This PR fixes the issue by making calls to the native layer synchronous when the New Architecture is enabled.

As part of this, RN 0.74 has been added to the test matrix and the e2e tests have been updated accordingly.

#### Synchronous native methods

There are 2 explicitly async (i.e. promise-returning) methods involved in post-error communication: 
- `getPayloadInfo` (used in `plugin-react-native-event-sync`)
- `dispatch` (used in `delivery-react-native`)

These have been converted to synchronous methods and separate async versions `getPayloadInfoAsync` and `dispatchAsync` added. Both packages have been updated to use the synchronous methods when the New Arch is enabled.

For the remaining native module methods, a preprocessor directive is used on iOS to export those methods as synchronous when the new architecture is enabled (on iOS, `void` native methods that don't return a value in JS are also asynchronous unless exported on the native side with `RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD`)

## Testing

React Native 0.74 has now been added to the test matrix. A new step definition has been added to support testing payload values for a given architecture (old vs new) and RN version combination.

### DIFF
--- a/test/react-native/features/steps/react-native-steps.rb
+++ b/test/react-native/features/steps/react-native-steps.rb
@@ -149,3 +149,47 @@ Then('the stacktrace contains {string} equal to {string}') do |field_path, expec
   end
   fail("No field_path #{field_path} found with value #{expected_value}") unless found
 end
+
+# Tests that the given payload value is correct for the target arch and RN version combination.
+# This step will assume the expected and payload values are strings.
+#
+# The DataTable used for this step should have the headers `arch` `version` and `value`
+# The `arch` column must contain either `old` or `new`
+# The `version` column should contain the version of the app to test the value against, or `default` for unspecified versions
+# The `value` column should contain the expected value for the given arch and version combination
+#
+#   | arch | version | value                      |
+#   | new  | 0.74    | Error                      |
+#   | new  | default | java.lang.RuntimeException |
+#   | old  | default | java.lang.RuntimeException |
+#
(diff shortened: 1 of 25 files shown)


test/react-native/features/steps/react-native-steps.rb
```diff
@@ -149,3 +149,47 @@ def get_app_state
...
+  if version_values.empty?
+    version_values = arch_values.select { |hash| hash['version'] == 'default' }
+  end
+
+  raise("There is no expected value for the current version \"#{current_version}\"") if version_values.empty?
+  raise("Multiple expected values found for arch \"#{arch}\" and version \"#{current_version}\"") if version_values.length() > 1
+
+  expected_value = version_values[0]['value']
+  assert_equal_with_nullability(expected_value, payload_value)
+end
```
gingerbenw, line 195: 👌🏻

gingerbenw (approved):
seems good to me from the js/feature side of things, I can't say I fully understand the java/impl functions, but they seem clear enough

lemnik (approved):
LGTM

PR #2153, opened 2024-06-05
ci(react-native): remove unused test fixture and docker images

yousif-bugsnag:
## Goal

- Remove the unused static test fixture for 0.73 (this is now generated dynamically)
- Remove the unused java 17 android builder docker images (we now use the macos boxes for these builds)


twometresteve (approved):
LGTM pending a full CI run

PR #2154, opened 2024-06-06
Don't start sessions when ReactNative apps are loaded in the background

lemnik:
## Goal
Don't start/resume sessions in ReactNative apps when the bundle is pre-loaded in the background (leaving it to the native layer to start the session when appropriate).

## Changeset
Replace the `resumeSession` call in the ReactNative `Bugsnag.start` implementation with a new `NativeClient.resumeSessionOnStartup` call. The Android native layer first checks that the app is in the foreground before calling `resumeSession` as normal.

## Testing
Manually tested using system broadcasts.


yousif-bugsnag (approved):
looks good to me

tomlongridge (approved):

PR #2159, opened 2024-06-14
[PLAT-12224] Add setTraceCorrelation method to event

gingerbenw:
## Goal

Facilitate trace correlation with the performance client

## Design

Add new method to events to enable adding span and trace id's, which can be leveraged by `bugsnag-js-performance` to correlate traces and errors, but can also be called manually if required

## Changeset

Add a new `setTraceCorrelation` method to events

## Testing

Add unit test that calls setTraceCorrelation and verifies the span and trace id are present in the serialised event


yousif-bugsnag (approved):

packages/core/test/event.test.ts
```diff
@@ -420,4 +420,13 @@ describe('@bugsnag/core/event', () => {
         }))
     })
   })
+
+  describe('event.setTraceCorrelation()', () => {
+    it('allows setting the span and trace id', () => {
```
imjoehaines, line 425: There's a couple of scenarios missing tests:

- if `spanId` is `undefined` or `null` it's still added
- if `traceId` is `undefined` or `null` it's not added

gingerbenw: Should we also update the api for `setTraceCorrelation` to accept `undefined | null` for the spanId?

gingerbenw: could they _both_ be expected to be null? 🤔

imjoehaines: > could they both be expected to be null? 🤔

Presumably we shouldn't call `setTraceCorrelation` if there's no trace ID so I think only span ID should be optional in the API

I don't think it makes sense for span ID to be `null` as an expected value (i.e. in the TS type definition) but it's javascript so it can easily end up that way at runtime

so I think the signature should be:

``ts
setTraceCorrelation(traceId: string, spanId?: string): void
```

but in the method we can't assume they have the right types

imjoehaines (approved):

PR #2164, opened 2024-06-25
Add appium version definitions on all android bitbar test runs

Cawllec:
## Goal

Helps prevent potential issues in appium 2.0 for our tests by setting appium to 1.22 on all of our android related test runs.


twometresteve (approved):
Looks good but we should probably trigger a full build.

Issue #2150, opened 2024-06-04
@bugsnag/plugin-vue is throwing: TypeError: Cannot read properties of undefined (reading 'errorHandler')


brunomonteiro-asknicely:
### Describe the bug
The `bugsnag-vue.js` file is using Vue 2 syntax in a Vue 3 project due to a problematic if check. This then causes Bugsnag to try to access `Vue.config` in a Vue 3 project where global config is not available anymore.

This issue was also reported [here](https://github.com/bugsnag/bugsnag-js/issues/1754), but it was closed due to lack of activity.

Problematic line seems to be [this one](https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-vue/src/index.js#L19). I believe a simple update to check `Vue.version` should fix it.

### Steps to reproduce
1. Install Vue 3 via `package.json`
2. Import Vue 3 [via CDN](https://vuejs.org/guide/quick-start.html#using-vue-from-cdn) in the `html` file
3. Follow the [instructions](https://docs.bugsnag.com/platforms/javascript/vue/) to install Bugsnag in a Vue app

### Environment
* Bugsnag version: 7.23.0
* Browser framework version (if any): Vue 3

### Example Repo

- [X] Create a minimal repository that can reproduce the issue
- [X] Link to it here: https://codesandbox.io/p/devbox/xenodochial-bose-qvvrvp

<details><summary>Error messages:</summary>

```
TypeError: Cannot read properties of undefined (reading 'errorHandler')
    at _$vue2_3 (bugsnag-vue.js:82:1)
    at BugsnagPluginVue.load (bugsnag-vue.js:141:1)
    at Client._loadPlugin (bugsnag.js:1704:1)
    at eval (bugsnag.js:1592:1)
    at eval (bugsnag.js:338:1)
    at _$reduce_17 (bugsnag.js:7:1)
    at _$map_16 (bugsnag.js:337:1)
    at new Client (bugsnag.js:1591:1)
    at Object.createClient (bugsnag.js:3268:1)
    at Object.start (bugsnag.js:3285:1)
```
</details>

brunomonteiro-asknicely:
Just as an update: if I go to my `node_modules` folder and manually change the code to
```javascript
if (this.Vue.version[0] === '2') {
```
it works fine!

clr182:
Hi @brunomonteiro-asknicely 

Thanks for sharing this workaround. 

[...]

PR #2171, opened 2024-07-12
Check global vue config exists before installing vue 2 handler

yousif-bugsnag:
## Goal

Fixes an error in the Vue plugin initialisation when using Vue 3+ via CDN (reported in #2150)

## Design

Currently the plugin attempts to install the Vue 2 handler (via the global Vue.config) if a global Vue object exists, but CDN builds of Vue 3+ will also expose a global Vue object.

This PR updates the check in plugin load to ensure that Vue.config exists before installing the Vue 2 handler.

## Testing

Added a new unit test and tested manually with a Vue 3 CDN build.

### DIFF
--- a/packages/plugin-vue/src/index.js
+++ b/packages/plugin-vue/src/index.js
@@ -16,7 +16,7 @@ module.exports = class BugsnagPluginVue {
   }
 
   load (client) {
-    if (this.Vue) {
+    if (this.Vue && this.Vue.config) {
       installVue2(this.Vue, client)
       return {
         installVueErrorHandler: () => client._logger.warn('installVueErrorHandler() was called unnecessarily')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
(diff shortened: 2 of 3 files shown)


packages/plugin-vue/src/index.js
```diff
@@ -16,7 +16,7 @@ module.exports = class BugsnagPluginVue {
   }
 
   load (client) {
-    if (this.Vue) {
+    if (this.Vue && this.Vue.config) {
```
gingerbenw, line 19: any particular reason not to check against the vue version? we could even use a switch statement to make supporting new versions easier

yousif-bugsnag: Well the global Vue.config was removed in v3, but I just thought it made more sense to check if the required API exists rather than check for a particular version.

Really either way would work - and I guess the plugin is already 'versioned' so happy to change to a version check if you think a  would be better.

gingerbenw: fair point. API over version. 👍🏻

gingerbenw (approved):

PR #2174, opened 2024-07-22
[PLAT-12474] Support error correlation property in events

gingerbenw:
## Goal

Support error correlation when using React Native

## Changeset

Add `correlation` property to event payload

## Testing

Update unit tests to assert on correlation property

### DIFF
--- a/packages/delivery-react-native/delivery.js
+++ b/packages/delivery-react-native/delivery.js
@@ -29,7 +29,8 @@ module.exports = (client, NativeClient) => ({
       groupingHash: event.groupingHash,
       apiKey: event.apiKey,
       featureFlags: event.toJSON().featureFlags,
-      nativeStack: nativeStack
+      nativeStack: nativeStack,
+      correlation: event.toJSON().correlation
     }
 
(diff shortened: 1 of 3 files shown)


yousif-bugsnag (approved):

packages/delivery-react-native/delivery.js
```diff
@@ -29,7 +29,8 @@ module.exports = (client, NativeClient) => ({
       groupingHash: event.groupingHash,
       apiKey: event.apiKey,
       featureFlags: event.toJSON().featureFlags,
-      nativeStack: nativeStack
+      nativeStack: nativeStack,
+      correlation: event.toJSON().correlation
```
imjoehaines, line 33: Do we need to call `toJSON` here? It looks like we could read directly from `event._correlation` instead

gingerbenw: I had started with that, but for some reason that was always `undefined` in my initial testing, but I may have conflated it with another issue I was seeing. Unit tests still passing with `event._correlation` so that's what we'll do :)

imjoehaines (approved):

PR #2184, opened 2024-08-19
fix types

mifi:
## Goal

if not i'm getting typescript errors when importing

## Design

seems to be the way it's done

## Changeset

- change export to include types, so it works with bundler

## Testing

tested in a project

### DIFF
--- a/packages/electron/package.json
+++ b/packages/electron/package.json
@@ -3,7 +3,11 @@
   "version": "7.25.0",
   "main": "src/notifier.js",
   "exports": {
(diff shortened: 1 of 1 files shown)


yousif-bugsnag:
Hi @mifi, thanks for raising this.

This has highlighted some wider issues with the separate main and renderer exports for this package, particularly for projects using ESM.

We have some upcoming tasks scheduled to re-work our JS packages and build processes that should resolve these, so we're going to hold off on this PR in order to address these issues together.

PR #2196, opened 2024-09-13
ci(react-native): fix react-navigation 0.69 test fixture builds

yousif-bugsnag:
## Goal

Fixes the react-navigation 0.69 test fixture builds in CI

The test fixture was configured to install the latest v2 version of react-native-gesture-handler, and it looks like something in the latest release (which coincides with when these builds began failing) has broken compatibility with RN 0.69.

Updated the build script to pin the dependency to the last known working version.

## Testing

Covered by a full CI run


gingerbenw (approved):

gingerbenw:
👌🏻 excellent detective work!

PR #2202, opened 2024-09-19
Move react-native-navigation tests to dynamic fixture setup

yousif-bugsnag:
## Goal

Converts the e2e tests for `@bugsnag/plugin-react-native-navigation` to the dynamic fixture setup and adds e2e tests for React Native 0.71 and 0.72.

## Design

This requires a separate test fixture, which is generated dynamically if the `REACT_NATIVE_NAVIGATION` environment variable is set.

The scenarios and feature files have also been separated as previously these used the same naming as the react-navigation scenarios.

Due to some issues in react-native-navigation itself, these tests are currently skipped for:
- React Native versions 0.73+ (not yet supported by `rnn-link`)
- iOS New Architecture (react-native-navigation fails to build on iOS when New Architecture is enabled)

The legacy static e2e tests for react-navigation and react-native-navigation have been removed.

## Testing

Covered by a full CI run


.buildkite/full/react-native-ios-pipeline.full.yml
```diff
@@ -340,48 +352,75 @@ steps:
...
-              - --farm=bs
-              - --device=IOS_16
-              - --a11y-locator
-              - --fail-fast
-              - features/navigation.feature
-        concurrency: 5
-        concurrency_group: "browserstack-app"
-        concurrency_method: eager
+      # Skipped due to build failures with react-native-navigation when new arch is enabled
+      # TODO: Investigate and try to re-enable when we add tests for more recent React Native versions
```
gingerbenw, line 393: prefer raising a ticket and adding `skip: "See PLAT-1234"` than commenting out

gingerbenw (approved):

.buildkite/full/react-native-ios-pipeline.full.yml
yousif-bugsnag: Discussed this with @tomlongridge and we decided not to raise a ticket for this because it's not something we can fix on our side - instead we'll re-assess when we come to add the next RN version to the test matrix.

PR #2213, opened 2024-10-02
Add buildkite analytics test collection to react and react-native e2e tests

Cawllec:
There's some overlap with maze-runner options between config files and options in the pipeline, once the maze-runner mechanism for reading the config files is fixed we should be able to remove the pipeline options.

### DIFF
--- a/test/react-native/features/support/maze.all.cfg
+++ b/test/react-native/features/support/maze.all.cfg
@@ -1,3 +1,1 @@
---no-tunnel
---aws-public-ip
 --a11y-locator
--- a/test/react-native/features/support/maze.all.cfg
+++ b/test/react-native/features/support/maze.all.cfg
@@ -1,0 +1,3 @@
+--no-tunnel
+--aws-public-ip
+--a11y-locator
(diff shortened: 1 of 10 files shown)


test/react-native/features/support/maze.all.cfg
```diff
@@ -0,0 +1,3 @@
+--no-tunnel
+--aws-public-ip
```
twometresteve, line 2: These two should only be in place on Buildkite (BrowserStack will be used locally with the secure tunnel).

twometresteve (approved):
One small tweak to make before merging, but happy to approve now.

test/react-native/features/support/maze.all.cfg
Cawllec: True, but all the React-native tests use Buildkite AFAICT

twometresteve: In principle engineers could be running tests on their local machine when writing new tests, even if it's awkward to build the test fixture.

PR #2221, opened 2024-10-15
add sendPayloadChecksums config option and implement Bugsnag-Integrity header

djskinner:
## Goal

Add Bugsnag-Integrity request header (where required APIs are available) and implement the `sendPayloadChecksums` core config option to allow opting out of this behavior.

Bugsnag-Integrity headers are set by default unless the `endpoints` configuration option is set, in which case they are disabled. This behavior can be overriden with the new `sendPayloadChecksums` config option.

## Design

- Bugsnag-Integrity headers are set where the environment supports it (requires native promises, subtle crypto, and a secure context)
- The behaviour of `sendPayloadChecksums` was chosen to avoid possible CORS issues where custom endpoints are being used and they are not configured to accept the new `Bugsnag-Integrity` in requests.

## Changeset

- Add `sendPayloadChecksums` core config option
- Set `Bugsnag-Integrity` in delivery-fetch, which is used in `@bugsnag/web-worker`.
- Set `Bugsnag-Integrity` in delivery-xml-http-request, which is used in `@bugsnag/browser`.

## Testing

- unit tests on delivery-fetch and delivery-xml-http-request
- manually tested using the web-worker example: event appears when integrity header is correct, event does not appear in the dashboard when the integrity header is wrong
- integration tests for browser and web worker
- e2e tests for browser and web worker


jest/FixJSDOMEnvironment.js
```diff
@@ -0,0 +1,18 @@
...
+  constructor (...args) {
+    super(...args)
+
+    this.global.TextEncoder = TextEncoder
+    this.global.TextDecoder = TextDecoder
+    this.global.crypto = {
+      subtle: crypto.webcrypto.subtle
+    }
+  }
+}
```
djskinner, line 16: More info: https://mswjs.io/docs/migrations/1.x-to-2.x/#requestresponsetextencoder-is-not-defined-jest

We could also use https://github.com/mswjs/jest-fixed-jsdom but would still need to manually add the crypto APIs

I think this is fairly simple and clear

packages/delivery-fetch/delivery.js
```diff
@@ -1,19 +1,37 @@
...
 
-const delivery = (client, fetch = global.fetch) => ({
+async function addIntegrityHeader (windowOrWorkerGlobalScope, requestBody, headers) {
+  if (windowOrWorkerGlobalScope.isSecureContext && windowOrWorkerGlobalScope.crypto && windowOrWorkerGlobalScope.crypto.subtle && windowOrWorkerGlobalScope.crypto.subtle.digest && typeof TextEncoder === 'function') {
+    const msgUint8 = new TextEncoder().encode(requestBody)
+    const hashBuffer = await windowOrWorkerGlobalScope.crypto.subtle.digest('SHA-1', msgUint8)
+    const hashArray = Array.from(new Uint8Array(hashBuffer))
+    const hashHex = hashArray
+      .map((b) => b.toString(16).padStart(2, '0'))
+      .join('')
```
djskinner, line 10: See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string

packages/web-worker/src/notifier.js
```diff
@@ -36,7 +36,7 @@ export const Bugsnag = {
     // configure a client with user supplied options
     const bugsnag = new Client(opts, schema, internalPlugins, { name, version, url })
 
-    bugsnag._setDelivery(delivery)
+    bugsnag._setDelivery(client => delivery(client, undefined, self))
```
djskinner, line 39: Same way the global `self` is passed to plugin, see above, e.g. `pluginWindowUnhandledRejection(self)`

jest/FixJSDOMEnvironment.js
djskinner: For future reference, when we come to upgrade jest, it will need rewriting like this:

```
const { TextDecoder, TextEncoder } = require('node:util')
const crypto = require('crypto')

const JSDOMEnvironment = require('jest-environment-jsdom').default

function FixJSDOMEnvironment (...args) {
  var _this = Reflect.construct(JSDOMEnvironment, args)

  _this.global.TextEncoder = TextEncoder
  _this.global.TextDecoder = TextDecoder

  Object.defineProperty(_this.global, 'crypto', {
    value: {
      subtle: crypto.webcrypto.subtle
    }
  })

  return _this
}

FixJSDOMEnvironment.prototype = Object.create(JSDOMEnvironment.prototype)
FixJSDOMEnvironment.prototype.constructor = FixJSDOMEnvironment

module.exports = FixJSDOMEnvironment
```

This is because `JSDOMEnvironment` is an ES6 class but we transpile classes in our code down to es5. This is how I managed to get it to work having an es5 class extend from an es6 class

packages/browser/test/index.test.ts
```diff
@@ -4,38 +4,55 @@ const DONE = window.XMLHttpRequest.DONE
```
djskinner, line 4: I reworked the mocking in this test suite to get a callback on the xhr send where I could call done. The existing mocking wasn't able to handle the awaiting of the getIntegrity promise.

packages/browser/test/index.test.ts
```diff
@@ -247,4 +264,96 @@ describe('browser notifier', () => {
...
+
+    afterEach(() => {
+      // @ts-ignore
+      window.isSecureContext = false
+    })
+
+    it('includes the integrity header by default', (done) => {
+      const onSessionSend = (session: MockXHR) => {
+        expect(session.open).toHaveBeenCalledWith('POST', 'https://sessions.bugsnag.com')
+        expect(session.setRequestHeader).toHaveBeenCalledWith('Bugsnag-Integrity', expect.any(String))
```
djskinner, line 282: I spent a lot of time here trying to mock out cuid to get a stable value but for reasons I don't understand `jest.mock('@bugsnag/cuid')` was just not doing anything. It does work if I update jest to v29 but that causes a load of other problems elsewhere and it was too much work to upgrade for this, hence the `expect.any(String)`.

Note for when we upgrade jest, a stable value can be obtained with:

```
jest.mock('@bugsnag/cuid', () => () => 'abc123')
jest.useFakeTimers()
jest.setSystemTime(new Date(...))
```

packages/delivery-fetch/delivery.js
```diff
@@ -1,19 +1,40 @@
...
 
-const delivery = (client, fetch = global.fetch) => ({
+function addIntegrityHeader (windowOrWorkerGlobalScope, requestBody, headers) {
+  if (windowOrWorkerGlobalScope.isSecureContext && windowOrWorkerGlobalScope.crypto && windowOrWorkerGlobalScope.crypto.subtle && windowOrWorkerGlobalScope.crypto.subtle.digest && typeof TextEncoder === 'function') {
+    const msgUint8 = new TextEncoder().encode(requestBody)
+    return windowOrWorkerGlobalScope.crypto.subtle.digest('SHA-1', msgUint8).then((hashBuffer) => {
+      const hashArray = Array.from(new Uint8Array(hashBuffer))
+      const hashHex = hashArray
+        .map((b) => b.toString(16).padStart(2, '0'))
+        .join('')
```
djskinner, line 10: See See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string

packages/react-native/src/config.js
```diff
@@ -3,7 +3,21 @@ const stringWithLength = require('@bugsnag/core/lib/validators/string-with-lengt
...
+  'logger',
+  'metadata',
+  'user',
+  'context',
+  'codeBundleId',
+  'plugins',
+  'featureFlags',
+  'reportUnhandledPromiseRejectionsAsHandled',
+  // sendPayloadChecksums gets set in core because its 'default value' depends on other config values and so cannot remain unset
+  'sendPayloadChecksums'
```
djskinner, line 18: this doesn't seem ideal but I'm not sure there is a better way to get the "default value" behaviour of `sendPayloadChecksums` without actually setting its value when the client is created, i.e. this part:

```
    // sendPayloadChecksums is false by default unless custom endpoints are not specified
    if (!opts.endpoints) {
      opts.sendPayloadChecksums = 'sendPayloadChecksums' in opts ? opts.sendPayloadChecksums : true
    }
```

I think it would be better if the `defaultValue` part of config could produce its value based on the rest of the config but that seemed way out of scope here,

packages/core/client.js
```diff
@@ -121,6 +121,11 @@ class Client {
       return schema
     }, this._schema)
 
+    // sendPayloadChecksums is false by default unless custom endpoints are not specified
+    if (!opts.endpoints) {
+      opts.sendPayloadChecksums = 'sendPayloadChecksums' in opts ? opts.sendPayloadChecksums : true
+    }
```
djskinner, line 127: the `sendPayloadChecksums` is added to core so the behaviour is the same for all clients i.e. browser and web worker. It doesn't apply to react native as that uses native delivery so they would need separate support and configuration in the native notifiers (which I believe they already have)

packages/delivery-fetch/delivery.js
```diff
@@ -1,18 +1,44 @@
 import payload from '@bugsnag/core/lib/json-payload'
 
-const delivery = (client, fetch = global.fetch) => ({
+function getIntegrityHeaderValue (sendPayloadChecksums, windowOrWorkerGlobalScope, requestBody, headers) {
+  if (sendPayloadChecksums && windowOrWorkerGlobalScope.isSecureContext && windowOrWorkerGlobalScope.crypto && windowOrWorkerGlobalScope.crypto.subtle && windowOrWorkerGlobalScope.crypto.subtle.digest && typeof TextEncoder === 'function') {
```
djskinner, line 4: see https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto and https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

packages/delivery-xml-http-request/test/delivery.test.ts
```diff
@@ -66,6 +67,118 @@ describe('delivery:XMLHttpRequest', () => {
...
+      this.data = data
+      this.readyState = XMLHttpRequest.HEADERS_RECEIVED
+      this.onreadystatechange()
+
+      setTimeout(() => {
+        this.status = 200
+        this.readyState = XMLHttpRequest.DONE
+        this.onreadystatechange()
+      }, 500)
+    }
```
gingerbenw, line 112: can this code be shared between tests?

test/browser/features/integrity.feature
```diff
@@ -0,0 +1,19 @@
+@skip_ie_11 @skip_ios_10 @skip_safari_10 @skip_safari_16 @skip_edge_17 @skip_chrome_43
+Feature: Bugsnag-Integrity header
+
+Scenario: Integrity headers are set when setPayloadChecksums is true
+  When I navigate to the test URL "/integrity/script/a.html"
```
gingerbenw, line 5: could we name the html something closer to what the test actually does? otherwise we have to reference between the feature and the fixture to figure out what `a.html` does

djskinner:
> Do we also need to set this in the `delivery-x-domain-request` package? I suppose IE8 and 9 won't support an integrity header?

No, XDomainRequest is only available on ie8 and ie9 neither which support the necessary APIs for calculating and setting the integrity header. even ie11 is excluded because it doesn't have full native promise support, hence the `typeof Promise !== 'undefined' && Promise.toString().indexOf('[native code]') !== -1` check.

test/browser/features/integrity.feature
gingerbenw: yeah, that sounds good

gingerbenw (approved):

PR #2224, opened 2024-10-16
[Integration] Typescript - Duplicate

gingerbenw:
## Goal

Convert packages to TypeScript using rollup to bundle both ES modules and Common JS modules for maximum compatibility

## Changeset

- Install and configure rollup
- Add type annotations to browser packages
- Update build scripts to generate es modules

## Testing

Covered by CI

## Outstanding

- [ ] Test tree shaking manually


packages/plugin-react/src/plugin.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  componentDidCatch (error: Error, info: ErrorInfo) {
+    const { onError } = this.props
+    const handledState = { severity: 'error', unhandled: true, severityReason: { type: 'unhandledException' } }
+    // @ts-ignore internal API
+    const event = client.Event.create(
+      error,
+      true,
+      handledState,
+      1
```
gingerbenw, line 103: Looking at the signature for `Event.create` the `errorFramesToSkip` parameter should be the 5th - the 4th parameter should be for the component name

PR #2235, opened 2024-10-23
fix: :rocket: update release script after lerna version change

gingerbenw:
## Goal

Ensure release script runs correctly

### DIFF
--- a/.github/workflows/pr-diff.yml
+++ b/.github/workflows/pr-diff.yml
@@ -22,7 +22,6 @@ jobs:
         run: |
           mkdir .diff
           npm ci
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-diff.yml
```diff
@@ -22,7 +22,6 @@ jobs:
         run: |
           mkdir .diff
           npm ci
-          npx lerna bootstrap
```
gingerbenw, line 25: added in release branch as main still required the bootstrap command - but doesn't any more

tomlongridge (approved):

PR #2247, opened 2024-10-29
Convert @bugsnag/plugin-network-breadcrumbs to TypeScript

AnastasiiaSvietlova:
## Goal

Convert package to TypeScript and ES Modules

## Changeset

Refactor @bugsnag/plugin-network-breadcrumbs

### DIFF
--- a/packages/react-native/window.d.ts
+++ b/packages/react-native/window.d.ts
@@ -1,4 +1,0 @@
-// Implement basic Window type for plugin-network-breadcrumbs
-declare interface Window {
-  readonly fetch: unknown
-}
--- a/packages/react-native/window.d.ts
+++ b/packages/react-native/window.d.ts
@@ -1,0 +1,4 @@
+// Implement basic Window type for plugin-network-breadcrumbs
+declare interface Window {
+  readonly fetch: unknown
+}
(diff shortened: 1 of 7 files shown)


gingerbenw:
@yousif-bugsnag I've made quite a few changes from @AnastasiiaSvietlova's original PR, so would appreciate another pair of eyes 🙏🏻

packages/react-native/window.d.ts
yousif-bugsnag, line 1: I don't think I understand why this is necessary?

gingerbenw: it would otherwise fail to compile because window is an unknown type in react native

yousif-bugsnag: oh, I see - I think we could fix that by declaring our own interface in plugin-network-breadcrumbs, e.g:

```ts
interface GlobalWithFetchAndXHR {
  fetch: typeof fetch
  XMLHttpRequest: typeof XMLHttpRequest
}

export default (_ignoredUrls = [], win: GlobalWithFetchAndXHR = window): Plugin => {
  // ...
}
```

yousif-bugsnag (approved):

PR #2252, opened 2024-11-12
Refactor @bugsnag/browser

gingerbenw:
## Goal

This PR converts the last of the internal browser plugins and the browser client to TypeScript, using Rollup to bundle both common js and es modules

## Changeset

- Convert browser package to TypeScript
- Convert plugin-simple-throttle to TypeScript
- Convert plugin-interaction-breadcrumbs to TypeScript
- Convert plugin-navigation-breadcrumbs to TypeScript
- Create browser bundle for both npm and cdn using rollup

## Testing


jest.config.js
```diff
@@ -49,7 +49,9 @@ module.exports = {
       'plugin-simple-throttle',
       'plugin-console-breadcrumbs',
       'plugin-browser-session'
-    ]),
+    ], {
+      setupFiles: ['<rootDir>/jest/setup/mockEventTarget.js']
```
djskinner, line 53: See https://github.com/jsdom/jsdom/issues/2156

This is needed to fix the `TypeError: 'addEventListener' called on an object that is not a valid instance of EventTarget.` seen in the integration tests following the rollup changes

packages/browser/package.json
```diff
@@ -22,7 +29,8 @@
   "scripts": {
     "size": "../../bin/size dist/bugsnag.min.js",
     "clean": "rm -fr dist && mkdir dist",
-    "build": "npm run clean && npm run build:dist && npm run build:dist:min",
+    "build": "npm run clean && npm run build:npm",
+    "build:npm": "rollup --config rollup.config.npm.mjs",
     "build:dist": "cross-env NODE_ENV=production bash -c '../../bin/bundle src/notifier.js --standalone=Bugsnag | ../../bin/extract-source-map dist/bugsnag.js'",
     "build:dist:min": "cross-env NODE_ENV=production bash -c '../../bin/bundle src/notifier.js --standalone=Bugsnag | ../../bin/minify dist/bugsnag.min.js'",
```
djskinner, line 35: can these be removed?

gingerbenw: yep - and possibly the minification scripts, too?

packages/browser/src/index-umd.ts
```diff
@@ -0,0 +1,16 @@
...
+import Breadcrumb from '@bugsnag/core/breadcrumb'
+
+import assign from '@bugsnag/core/lib/es-utils/assign'
+
+import Bugsnag from './bugsnag'
+
+export default assign(Bugsnag, { Client, Event, Session, Breadcrumb })
+export type { BrowserBugsnagStatic, BrowserConfig } from './bugsnag'
+
+export type { Client, Event, Session, Breadcrumb, Plugin } from '@bugsnag/core'
```
djskinner, line 16: Ideally we would do `export type * from '@bugsnag/core'` but that's not supported with the current version of TypeScript we're using

djskinner: TODO: Check what other types need exporting here

packages/browser/src/bugsnag.ts
```diff
@@ -0,0 +1,133 @@
...
+import pluginDevice from '@bugsnag/plugin-browser-device'
+import pluginContext from '@bugsnag/plugin-browser-context'
+import pluginRequest from '@bugsnag/plugin-browser-request'
+import pluginThrottle from '@bugsnag/plugin-simple-throttle'
+import pluginConsoleBreadcrumbs from '@bugsnag/plugin-console-breadcrumbs'
+import pluginNetworkBreadcrumbs from '@bugsnag/plugin-network-breadcrumbs'
+import pluginNavigationBreadcrumbs from '@bugsnag/plugin-navigation-breadcrumbs'
+import pluginInteractionBreadcrumbs from '@bugsnag/plugin-interaction-breadcrumbs'
+// @ts-ignore
+import pluginInlineScriptContent from '@bugsnag/plugin-inline-script-content'
```
djskinner, line 24: I reverted the TS conversion of this plugin from this PR because it was broken

packages/browser/rollup.config.npm.mjs
```diff
@@ -0,0 +1,86 @@
...
+import typescript from '@rollup/plugin-typescript'
+import fs from 'fs'
+
+import createRollupConfig, { sharedOutput as commonSharedOutput } from "../../.rollup/index.mjs"
+
+const packageJson = JSON.parse(fs.readFileSync('./package.json'))
+
+const sharedOutput = {
+  ...commonSharedOutput,
+  strict: false, // 'use strict' in WebKit enables Tail Call Optimization, which breaks stack trace handling
```
djskinner, line 15: There is one tail call in out stack and our depth handling does not handle it. With TCO enabled the stack frame for the tail call is removed, which then causes problems with stack processing leading to unexpected results (only in WebKit browsers since that's the only one supporting TCO):

Screenshot 2024-12-27 at 10 14 05

Dropping `'use strict'` prevents TCO being enabled and resolves the issue

packages/browser/package.json
djskinner: Not yet as they are being used for e.g. plugin-react and others

packages/browser/src/index-umd.ts
djskinner: Actually I don't think any types need exporting from cjs/umd entry points

packages/browser/package.json
```diff
@@ -46,7 +50,13 @@
...
     "@bugsnag/plugin-strip-query-string": "^8.1.1",
     "@bugsnag/plugin-window-onerror": "^8.1.1",
-    "@bugsnag/plugin-window-unhandled-rejection": "^8.1.1"
+    "@bugsnag/plugin-window-unhandled-rejection": "^8.1.1",
+    "@rollup/plugin-babel": "^6.0.4",
+    "@rollup/plugin-commonjs": "^28.0.2",
+    "@rollup/plugin-node-resolve": "^16.0.0",
+    "@rollup/plugin-replace": "^6.0.2",
+    "@rollup/plugin-terser": "^0.4.4",
+    "tslib": "^2.8.1"
```
djskinner, line 59: tslib is required for `target: es3`

packages/browser/src/index-cjs.ts
```diff
@@ -0,0 +1,13 @@
+import Client from '@bugsnag/core/client'
+// @ts-ignore
+import Event from '@bugsnag/core/event'
+// @ts-ignore
+import Session from '@bugsnag/core/session'
+// @ts-ignore
+import Breadcrumb from '@bugsnag/core/breadcrumb'
```
gingerbenw, line 7: Assuming these ignores will be removed once we've finished conversion of the core package

packages/plugin-interaction-breadcrumbs/src/interaction-breadcrumbs.ts
```diff
@@ -1,10 +1,13 @@
+import { Plugin } from '@bugsnag/core'
+import type ClientWithInternals from 'packages/core/client'
```
gingerbenw, line 2: so this import is working fine?

packages/plugin-interaction-breadcrumbs/src/interaction-breadcrumbs.ts
```diff
@@ -23,7 +26,8 @@ module.exports = (win = window) => ({
 
 const trim = /^\s*([^\s][\s\S]{0,139}[^\s])?\s*/
 
-function getNodeText (el) {
+// TODO: Fix Type
```
gingerbenw, line 29: Want me to take some of these TODOs ?

packages/plugin-interaction-breadcrumbs/src/interaction-breadcrumbs.ts
djskinner: This one happens to work fine because that import path doesn't appear in the final generated types:

```

PR #2253, opened 2024-11-14
run e2e tests over https

djskinner:
## Goal

Get the browser e2e tests running over https so we can test functionality that requires a secure context.

## Design/Changeset

- Set the new --https flag on maze runner
- Update the test fixtures to use the new docs endpoint being served from `/docs`
- Do not use https for the following: ie11, safari16, ios10

## Testing

e2e tests passing locally and on CI


test/browser/features/support/env.rb
```diff
@@ -1,26 +1,24 @@
...
 
 def get_test_url(path)
 
   if Maze.config.aws_public_ip
     host = Maze.public_document_server_address
     api_host = Maze.public_address
   else
-    host = "#{ENV['HOST']}:#{Maze.config.document_server_port}"
-    api_host = "#{ENV['API_HOST']}:#{Maze.config.port}"
+    host = "#{ENV['HOST']}:9339/docs"
```
twometresteve, line 11: We can simplify this whole if/else block now:

```
  if Maze.config.aws_public_ip
    maze_runner = Maze.public_address
  else
    maze_runner = "#{ENV['HOST']}:9339"
  end
```

test/browser/features/support/env.rb
```diff
@@ -1,26 +1,24 @@
...
   end
 
-  notify = "http://#{api_host}/notify"
-  sessions = "http://#{api_host}/sessions"
-  logs = "http://#{api_host}/logs"
-  reflect= "http://#{api_host}/reflect"
+  notify = "https://#{api_host}/notify"
+  sessions = "https://#{api_host}/sessions"
+  logs = "https://#{api_host}/logs"
+  reflect= "https://#{api_host}/reflect"
```
twometresteve, line 18: This this becomes:
```suggestion
  notify = "https://#{maze_runner}/notify"
  sessions = "https://#{maze_runner}/sessions"
  logs = "https://#{maze_runner}/logs"
  reflect= "https://#{maze_runner}/reflect"
```

test/browser/features/support/env.rb
```diff
@@ -1,26 +1,24 @@
...
   config_query_string = "NOTIFY=#{notify}&SESSIONS=#{sessions}&API_KEY=#{$api_key}&LOGS=#{logs}&REFLECT=#{reflect}"
 
-  uri = URI("http://#{host}#{path}")
+  uri = URI("https://#{host}#{path}")
```
twometresteve, line 21: And this:
```suggestion
  uri = URI("https://#{maze_runner}/docs#{path}")
```

.buildkite/basic/browser-pipeline.yml
```diff
@@ -171,3 +168,32 @@ steps:
         concurrency: 2
         concurrency_group: "browserstack"
         concurrency_method: eager
+
+      #
+      # BrowserStack non-https tests
+      #
```
djskinner, line 174: not sure if there's a DRYer way of doing this rather than duplicating most of the previous section? A parameterized matrix, perhaps?

twometresteve: I don't think so unfortunately, the Buildkite YAML schema is quite limited when it comes to any kind of commonality or logic between steps.

test/browser/Gemfile.lock
twometresteve, line 1: We're just removing `Gemfile.lock`s from Git these days to reduce dependabot warnings - but you could just do that on your next PR to avoid another build.

twometresteve (approved):
Nice one - thanks.

PR #2255, opened 2024-11-20
Refactor delivery-x-domain-request to TypeScript and ES Modules

gingerbenw:
## Goal

Convert source code to TypeScript and export both es modules and common js modules

## Testing


packages/delivery-x-domain-request/package.json
```diff
@@ -16,6 +24,11 @@
   ],
```
AnastasiiaSvietlova, line 24: Shouldn't "files" have ["dist"]?

gingerbenw: yes! good spot.

PR #2260, opened 2024-11-22
Refactor @bugsnag/plugin-client-ip

AnastasiiaSvietlova:
## Goal

Convert source code to TypeScript and export both es modules and common js modules

## Testing


packages/plugin-client-ip/src/client-ip.ts
```diff
@@ -0,0 +1,32 @@
+import { Plugin } from 'packages/core/types'
+import assign from '@bugsnag/core/lib/es-utils/assign'
+
+/*
+ * Prevent collection of user IPs
+ */
+const plugin: Plugin = {
+  load: client => {
+    // @ts-expect-error _config is private API
```
gingerbenw, line 9: could we extend the `Plugin` interface here to get rid of these instances of ts-expect-error?

packages/plugin-client-ip/src/client-ip.ts
```diff
@@ -0,0 +1,37 @@
...
+interface InternalClient extends Client {
+  _config: {
+    collectUserIp: boolean
+  }
+}
+
+/*
+ * Prevent collection of user IPs
+ */
+const plugin: Plugin = {
```
gingerbenw, line 13: you can also change this to 

```interface ExtendedPlugin extends Plugin {
  configSchema: Record<string, ValidationOption>
}```

to remove the need for the second `@ts-expect-error`

gingerbenw (approved):

Issue #1259, opened 2021-01-22
Missing clearUser method on react-native


ghsdh3409:
### Describe the bug
* As an official reference [document](https://docs.bugsnag.com/platforms/react-native/react-native/legacy/configuration-options/#clearuser), there is `clearUser` method. But, after I upgrading v7, there is no `clearUser`.
* Is this breaking change? If not, what should I use instead of `clearUser`?

### Steps to reproduce
1. Use `Bugsnag.clearUser()`

### Environment
* Bugsnag version: 7.6
* React Native: 0.63.4
* Device (e.g. iphonex): Android and iOS

### Error screenshot
<img width="323" alt="스크린샷 2021-01-22 오후 3 58 41" src="https://user-images.githubusercontent.com/6203798/105458048-2ee7ed00-5ccb-11eb-8cf8-d84acc2049c0.png">

xander-jones:
Hey @ghsdh3409, the docs you've linked to are our legacy documents for the deprecated `bugsnag-react-native` package. The `clearUser()` method does not exist in the new React Native package.

In Bugsnag for React Native using `bugsnag-js` v7.3.0+, you can clear the user data though by calling `Bugsnag.setUser();` without any arguments in the JS layer.

jdferreira-biai:
@xljones Thanks!

Btw, that should be added in the docs. I spent some time reading the docs looking for how to clear the user data.

hannah-smartbear:
Hi @jdferreira-biai,

You can find information on how to clear user data using `setUser` in our docs [here](https://docs.bugsnag.com/platforms/javascript/customizing-error-reports/#setuser:~:text=Nag%27)-,null,be%20used%20as%20a%20parameter%20value%20to%20clear%20the%20user%20attribute.,-severity). Please let us know if you have any further questions!

jdferreira-biai:
> Hi @jdferreira-biai,
> 
> You can find information on how to clear user data using `setUser` in our docs [here](https://docs.bugsnag.com/platforms/javascript/customizing-error-reports/#setuser). Please let us know if you have any further questions!

@hannah-smartbear There is an issue. When I try to send `null` as value for the `setUser` function I get this type error:

```
Argument of type 'null' is not assignable to parameter of type 'string | undefined'. ts(2345)
```

### Example:
```
import Bugsnag from '@bugsnag/js'

Bugsnag.setUser(null) // here i get the type error
```

I'm using `@bugsnag/js@8.1.2`.

[...]

PR #2262, opened 2024-11-25
Corrected setUser TS interface to allow for null arguments

agrzegorzewski:
## Goal

The `setUser` method in the notifier can take null as a parameter to clear the value from the user object. The current TypeScript interface only allows for `string | undefined`, which causes type errors when trying to pass null as a parameter.

[BugSnag documentation](https://docs.bugsnag.com/platforms/javascript/customizing-error-reports/#setuser) mentions passing null to `setUser` as a possibility.

## Design

The type definition has been updated to accurately reflect the capabilities od the notifier. 

## Changeset

The type definition for `setUser`'s arguments has been extended from `id?: string` to `id?: string | null`.

## Testing

The change was tested manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - (plugin-angular) Generate type definition using Angular 17 [#2275](https://github.com/bugsnag/bugsnag-js/pull/2275)
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


mclack:
Highlighted in https://github.com/bugsnag/bugsnag-js/issues/1259#issuecomment-2465795944

gingerbenw (approved):

PR #2269, opened 2024-11-29
Soft fail iOS 10 tests

gingerbenw:
## Goal

Resolve pipeline instability by enabling `soft_fail` for iOS 10 tests due to BrowserStack availability


.buildkite/basic/browser-pipeline.yml
```diff
@@ -178,6 +178,7 @@ steps:
           - ios_10
         depends_on: "browser-maze-runner-bs"
         timeout_in_minutes: 30
+        soft_fail: true
```
twometresteve: As discussed IRL this would unfortunately mean allow fails on safari 16 too.  Suggest bumping to ios_11 or remove use of the matrix and just have separate steps, with one set to soft_fail.

twometresteve (approved):

test/browser/features/plugin_angular.feature
```diff
@@ -1,7 +1,7 @@
 @plugin_angular
 
 # Skipped on older Safari versions not supported by Angular 10 - Angular renders the fixture component twice, causing duplicate events to be reported
-@skip_safari_10 @skip_ios_10
+@skip_safari_10 @skip_below_ios_13
```
twometresteve, line 4: For this to work you have to implement the tag `@skip_below_ios_13` yourself, there's no magic behind it.  See `skip.rb` for the definitions of them.

gingerbenw: there is an implementation present in `packages/react-native/ios/vendor/bugsnag-cocoa/features/support/env.rb` which I thought would be used?

PR #2272, opened 2024-12-06
Restrict access for built-in github token

Cawllec:
## Goal

Prevents the workflow token having default access.


tomlongridge (approved):

Cawllec:
Absolutely 👍

PR #2275, opened 2024-12-11
Generate type definition using Angular 17

gingerbenw:
## Goal

Ensure `@bugsnag/plugin-angular` works with recent versions of angular

## Design

Update dev dependency to v17 to ensure that generated types are compatible with Angular 17 (earliest supported version according to the [angular documentation](https://angular.dev/reference/versions))

## Changeset

Update Angular dependencies to v17
Remove code for dropped `ngDebugContext` 

## Testing

Update test fixture to v17 and test against full ci build

### DIFF
--- a/examples/ts/angular/src/polyfills.ts
+++ b/examples/ts/angular/src/polyfills.ts
@@ -57,7 +57,7 @@ import 'core-js/es7/reflect';
 /***************************************************************************************************
  * Zone JS is required by default for Angular itself.
  */
-import 'zone.js/dist/zone';  // Included with Angular CLI.
+import 'zone.js';  // Included with Angular CLI.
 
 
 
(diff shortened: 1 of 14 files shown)


examples/ts/angular/src/polyfills.ts
```diff
@@ -57,7 +57,7 @@ import 'core-js/es7/reflect';
 /***************************************************************************************************
  * Zone JS is required by default for Angular itself.
  */
-import 'zone.js/dist/zone';  // Included with Angular CLI.
+import 'zone.js';  // Included with Angular CLI.
```
gingerbenw, line 60: as per [update guide](https://angular.dev/update-guide?v=16.0-17.0&l=2)

test/browser/features/plugin_angular.feature
```diff
@@ -9,4 +9,3 @@ Scenario: basic error handler usage
   And the test should run in this browser
   Then I wait to receive an error
   And the error is a valid browser payload for the error reporting API
-  And the event "metaData.angular" is not null
```
gingerbenw, line 12: Dropped as `ngDebugContext` is no longer present, and previously only worked in debug builds. Will raise a separate task to reintroduce similar metadata if available in ivy builds.

lemnik (approved):
LGTM

PR #2278, opened 2024-12-18
Support Angular 19

gingerbenw:
## Goal

This PR fixes a number of bugs and updates the current level of Angular support, using the latest build system

## Changeset

- Update angular dev dependencies to v19
- Build plugin using ng-packagr with [partial ivy](https://angular.dev/tools/libraries/creating-libraries#ensuring-library-version-compatibility) compilation mode, in order to support Angular 12 and above
- Use [`publishConfig.directory`](https://github.com/lerna/lerna/tree/main/libs/commands/publish#publishconfigdirectory) to ensure only the `dist` directory is published for updated angular package
- Use `prepack` script to re-build and ensure correct version number is used for published package

## Testing

- [x] Update Angular 17 end-to-end fixture to use `browser-esbuild` builder
- [x] Add Angular 12 end-to-end fixture


packages/plugin-angular/projects/plugin-angular/src/lib/bugsnag-error-handler.ts
```diff
@@ -0,0 +1,42 @@
...
+      handledState,
+      'angular error handler',
+      1
+    )
+
+    this.bugsnagClient._notify(event)
+
+    // Default Angular error handling
+    // ErrorHandler.prototype.handleError.call(this, error)
+    console.error(error)
```
djskinner, line 38: is this meant to be like this?

gingerbenw: oops, not the commented out code - but yes! this is the default behaviour of the angular error handler which we're replicating. calling the prototype to do this can cause a mismatch between the compiled code and the angular version. I'll add a comment covering this.

djskinner (approved):

yousif-bugsnag (approved):

PR #2279, opened 2025-01-09
plugin-react: TypesScript and rollup

djskinner:
## Goal

- Convert @bugsnag/plugin-react to TypeScript and distribute as es modules

## Design

- Use TypeScript to transpile and rollup to bundle consistent with @bugsnag/browser

## Changeset

- Convert plugin source to typescript
- Use rollup in build script to produce: esm, cjs and umd outputs

## Testing

- Existing plugin-react fixture is in TypeScript, tests passing


packages/plugin-react/src/plugin.ts
```diff
@@ -0,0 +1,121 @@
...
+  componentDidCatch (error: Error, info: ErrorInfo) {
+    const { onError } = this.props
+    const handledState = { severity: 'error', unhandled: true, severityReason: { type: 'unhandledException' } }
+    // @ts-ignore internal API
+    const event = client.Event.create(
+      error,
+      true,
+      handledState,
+      1
+    )
```
djskinner, line 105: this is not exposed on the manually maintained types of @bugsnag/core

gingerbenw: @tomlongridge this is one of those instances where we're using an optional parameter that isn't documented (in the types) - should we go ahead and expose these moving forwards?

gingerbenw (approved):

PR #2281, opened 2025-01-14
Remove node-fetch dependency

gingerbenw:
## Goal

Remove node-fetch to resolve CVE-2022-0235

## Testing

Covered by existing tests


djskinner:
since fetch is part of node 18 we should be able to drop node-fetch entirely

djskinner (approved):

PR #2282, opened 2025-01-16
Pin concurrent-ruby to 1.3.4 in react-native fixtures

gingerbenw:
## Goal

Fix failing iOS build due to latest concurrent-ruby release dropping requirement for logger


joshedney (approved):
LGTM!

PR #2286, opened 2025-01-21
node: TypeScript and rollup

djskinner:
## Goal

Convert `@bugsnag/node` to TypeScript and bundle with rollup, replacing the browserify bundling.

## Design

Consistency with new approach to other packages. Needed doing before the `@bugsnag/core` conversion because `browserify` does not understand `package.exports` (see https://github.com/bugsnag/bugsnag-js/pull/2296). 

## Testing

e2e tests passing


packages/browser/rollup.config.npm.mjs
```diff
@@ -33,7 +33,7 @@ const plugins = [
     preventAssignment: true,
     values: {
       'process.env.NODE_ENV': JSON.stringify('production'),
-      __BUGSNAG_NOTIFIER_VERSION__: JSON.stringify(packageJson.version),
+      __BUGSNAG_NOTIFIER_VERSION__: packageJson.version,
```
djskinner, line 36: I've reverted how the version handling is done across all packages because I'd misunderstood how it was working. As can be seen below the version is injected inside a string rather than as a variable. This is how it was done before. The problem with it being a variable is that outside of rollup you can't import or require the file without somehow defining the global variable first. This is done in some of the integration tests for example.

packages/node/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "types": ["jest", "node"],
```
djskinner, line 4: As mentioned before, the root `tsconfig.json` has `"types: ["jest"]`, which excludes other types - but we need `@types/node` here hence why it is redefined.

It would be better to remove `"types: ["jest"]` from the root `tsconfig.json` but doing so means that `@types/react-native` are also included, which then conflict with lib dom, as defined by `"lib": [ "dom", "esnext" ],`.

So to do that we'll need to restructure the way jest / typescript is set up. One for a later PR.

gingerbenw (approved):

PR #2289, opened 2025-01-22
web worker: TypeScript and rollup

djskinner:
## Goal

- Convert @bugsnag/web-worker TypeScript and distribute as es modules

## Design

- Use TypeScript to transpile and rollup to bundle consistent with @bugsnag/browser

## Changeset

- Convert notifier source to typescript
- Use rollup in build script to produce: esm and umd outputs

## Testing

- Existing web-worker fixture is in TypeScript, tests passing


packages/web-worker/src/bugsnag.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface WorkerBugsnagStatic extends BugsnagStatic {
+  start(apiKeyOrOpts: string | WorkerConfig): Client
+  createClient(apiKeyOrOpts: string | WorkerConfig): Client
+}
+
+const name = 'Bugsnag Web Worker'
+const url = 'https://github.com/bugsnag/bugsnag-js'
+// @ts-ignore
+const version = __BUGSNAG_NOTIFIER_VERSION__ // eslint-disable-line no-undef
```
djskinner, line 31: Not sure why but there seems to be a problem with this being `__VERSION__` specifically. If I try to use that rollup blows up with:

```
[!] RollupError: src/bugsnag.ts (14:25): Expected a semicolon (Note that you need plugins to import files that are not JavaScript)
src/bugsnag.ts (14:25)
12: var name = 'Bugsnag Web Worker';
13: var url = 'https://github.com/bugsnag/bugsnag-js';
14: var version = __VERSION__;
                             ^
15: var schema = assign({}, baseConfig, workerConfig);
```

packages/web-worker/package.json
```diff
@@ -28,11 +34,8 @@
     ],
     "scripts": {
         "clean": "rm -fr dist && mkdir dist",
-        "build": "npm run clean && npm run build:dist && npm run build:dist:min",
-        "build:dist": "webpack",
-        "build:dist:min": "webpack --optimization-minimize --output-filename=bugsnag.web-worker.min.js",
-        "build:dist:esm": "webpack --config esm.config.js",
-        "build:dist:esm.min": "webpack --config esm.config.js --optimization-minimize --output-filename=bugsnag.web-worker.min.mjs",
```
djskinner, line 35: these don't appear to have been used. it looks like we're also publishing this to cdn but then there are no docs for that. so do we need to support cdn for web workers?

packages/browser/rollup.config.npm.mjs
```diff
@@ -33,7 +33,7 @@ const plugins = [
     preventAssignment: true,
     values: {
       'process.env.NODE_ENV': JSON.stringify('production'),
-      values: { __VERSION__: packageJson.version },
+      __BUGSNAG_NOTIFIER_VERSION__: JSON.stringify(packageJson.version),
```
djskinner, line 36: fix version in browser notifier. maybe we need a test for this, if possible. See also comment below about strange problems with __VERSION__

packages/web-worker/src/bugsnag.ts
```diff
@@ -0,0 +1,101 @@
...
+      pluginClientIp,
+      pluginPreventDiscard,
+      pluginWindowOnError(self, 'worker onerror'),
+      pluginWindowUnhandledRejection(self)
+    ]
+
+    // configure a client with user supplied options
+    const bugsnag = new ClientWithInternals(opts, schema, internalPlugins, { name, version, url })
+
+    bugsnag._setDelivery(client => delivery(client, self.fetch))
```
djskinner, line 62: delivery-fetch` is implemented like `const delivery = (client, fetch = global.fetch) => ({`

In webpack this `global.fetch` gets converted to a webpack require that somehow resolves this properly for web workers. However in rollup it remains as `global.fetch` which is undefined. We could perhaps replace the default value with `globalThis.fetch` but I'd think I'd rather see the default omitted requiring the notifier implementation to explicitly state what fetch to use.

packages/web-worker/src/bugsnag.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const notifier: WorkerClient = {
+  _client: null,
+  createClient: (opts) => {
+    // handle very simple use case where user supplies just the api key as a string
+    if (typeof opts === 'string') opts = { apiKey: opts }
+    if (!opts) opts = {} as unknown as Config
+
+    const internalPlugins = [
+      pluginBrowserDevice(navigator, null),
```
djskinner, line 51: As per `delivery-fetch` comment below maybe these plugins would be better off not assuming what environment they are running in

packages/core/client.d.ts
```diff
@@ -82,4 +82,6 @@ export default class ClientWithInternals<T extends Config = Config> extends Clie
   _loadPlugin(plugin: Plugin): void
 
   _isBreadcrumbTypeEnabled(type: string): boolean
+
+  public addOnError(fn: OnErrorCallback, front?: boolean): void;
```
djskinner, line 86: the public types do not currently expose the second param

packages/web-worker/src/bugsnag.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface WorkerBugsnagStatic extends BugsnagStatic {
+  start(apiKeyOrOpts: string | WorkerConfig): Client
+  createClient(apiKeyOrOpts: string | WorkerConfig): Client
+}
+
+const name = 'Bugsnag Web Worker'
+const url = 'https://github.com/bugsnag/bugsnag-js'
+// @ts-ignore
+const version = __BUGSNAG_NOTIFIER_VERSION__
```
djskinner, line 31: Not sure why but there seems to be a problem with this being `__VERSION__` specifically. If I try to use that rollup blows up with:

```
[!] RollupError: src/bugsnag.ts (14:25): Expected a semicolon (Note that you need plugins to import files that are not JavaScript)
src/bugsnag.ts (14:25)
12: var name = 'Bugsnag Web Worker';
13: var url = 'https://github.com/bugsnag/bugsnag-js';
14: var version = __VERSION__;
                             ^
15: var schema = assign({}, baseConfig, workerConfig);
```

packages/web-worker/package.json
gingerbenw: I think that's one for @tomlongridge - as far I understand the CDN was a requirement, but if it's not been documented we will need to remedy that

tomlongridge: It sounds like it, but also opens the question about whether we have tests to cover that scenario. I'd suggest ticketing for future, but keep doing what we're doing now as far as publishing is concerned so we don't stop publishing if anyone is using it?

gingerbenw (approved):

PR #2295, opened 2025-01-28
fix: :bug: fix issue with addOnSession reference when Bugsnag is not …

gingerbenw:
…provided to BugsnagErrorHandler


yousif-bugsnag (approved):

gingerbenw:
@tomlongridge happy with this one in principal before it goes in?

tomlongridge (approved):

PR #2231, opened 2024-10-17
bump jest

djskinner:
I was having some issues with `jest.mock` not working and bumping to the latest version of jest fixes it

PR #2296, opened 2025-01-28
core: move to src, add TypeScript and rollup

djskinner:
## Goal

Convert @bugsnag/core to TypeScript and bundling with rollup in a series of smaller PRs that are easy to review.

The bulk of this PR is to move source code from the root of the package to `src` whilst allowing existing imports to work (still pointing to the source code) using `package.exports`.

TypeScript and rollup config and a build step is added but the `dist` assets are not yet used.


bin/local-test-util
```diff
@@ -123,7 +123,9 @@ async function packNotifiers (notifier) {
     ? [ notifier ]
     : [ 'js', 'browser', 'node', 'web-worker', 'plugin-angular', 'plugin-react', 'plugin-vue' ]
   for (const n of notifiers) {
-    await ex(`npm`, [ `pack`, `--verbose`, `packages/${n}/` ])
+    let packageLocation = `packages/${n}/`
+    if (n === 'plugin-angular') packageLocation += 'dist/'
+    await ex(`npm`, [ `pack`, `--verbose`, packageLocation ])
```
djskinner, line 128: cherry pick from https://github.com/bugsnag/bugsnag-js/pull/2300

jest.config.js
```diff
@@ -1,6 +1,7 @@
 const testsForPackage = (packageName) => `<rootDir>/packages/${packageName}/**/*.test.[jt]s?(x)`
 
 const project = (displayName, packageNames, config = {}) => ({
+  resolver: '<rootDir>/jest/node-exports-resolver',
```
djskinner, line 4: the jest version being used doesn't support `package.exports` so we need to use a custom resolver. This is largely based on https://www.npmjs.com/package/jest-node-exports-resolver but required further changes:
- Using pkg-up to find the right package json for monorepo packages as the existing logic wasn't working
- Setting some default "conditions" as `jest-node-exports-resolver` out of the box no longer supports our version of jest (see https://github.com/k-g-a/jest-node-exports-resolver/issues/17)

An attempt to update jest was previously made (https://github.com/bugsnag/bugsnag-js/pull/2231) but there were issues because the helpers we use to run electron tests are no longer maintained and do not work with newwer versions of jest.

As alluded to in the previous PR its possible we could work around this by using multiple versions of jest (if a particular package needs a different version temporarily) rather than the single centralized way it is managed now.

packages/core/package.json
```diff
@@ -1,8 +1,40 @@
 {
   "name": "@bugsnag/core",
-  "main": "index.js",
+  "main": "src/index.js",
```
djskinner, line 3: In this PR the package source is moved to a `src` folder but the package exports are still referring to the source (not the generated TypeScript). This allows the file move to be separated from the TypeScript conversion allowing for smaller PRs that are easier to review

packages/core/package.json
```diff
@@ -1,8 +1,40 @@
...
   "version": "8.1.1",
   "types": "types/index.d.ts",
+  "exports": {
```
djskinner, line 6: due to the introduction of a `src` folder existing import references such as `@bugsnag/core/lib/es-utils/map` require an export map for them to point to the correct folder (src for now, will migrate to dist is subsequent PRs)

Presumably we'll want to get back to exporting/importing everything from just `@bugsnag/core` in subsequent PRs

packages/core/package.json
```diff
@@ -13,10 +45,17 @@
...
-    "*.js"
+    "*.js",
+    "dist"
   ],
+  "scripts": {
+    "size": "../../bin/size dist/bugsnag.min.js",
+    "clean": "rm -fr dist && mkdir dist",
+    "build": "npm run clean && npm run build:npm",
+    "build:npm": "rollup --config rollup.config.npm.mjs"
+  },
```
djskinner, line 58: TypeScript compilation is added (with support for .js files) to the build step but the dist assets are not yet being used

packages/core/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "lib": [ "dom", "es2022" ],               /* Specify library files to be included in the compilation. */
+    "allowJs": true,
```
djskinner, line 5: note: allowJs true

packages/core/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "outDir": "dist",
```
djskinner, line 6: I was having some issues with allowJs and for unclear reasons specifying this solved them. See https://stackoverflow.com/questions/42609768/typescript-error-cannot-write-file-because-it-would-overwrite-input-file

tsconfig.json
```diff
@@ -1,9 +1,11 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "module": "commonjs",
+    "module": "es2015",
+    "moduleResolution": "bundler",
```
djskinner, line 5: this allows TypeScript to work with package.exports

djskinner: https://stackoverflow.com/questions/70296652/how-can-i-use-exports-in-package-json-for-nested-submodules-and-typescript

packages/node/src/bugsnag.ts
```diff
@@ -6,8 +6,6 @@ import assign from '@bugsnag/core/lib/es-utils/assign'
 import { schema as baseConfig } from '@bugsnag/core/config'
 import browserConfig from './config'
 
-import Event from '@bugsnag/core/event'
```
djskinner, line 9: Using this export from `'@bugsnag/core/event'` was causing `'@bugsnag/core/event'` to appear in the generated types, which was causing issues with some older versions of TypeScript because it would need to resort to `package.exports`, which requires a certain version of TypeScript and `moduleResolution: bundler`.

jest.config.js
gingerbenw: I personally prefer the de-centralised approach to running the tests, but I think it was harder to track changes in test coverage? if we can get around that I'd be happy to take that approach in the future

packages/core/rollup.config.npm.mjs
```diff
@@ -0,0 +1,47 @@
...
+  nodeResolve({
+    browser: true
+  }),
+  commonjs(),
+  typescript(),
+  babel({ babelHelpers: 'bundled' }),
+]
+
+const external = [
+  // '@bugsnag/cuid',
```
gingerbenw, line 18: commented out code - is this being added back in for a subsequent PR or can this just be an empty array?

packages/node/src/bugsnag.ts
gingerbenw: I get the feeling a lot of these have come from auto-complete

gingerbenw (approved):

packages/core/rollup.config.npm.mjs
djskinner: see https://github.com/bugsnag/bugsnag-js/pull/2316/files#r1949315462

packages/node/src/bugsnag.ts
djskinner: Yeah, it would be good to get around preventing invalid imports with eslint rules as we do on the dashboard

PR #2305, opened 2025-02-04
core: use dist assets for root export

djskinner:
For the default export of `@bugsnag/core` switch to using the `dist` assets created by rollup/TypeScript. This means all imports of `@bugsnag/core` will now use the compiled assets rather than the source code. Note that imports such as `@bugsnag/core/breadcrumb` are still pointing directly at the source code for now


jest/node-exports-resolver.js
```diff
@@ -54,7 +54,7 @@ function getPackageJson (packageName) {
     return require(`${packageName}/package.json`)
```
djskinner, line 54: In summary I had to change the temporary jest resolver to fix some issues with resolving `packageName=@bugsnag/core, submoduleName=.` by removing the ` && submoduleName !== '.'` condition. But then needed to return null instead of erroring for node modules that resolve with e.g. `packageName=https, submoduleName=.`

Hopefully we'll be able to bump jest later and we'll be able to drop this

packages/core/test/types.test.ts
```diff
@@ -1,22 +1,23 @@
-import Bugsnag, { Client, Config } from '..'
```
djskinner, line 1: Not sure what was going on here but `@bugsnag/core` never had a default export but this is a confusion from mising cjs exports and esm imports. In cjs it would have been:

```
const { Client, Breadcrumb, Event, Session } = require('@bugsnag/core')` which makes sense against the deleted index file above
```

packages/core/types/client.d.ts
```diff
@@ -6,15 +6,98 @@ import {
   OnSessionCallback,
```
djskinner, line 6: On the question of public versus internal interfaces for the notifiers themselves we can address that later. However I don't think it makes much sense at all for `@bugsnag/core` to have this concept. We don't document/expect our customers to use `@bugsnag/core` directly anyway so we can just keep things simple and export the actual interfaces.

For now though, we're still using the manual types - we can't switch to generated types until all the JS files that make up the external interfaces are converted to TypeScript otherwise invalid types are generated.

packages/core/package.json
```diff
@@ -6,12 +6,12 @@
   "exports": {
     ".": {
       "types": "./types/index.d.ts",
-      "import": "./src/index.js",
-      "default": "./src/index.js"
+      "import": "./dist/index.mjs",
+      "default": "./dist/index.cjs"
```
djskinner, line 10: This is the main change, the rest of this PR are the consequences

gingerbenw (approved):

PR #2316, opened 2025-02-10
drop @bugsnag/core/breadcrumb export

djskinner:

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,8 +40,7 @@
       "default": "./src/event.js",
       "types": "./src/event.d.ts"
     },
-    "./session": "./src/session.js",
-    "./breadcrumb": "./src/breadcrumb.js"
+    "./session": "./src/session.js"
   },
   "description": "Core classes and utilities for Bugsnag notifiers",
   "homepage": "https://www.bugsnag.com/",
--- a/packages/core/rollup.config.npm.mjs
+++ b/packages/core/rollup.config.npm.mjs
@@ -15,7 +15,7 @@ const plugins = [
 ]
 
 const external = [
-  // '@bugsnag/cuid',
+  '@bugsnag/cuid',
 ]
 
 export default [
--- a/packages/core/types/client.d.ts
+++ b/packages/core/types/client.d.ts
@@ -6,15 +6,30 @@ import {
   OnSessionCallback,
   OnBreadcrumbCallback,
   User,
-  FeatureFlag
+  FeatureFlag,
+  Config,
+  Plugin,
 } from './common'
 import Event from './event'
 import Session from './session'
 
-declare class Client {
-  protected constructor();
+interface Notifier {
+  name: string
+  version: string
+  url: string
+}
 
-  // reporting errors
+declare class Client<T extends Config = Config> {
+  // "private" interfaces
+  public constructor(opts: T, schema?: {[key: string]: any}, internalPlugins?: Plugin[], notifier?: Notifier)
+  _breadcrumbs: Breadcrumb[]
+
+  // access to internal classes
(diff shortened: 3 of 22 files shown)


packages/core/types/client.d.ts
```diff
@@ -6,15 +6,30 @@ import {
...
-  protected constructor();
+interface Notifier {
+  name: string
+  version: string
+  url: string
+}
+
+declare class Client<T extends Config = Config> {
+  // "private" interfaces
+  public constructor(opts: T, schema?: {[key: string]: any}, internalPlugins?: Plugin[], notifier?: Notifier)
```
djskinner, line 24: On the question of public versus internal interfaces for the notifiers themselves we can address that later. However I don't think it makes much sense at all for `@bugsnag/core` to have this concept. We don't document/expect our customers to use `@bugsnag/core` directly anyway so we can just keep things simple and export the actual interfaces.

For now though, we're still using the manual types - we can't switch to generated types until all the JS files that make up the external interfaces are converted to TypeScript otherwise invalid types are generated.

packages/plugin-interaction-breadcrumbs/test/interaction-breadcrumbs.test.ts
```diff
@@ -1,7 +1,6 @@
 import plugin from '../src/interaction-breadcrumbs'
 
-import Client from '@bugsnag/core/client'
-import Breadcrumb from '@bugsnag/core/breadcrumb'
+import { Breadcrumb, Client } from '@bugsnag/core'
```
djskinner, line 3: Importing Client from `@bugsnag/core` also because otherwise in the test below the Breadcrumb prototype used in `c._breadcrumbs` is different to that references by `new Breadcrumb` because `import Client from '@bugsnag/core/client'` actually refers to the source file. Hence the above changes to the public types

packages/core/rollup.config.npm.mjs
```diff
@@ -15,7 +15,7 @@ const plugins = [
 ]
 
 const external = [
-  // '@bugsnag/cuid',
+  '@bugsnag/cuid',
```
djskinner, line 18: In the jest tests imports from `@bugsnag/core/breadcrumb` refer to the source code and changing this to `import { Breadcrumb } from '@bugsnag/core'` now references the built dist code. Jest blew up at this point as it parses the whole cjs bundle and sees undefined globals like `navigator` and `self`.

I think core should really externalise all its dependencies and it be left to the individual notifiers to decide what, if any node modules to externalise.

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bugsnag/core",
-  "main": "src/index.js",
+  "main": "./dist/index.cjs",
```
djskinner, line 3: Forgot to update this in previous PR but it didn't matter because it seems only react-native uses it and before this PR the react-native notifier wasn't doing any imports from `@bugsnag/core`, only direct to source ones

.rollup/index.mjs
```diff
@@ -39,7 +39,7 @@ function createRollupConfig (options = defaultOptions()) {
         format: 'esm'
       }
     ],
-    external: ['@bugsnag/core'].concat(options.external),
+    external: options.external,
```
djskinner, line 42: the notifiers themselves use this `createRollupConfig` function but we need `@bugsnag/core'` bundling in those. Its more code but I think clearer to be explicit about this.

djskinner: This is apparent now because previously the imports of `@bugsnag/core/breadcrumb` etc weren't being treated as external, and the notifiers we not importing from `@bugsnag/core` until now

gingerbenw (approved):

PR #2327, opened 2025-02-14
switch to generated type declarations

djskinner:
Replace the manually created type definitions with those now being generated by TypeScript.

This exposed a few areas since the previous PR where the types can be improved. However, there is also liberal usage of `@ts-expect-error` for areas where the typings fail to work whilst avoiding large rewriting of the code.


packages/core/src/client.ts
```diff
@@ -2,66 +2,22 @@ import configSchema from './config'
...
+import { BreadcrumbType, Config, Delivery, FeatureFlag, LoggerConfig, Notifier, OnBreadcrumbCallback, OnErrorCallback, OnSessionCallback, Plugin, SessionDelegate, User } from './common'
 
 const noop = () => {}
-
-interface LoggerConfig {
-  debug: (msg: any) => void
-  info: (msg: any) => void
-  warn: (msg: any) => void
-  error: (msg: any, err?: unknown) => void
-}
```
djskinner, line 25: One downside to having babel transpile from TypeScript is that it's not type aware. So if you do:

```
export { LoggerConfig } from './common'
```

It fails because `LoggerConfig` is only a type.

Due to our requirement to support lower TS versions we couldn't do `export { type LoggerConfig } from './common'` which does fix the issue.

So I moved all the types to `common.js` which are exported like `export * from './common'`` and there are no complaints.

However, the generated types that TyeScript is now producing still do include the `type` modifier:

```
export { default } from './bugsnag';
export type { BrowserBugsnagStatic, BrowserConfig } from './bugsnag';
export * from '@bugsnag/core';
```

And this can't be solved by us internally not using the `type` modifier.

To get around this we'll have to use https://www.npmjs.com/package/downlevel-dts as a post-processing step on the types.

An added benefit of this is that internally we'll be able to use `type` modifier and whatever else we like that is supported by our more recent version of TypeScript

packages/core/src/common.ts
```diff
@@ -130,9 +126,9 @@ export interface Request {
 }
 
 export interface User {
-  id?: string | null
-  email?: string | null
-  name?: string | null
+  id?: string | null | undefined
+  email?: string | null | undefined
+  name?: string | null | undefined
```
gingerbenw, line 131: doesn't the `?` cover `undefined` already?

packages/core/test/types.test.ts
```diff
@@ -46,7 +46,8 @@ describe('Type definitions', () => {
   // eslint-disable-next-line jest/expect-expect
   it('works for reporting sessions', () => {
     const client = createClient({ apiKey: 'API_KEY' })
-    const sessionClient = client.startSession()
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const sessionClient = client.startSession()!
```
gingerbenw, line 50: could we just wrap this line in a logic test and drop the non-null assertion?

PR #2336, opened 2025-02-17
Remove @bugsnag/core/lib/es-utils/map

djskinner:
Remove @bugsnag/core/lib/es-utils/map replacing with builtin `Array.map`. As we have dropped very old IE support these are no longer necessary and hopefully removing them will improve the bundle size a bit. Furthermore for things that aren't supported by our target `browserslist` babel will provide the appropriate helpers/polyfills so we're able to write modern source and browser compatibility is handled automatically


packages/browser/src/config.ts
```diff
@@ -1,24 +1,29 @@
...
-  releaseStage: assign({}, schema.releaseStage, {
-    defaultValue: () => {
-      if (/^localhost(:\d+)?$/.test(window.location.host)) return 'development'
-      return 'production'
+  releaseStage: {
+    ...schema.releaseStage, ...{
+      defaultValue: () => {
+        if (/^localhost(:\d+)?$/.test(window.location.host)) return 'development'
+        return 'production'
+      }
```
gingerbenw, line 10: is the second spread object here unnecessary?

gingerbenw (approved):

PR #2337, opened 2025-02-17
external all node_modules in plugins

djskinner:
We don't want rollup bundling any external dependencies for plugins and other modules that are ultimately bundled into a notifier because that may introduce duplicate code into the final bundle when the notifier itself bundles everything with `external: []`


packages/delivery-x-domain-request/rollup.config.npm.mjs
```diff
@@ -2,5 +2,5 @@ import createRollupConfig from "../../.rollup/index.mjs";
 
 export default createRollupConfig({
   input: "src/delivery.ts",
-  external: ['@bugsnag/core', '@bugsnag/core/lib/json-payload']
+  external: [/node_modules/]
```
gingerbenw, line 5: so it this just a shorter way to add all the specific dependencies for each package?

PR #2339, opened 2025-02-18
Move AWS lambda tests from MacOS/Buildkite to GHA

Cawllec:
## Goal

This allows us to remove docker as a requirement from our MacOS boxes, and have the lambda tests run more consistently.

Note - We should wait for a maze-runner release feat. https://github.com/bugsnag/maze-runner/pull/722 before this is merged, as that PR fixes an issue with lambda tests seemingly flaking.

### DIFF
--- a/.github/workflows/aws-lambda.yml
+++ b/.github/workflows/aws-lambda.yml
@@ -28,7 +28,6 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: '3.1'
-        bundler-cache: true
 
     - name: Install Node
       uses: actions/setup-node@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/aws-lambda.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+    - name: Install libcurl4-openssl-dev and net-tools
+      run: |
+        sudo apt-get update
+        sudo apt-get install libcurl4-openssl-dev
+        sudo apt-get install net-tools
+
+    - run: sam --version
+
+    - uses: actions/checkout@v2
```
gingerbenw, line 22: should we be pinning these to a specific sha?

Cawllec: Good idea, done 👍

PR #2362, opened 2025-03-04
Fix main branch lockfile

yousif-bugsnag:
## Goal

The lockfile in main has somehow become out of sync with next, causing build and merge issues.

## Design

Copied the lockfile from next and ran `npm install` to apply the version updates for v8.3.0


yousif-bugsnag:
Force-merging with PR diff bot error due to lockfile

PR #2380, opened 2025-03-17
Ensure detect changes script runs on node queue

gingerbenw:
## Goal

Ensure detect changes script runs on node queue


joshedney (approved):
LGTM!

PR #2394, opened 2025-03-28
convert @bugsnag/core/lib/clone-client to TypeScript

AnastasiiaSvietlova:
## Goal

Convert @bugsnag/core/lib/clone-client to TypeScript

This PR also resolves an issue in the node end to end tests where the latest version of `@bugsnag/core` was installed from npm, rather than the monorepo.

RN 0.78 iOS test is failing due to a separate issue and is out of the scope of this PR 

## Testing


yousif-bugsnag (approved):
LGTM

PR #2399, opened 2025-04-03
convert @bugsnag/core/lib/json-payload to TypeScript

AnastasiiaSvietlova:
## Goal

Convert @bugsnag/core/lib/json-payload to TypeScript

## Testing


packages/core/src/common.ts
```diff
@@ -206,4 +206,19 @@ export interface BugsnagError {
   errorMessage: string
   type: string
   stacktrace: Stackframe[]
+}
+
+interface InternalEvent {
```
AnastasiiaSvietlova, line 211: @gingerbenw I'm not sure about this InternalEvent type.

packages/core/src/safe-json-stringify.d.ts
gingerbenw, line 1: Could we write up an actual definition for this? something like this should do:

```
declare module '@bugsnag/safe-json-stringify' {
    export default function stringify(
        value: any,
        replacer?: null | ((this: any, key: string, value: any) => any), 
        space?: null | string | number,
        options?: {
            redactedKeys?: Array<string | RegExp>;
            redactedPaths?: string[];
        }
    ): string;
}
```

packages/core/src/lib/json-payload.ts
```diff
@@ -0,0 +1,36 @@
+import jsonStringify from '@bugsnag/safe-json-stringify'
+import { JsonPayloadEvent } from "@bugsnag/core";
```
gingerbenw, line 2: we're already in core, so we can import this with a relative path

packages/core/src/common.ts
gingerbenw: this should be the same as what is returned from `Event.toJSON` so you could use `ReturnType<typeof Event.prototype.toJSON>` to avoid typing it all out manually

gingerbenw (dismissed):

yousif-bugsnag (approved):

PR #2407, opened 2025-04-09
Fix PR diff bot

gingerbenw:
## Goal

Resolve issues with PR diff bot from the package-lock file being out of sync. I haven't yet figured out why this has updated in the first place, but keeps cropping when raising release branches

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3869,9 +3869,9 @@
       }
     },
     "node_modules/@inquirer/figures": {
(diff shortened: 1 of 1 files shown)


yousif-bugsnag:
I think this is being caused by new versions of a plugin-angular dependency (`@inquirer/figures`). It's just going to keep happening each time we run `npm ci` after they release a new version - it's not specific to the release branch, it's just that we don't update master as often.

PR #2410, opened 2025-04-11
Support React Native 0.79

yousif-bugsnag:
## Goal

Add support for React Native 0.79

## Changeset

- Updated the React Native CLI to handle AppDelegate template changes in React Native 0.79 ios projects
- Added 0.79 tests to CI

## Testing

Covered by CI


.buildkite/basic/react-native-android-full-pipeline.yml
```diff
@@ -102,6 +102,7 @@ steps:
           - "0.74"
           - "0.76"
           - "0.77"
+          - "0.78"
```
gingerbenw, line 105: should we drop 0.72 from the matrix yet? I can't remember what the decision was on the support matrix, but 0.76 is now unsupported by react native.

.buildkite/package_manifest.json
```diff
@@ -138,7 +138,10 @@
     {
         "pipeline": ".buildkite/basic/react-native-cli-pipeline.yml",
         "block": ".buildkite/block/react-native-cli-pipeline.block.yml",
-        "paths": []
+        "paths": [
+            "test/react-native-cli",
+            "packages/react-native-cli"
+        ]
```
gingerbenw, line 144: 👍🏻

packages/react-native-cli/src/lib/__test__/fixtures/AppDelegate-after-079.swift
```diff
@@ -0,0 +1,51 @@
+import Bugsnag
```
gingerbenw, line 1: just to confirm, this is the only change between these versions?

gingerbenw (approved):

.buildkite/basic/react-native-android-full-pipeline.yml
yousif-bugsnag: I don't think we can drop 0.72 - our test matrix policy is the latest 3 versions, plus even versions back to 0.72

packages/react-native-cli/src/lib/__test__/fixtures/AppDelegate-after-079.swift
yousif-bugsnag: There's also the `Bugsnag.start` call on line 18.

The reason we need new before/after fixture files is because the whole `AppDelegate.swift` template has changed in 0.79 with the introduction of `RCTReactNativeFactory`

irekrog:
Hi @yousif-bugsnag , I do not want to open a new issue, so I ask here.
1. When will this version `8.3.2` appears on npm? Because is still not there https://www.npmjs.com/package/@bugsnag/react-native?activeTab=versions
2. Does this version fix Android support? Because there is an error in `0.79.1`
```
> Task :bugsnag_react-native:generateCodegenSchemaFromJavaScript FAILED
node:internal/modules/cjs/loader:1080
  throw err;
  ^

Error: Cannot find module '.../node_modules/@react-native/codegen/lib/cli/combine/combine-js-to-schema-cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v18.18.0

[Incubating] Problems report is available at: .../android/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':bugsnag_react-native:generateCodegenSchemaFromJavaScript'.
```

yousif-bugsnag:
Hi @irekrog,

`bugsnag-js` is a monorepo, and so not all packages are published on each new release (unless it's a new major version).  Unfortunately this can make looking at the GitHub releases a little confusing.

There is no v8.3.2 of `@bugsnag/react-native` - this release was only for the CLI tool (`@bugsnag/react-native-cli`). So the latest version is still v8.3.1.

This PR did add tests for React Native 0.79 though, and we haven't seen any build issues on Android. It looks like the gradle task isn't able to resolve the codegen script for some reason. Do you see the same error if you invoke the codegen task manually before building?
```sh
./gradlew generateCodegenArtifactsFromSchema
 ```

If you're still having problems please feel free to write in to support@bugsnag.com with some more details, and ideally some steps to reproduce, and we can try to investigate further.

Thanks!

irekrog:
Ok, got it. Yes, after calling `./gradlew generateCodegenArtifactsFromSchema` there is the same error I posted above. I will try to do the investment on my side, if there is still an error I will write to support.

Thanks!

irekrog:
Ok, when I manually added the package:
`"@react-native/codegen: "0.79.1"`

then it works fine, I think there is a problem with monorepo/pnpm (which I use)

PR #2422, opened 2025-04-17
Convert @bugsnag/core/lib/iserror to TypeScript

AnastasiiaSvietlova:
## Goal

Convert @bugsnag/core/lib/iserror to TypeScript

## Testing


packages/core/src/lib/iserror.js
```diff
@@ -1 +0,0 @@
-module.exports = require('iserror')
```
gingerbenw, line 1: now that we've replicated this code in TS, we can remove the dependency on this library from `packages/core/package.json`

packages/plugin-window-unhandled-rejection/src/unhandled-rejection.ts
```diff
@@ -1,5 +1,5 @@
-import { Config, Plugin } from '@bugsnag/core'
-import isError from '@bugsnag/core/lib/iserror'
+import { isError } from '@bugsnag/core'
+import type { Config, Plugin } from '@bugsnag/core'
```
gingerbenw, line 2: 👌🏻 good separation of types and functionality here

gingerbenw (approved):
Looks good to me!

packages/core/src/lib/iserror.ts
```diff
@@ -0,0 +1,36 @@
+// MIT License
+
+// Copyright (c) 2017 Anton Yefremov
```
AnastasiiaSvietlova, line 3: @tomlongridge please check if this copyright note is good to go. Thank you!

tomlongridge: LGTM 👍

PR #2428, opened 2025-04-25
Convert @bugsnag/core/lib/derecursify to TypeScript

AnastasiiaSvietlova:
## Goal

- Create standalone `@bugsnag/derecursify` package (private)
- Add build step to packages that make use of derecursify

## Testing


gingerbenw (approved):
looks good to me!

packages/core/src/lib/derecursify.ts
```diff
@@ -53,14 +51,14 @@ module.exports = function (data) {
     }
 
     // handle arrays, and all iterable non-array types (such as Set)
-    if (isArray(obj) || obj[Symbol.iterator]) {
+    if (isArray(obj) || obj instanceof Set || obj instanceof Map) {
       seen.push(obj)
       const safeArray = []
       try {
         for (const value of obj) {
```
AnastasiiaSvietlova, line 58: This line is still geting TS error in [build](https://buildkite.com/bugsnag/bugsnag-js/builds/12882#0196c42d-e9f0-4099-b861-f29e2ba07c30): `TS2802: Type 'any[] | Map<any, any> | Set<any>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.`

Although I don't see that error in VSCode. 

I've already tried a different checks in line 54 to check for all iterable non-array types. Such as:
`const isIterable = (value: any) => !!value[Symbol.iterator];`
`obj as Iterator<typeof obj>`
`Symbol.iterator in Object(obj)`
`typeof obj[Symbol.iterator] === 'function'`
All of them giving that error. Kinda out of ideas.

AnastasiiaSvietlova: And I've updated the tsconfig with: `"target": "es2020",`

yousif-bugsnag (approved):
LGTM

PR #2437, opened 2025-05-12
Convert @bugsnag/core/lib/iserrormetadata delegate

AnastasiiaSvietlova:


packages/core/src/lib/metadata-delegate.ts
```diff
@@ -0,0 +1,70 @@
+import assign from './es-utils/assign'
+
+interface MetadataDelegate {
+  add: (state: { [key: string]: any }, section: string, keyOrObj?: any, maybeVal?: any) => any
+  get: (state: { [key: string]: any }, section: string, key?: any) => any
+  clear: (state: { [key: string]: any }, section: string, key?: any) => any
```
gingerbenw, line 6: can we just double check what these instances of `any` should be? for example `key` looks like it should only be `string` or `undefined` according to the `get` function, and `section` should probably be `unknown`

AnastasiiaSvietlova: I checked its usage in other parts of the code. So we use `metadataDelegate` in `client.ts` and `event.ts`. Since these files are already converted to TS, I took them as a source of possible types.
Based on what you're wrote I will change the type of key to string and in `client.ts` too. 
But the `section` is always `string` 🤔

packages/core/src/lib/metadata-delegate.ts
```diff
@@ -0,0 +1,70 @@
+import assign from './es-utils/assign'
+
+interface MetadataDelegate {
+  add: (state: { [key: string]: any }, section: string, keyOrObj?: object | string, maybeVal?: any) => any
+  get: (state: { [key: string]: any }, section: string, key?: string | undefined) => any
+  clear: (state: { [key: string]: any }, section: string, key?: string | undefined) => any
```
gingerbenw, line 6: the question mark in `key?: string` also allows for `undefined` so that's probably fine as it was.

gingerbenw (approved):

PR #2438, opened 2025-05-12
Convert @bugsnag/core/lib/callback-runner to TypeScript

AnastasiiaSvietlova:
## Goal

Convert @bugsnag/core/lib/callback-runner to TypeScript

## Testing

### DIFF
--- a/packages/core/src/lib/callback-runner.ts
+++ b/packages/core/src/lib/callback-runner.ts
@@ -1,24 +1,31 @@
-const some = require('./async-every')
+import some from'./async-every'
 
-module.exports = (callbacks, event, onCallbackError, cb) => {
+import { NodeCallbackType } from './async-every'
+
+const runCallbacks = <T>(
+  callbacks: any,
+  event: T,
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/callback-runner.ts
```diff
@@ -1,12 +1,19 @@
-const some = require('./async-every')
+import some from './async-every'
 
-module.exports = (callbacks, event, onCallbackError, cb) => {
+type NodeCallbackType<T = any> = (error?: Error | null, result?: T) => void;
+
+const runCallbacks = <T>(
+  callbacks: any[],
```
AnastasiiaSvietlova, line 6: Wanted to specify the type here based on usage of this function. 
```suggestion
  callbacks: (OnErrorCallback | undefined)[],
```
But got errors from callback-runner.test as we are using other types there. I'm assuming that we expecting any type of `callbacks`.

yousif-bugsnag (approved):
LGTM

PR #2472, opened 2025-06-17
Convert @bugsnag/core/lib/async-every to TypeScript

AnastasiiaSvietlova:
## Goal

Convert @bugsnag/core/lib/async-every to TypeScript

## Testing


AnastasiiaSvietlova:
I'm merging to the `PLAT-13697-callback-runner` branch because this function is used only there. And I wanted to be sure that it will pass tests after converting to TS. But in `callback-runner` it's imported as `import some from './async-every'`
although it's called `every`. I don't know if it make sense to leave it as it is or rename it.

gingerbenw:
> I'm merging to the `PLAT-13697-callback-runner` branch because this function is used only there. And I wanted to be sure that it will pass tests after converting to TS. But in `callback-runner` it's imported as `import some from './async-every'` although it's called `every`. I don't know if it make sense to leave it as it is or rename it.

I would update the import to match the function name

PR #2494, opened 2025-07-08
Add React Native 0.80 tests

hamudi-ahmed:
## Goal

Adds React Native 0.80 to the test matrix

It looks like there's an issue with codegen on Android in React Native 0.80 - it seems like the codegen task is no longer invoked early enough, so the C++ compilation fails. Observed the same issue with a few different 3rd-party packages so the issue doesn't seem to be specific to the Bugsnag SDK.

To work around this, the test fixture build script now runs the codegen task explicitly when building with the new architecture.

## Testing

Covered by CI


gingerbenw:
@yousif-bugsnag sorry for the slow response on this one - did this get resolved in the new react native patch or is this still required?

gingerbenw (approved):

PR #2496, opened 2025-07-11
Optimise bundle size: json-payload

gingerbenw:
## Goal

Optimise browser bundle size

## Design

Update rollup configuration to minimise increase in bundle size

## Changeset

- Move json-payload to standalone package
- Remove `dXDomainRequest` from browser bundle (IE8/IE9 Compatibility)
- Remove webpack3 tests
- Update Android browser testing to v8 (duplicate of changes in #2501
- Update browser rollup configuration to reduce bundle size, minimising the increase due to refactoring modules to ESM

## Testing

Covered by existing CI test suite


packages/browser/rollup.config.npm.mjs
```diff
@@ -15,11 +15,27 @@ const sharedOutput = {
   strict: false, // 'use strict' in WebKit enables Tail Call Optimization, which breaks stack trace handling
 }
 
+const treeshake = {
+  preset: 'smallest', // More aggressive than 'safest'
+  propertyReadSideEffects: false,
+  unknownGlobalSideEffects: false,
+  // Be more aggressive with module side effects
+  moduleSideEffects: false,
```
gingerbenw, line 23: Check if we are actually side effect free-  I don't think we are

packages/browser/rollup.config.npm.mjs
```diff
@@ -15,11 +15,27 @@ const sharedOutput = {
...
+  // Enable more aggressive tree shaking for annotations
+  annotations: true
```
gingerbenw, line 25: we aren't using any rollup annotations, so this can go

PR #2561, opened 2025-09-17
Electron grouping discriminator

gingerbenw:
## Goal

Adds support for grouping discriminator in Electron applications to enable additional event grouping capabilities across main and renderer processes.

- Implements setGroupingDiscriminator and getGroupingDiscriminator methods for both main and renderer processes
- Adds IPC communication to synchronize grouping discriminator values between processes
- Extends native client state persistence to store grouping discriminator data


yousif-bugsnag (approved):
LGTM

PR #2605, opened 2025-11-10
Pin yoctocolors-cjs to 2.1.2

gingerbenw:
## Goal

Resolve failing build due to `yoctocolors-cjs` dependency updating


twometresteve:
Change looks fine, but which branches of the build should be released to be confident that it all works at run time?

hamudi-ahmed (approved):

gingerbenw:
@twometresteve that'll be the browser build - but it's only a dev dependency so shouldn't make any difference. I've just triggered the build though!

PR #2626, opened 2025-12-04
Replace "the test should run in this browser" steps with Cucumber tags

twometresteve:
## Goal

Avoid running scenarios at all if the browser does not support the features needed, reducing the time taken for the whole test run and number of automation flakes.

## Design

This change moves away from dynamically determining if a browser supports the JS features needed for the test, to a knowledge-base approach.  Information (from caniuse.com) on browser support for all the features needed has been captured in a new Browser` class - as we already do in bugsnag-js-performance.  New `@requires_....` tags use this information to skip scenarios completely when the browser does not support the features needed.  This speeds up the whole test run and reduces test flakes (before we're doing less automation).

## Changeset

- New `Browser` class, acting as the knowledge base for browser version support of various features.
- Several new "requires" tags added for the skipping of scenarios not supported by the current browser version.
- Remove `the test should run in this browser` Cucumber step.  
- Remove `bugsnag-test-should-run` element from all test fixtures. 

## Testing

Covered by CI with the browser sub-pipeline unblocked.

### DIFF
--- a/test/browser/features/auto_detect_errors.feature
+++ b/test/browser/features/auto_detect_errors.feature
@@ -1,30 +1,30 @@
 @auto_detect_errors
 Feature: Switching off automatic reporting
 
-Scenario: setting autoDetectErrors option to false
-  When I navigate to the test URL "/auto_detect_errors/script/autodetect_false.html"
-  Then I wait to receive an error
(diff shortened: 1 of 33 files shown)


gingerbenw (approved):
Love this!

PR #2639, opened 2025-12-15
Create http errors plugin

gingerbenw:
## Goal

This PR implements a new plugin to automatically capture and report HTTP errors from web applications. The plugin monitors HTTP requests (both fetch and XHR) and creates structured error reports when responses return configured error status codes.

## Design

- Records comprehensive request/response metadata including:
  - Request: URL, method, headers, body, query parameters
  - Response: status code, headers, body (XHR only)
- Integrates with `@bugsnag/request-tracker` to intercept HTTP requests and responses
- Redacts sensitive data using `redactedKeys` configuration

## Changeset

- Create new http-errors plugin
- Add Response data to events in core library

## Testing

- Add unit tests for new functionality
- Add end to end tests for fetch and xhr requests


packages/core/types/common.d.ts
```diff
@@ -120,6 +120,13 @@ interface Request {
   [key: string]: any
 }
 
+interface Response {
+  statusCode: number
```
hamudi-ahmed, line 124: Should this be `statusCode` or `status`? I think the ED says `status`, but I'm not sure which one is correct

gingerbenw: The PD says `statusCode` so I went with that but happy to defer to @tomlongridge

packages/request-tracker/lib/xhr-tracker.js
```diff
@@ -63,6 +77,7 @@ function createXhrTracker (global, options = {}) {
           onRequestEnd({
             endTime: Date.now(),
             state: 'error',
+            headers: this.getAllResponseHeaders(),
             xhr: this
```
hamudi-ahmed: It looks like we're passing the `xhr` object back just to get the response body in `plugin-http-errors`. I think it would be better to have a `body` property on the `endContext` and set that here instead of passing the `xhr` instance to `onRequestEnd`. The `startContext` and `endContext` should be a consistent interface for both xhr and fetch.

packages/plugin-http-errors/http-errors.js
```diff
@@ -0,0 +1,194 @@
...
+          if (initialBody) {
+            const bodyStr = String(initialBody)
+            requestBody = truncate(initialBody, maxRequestSize)
+            requestBodyLength = bodyStr.length // Report the original length before truncating
+          }
+
+          // Extract response headers
+          let responseHeaders = {}
+          if (endContext.headers) {
+            responseHeaders = xhrResponseHeadersToObject(endContext.headers) // XHR case
```
hamudi-ahmed: Similar to the xhr response body, I think this should be handled in the request tracker code, so that the headers in `endContext` are in the same format interface for both xhr and fetch.

packages/request-tracker/lib/fetch-tracker.js
hamudi-ahmed: I don't think we should pass the actual response object to `onRequestEnd`. We should pass back a plain object with the relevant properties e.g. headers extracted into a common format

packages/plugin-http-errors/http-errors.js
```diff
@@ -0,0 +1,194 @@
...
+
+        try {
+          // Extract request information
+          const url = startContext.url
+          const method = startContext.method
+          const domain = extractDomain(url)
+
+          // Redact query parameters in URL
+          const redactedUrl = redactQueryParameters(url, redactedKeys)
+          const redactedQueryParams = parseQueryParams(redactedUrl)
```
hamudi-ahmed: Should the redaction take place before or after the `onHttpError` callback?

packages/plugin-http-errors/lib/headers-to-object.js
```diff
@@ -0,0 +1,20 @@
+/**
+   * Convert Headers object to plain object
+   * @param {Headers} headers - Headers object
+   * @returns {Object} Plain object with header key-value pairs
+   */
+module.exports = function (headers) {
+  if (!headers) return {}
+
+  const obj = {}
+  if (headers.entries) {
```
gingerbenw:
```suggestion
  if (typeof headers.entries === 'function') {
```

packages/request-tracker/lib/xhr-header-string-to-object.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function (headersString) {
```
hamudi-ahmed, line 7: this feels like it should be done in the request tracker

packages/plugin-http-errors/http-errors.js
gingerbenw: that's a great question, actually. @tomlongridge what do you think?

tomlongridge: Afterwards, so that callbacks can potentially take action with the raw data. I believe that's the order for a normal event?

packages/request-tracker/lib/xhr-tracker.js
```diff
@@ -27,6 +29,18 @@ function createXhrTracker (global, options = {}) {
       originalOpen.apply(this, arguments)
     }
 
+    global.XMLHttpRequest.prototype.setRequestHeader = function setRequestHeader (header, value) {
+      // it's possible for `this` to be `undefined`, which is not a valid key for a WeakMap
+      if (this) {
+        const requestData = trackedRequests.get(this)
+        if (requestData) {
+          requestData.headers = requestData.headers || {}
+          requestData.headers[String(header)] = String(value)
```
hamudi-ahmed: each time you call setRequestHeader() after the first time you call it, the value is actually appended to the end of the existing header value, but here we're overwriting it

gingerbenw: I've now updated this to reflect the actual behaviour of `setRequestHeader` - it doesn't appear there's a way to clear it afterwards using the xhr api, so I've done the same here.

hamudi-ahmed (approved):
LGTM

PR #2640, opened 2025-12-15
Create new request-tracker package

gingerbenw:
## Goal

This PR creates a new @bugsnag/request-tracker package that provides shared HTTP request tracking functionality for Bugsnag JavaScript plugins.

## Design

The package implements a singleton pattern to track fetch and XMLHttpRequest requests, allowing multiple plugins to register callbacks for the same requests without duplicating instrumentation.

## Changeset

- Introduces a `RequestTracker` class that manages callbacks from multiple plugins
- Implements `createFetchTracker` and `createXhrTracker` factory functions with singleton patterns
- Adds URL filtering utilities and duration calculation helpers

## Testing

Add unit tests for new package

### DIFF
--- a/packages/request-tracker/lib/request-tracker.js
+++ b/packages/request-tracker/lib/request-tracker.js
@@ -5,7 +5,6 @@
 class RequestTracker {
   constructor () {
     this.callbacks = []
-    this._isActive = false
   }
 
   /**
@@ -57,25 +56,10 @@ class RequestTracker {
   }
 
   /**
-   * Mark tracker as active (used by tracker factories)
-   */
-  _setActive () {
-    this._isActive = true
-  }
-
-  /**
-   * Check if tracker has been activated
-   */
-  isActive () {
-    return this._isActive
-  }
-
-  /**
    * Reset tracker (for testing)
    */
   _reset () {
     this.callbacks = []
-    this._isActive = false
   }
 }
 
--- a/packages/request-tracker/lib/request-tracker.js
+++ b/packages/request-tracker/lib/request-tracker.js
@@ -1,0 +1,82 @@
+/**
+ * Singleton RequestTracker class for managing HTTP request instrumentation
+ * Allows multiple plugins to register callbacks for the same requests
+ */
+class RequestTracker {
+  constructor () {
+    this.callbacks = []
+    this._isActive = false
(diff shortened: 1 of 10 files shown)


packages/request-tracker/lib/request-tracker.js
```diff
@@ -0,0 +1,82 @@
...
+   * Mark tracker as active (used by tracker factories)
+   */
+  _setActive () {
+    this._isActive = true
+  }
+
+  /**
+   * Check if tracker has been activated
+   */
+  isActive () {
```
hamudi-ahmed: I can't quite see what the active state is for?

packages/request-tracker/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "dependencies": {
+    "@bugsnag/core": "^8.0.0"
+  },
+  "devDependencies": {
+    "jest": "^29.0.0"
+  },
+  "jest": {
+    "testEnvironment": "jsdom"
+  }
```
hamudi-ahmed: Is there a reason for adding a separate jest dependency here?

packages/request-tracker/lib/request-tracker.js
gingerbenw: I thought that would be coming in a later PR... but turns out it's not. looks like it's unnecessary now.

packages/request-tracker/package.json
gingerbenw: I'll blame copilot for that one.

packages/request-tracker/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "url": "https://github.com/bugsnag/bugsnag-js.git"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "author": "Bugsnag",
+  "license": "MIT",
+  "scripts": {
+    "test": "npm run test:unit",
+    "test:unit": "jest --testPathPattern=test/unit"
```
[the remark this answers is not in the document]
hamudi-ahmed: If we're okay to use the root jest dependency then these scripts can just be removed too - unit tests are run from the repo root anyway

hamudi-ahmed (approved):
LGTM

PR #2642, opened 2025-12-15
(plugin-server-session): delay initialization until first use

hamudi-ahmed:
## Goal

Serverless node plugins such as AWS Lambda (and the incoming Cloudflare Workers plugin) use `plugin-browser-session` and not `plugin-server-session`, but because `plugin-server-session` is bundled with the node notifier as an internal plugin, it's session tracking is still initialized on `start`, before any of the external plugins are loaded.

This causes problems in some environments like Cloudflare Workers, and results in unnecessary work and leaked timers, since the session delegate is replaced when those plugins are loaded.

## Design

This PR fixes the issue by delaying the session tracker initialization until `startSession` is actually called - this preserves the existing behaviour for 'regular' node environments while allowing the serverless plugins to load `plugin-browser-session` before any initialization happens. 

## Testing

Covered by existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+(plugin-server-session) Delay session tracker initialization until first use [#2642](https://github.com/bugsnag/bugsnag-js/pull/2642)
+
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


gingerbenw (dismissed):

gingerbenw:
changelog!

gingerbenw (approved):

PR #2645, opened 2025-12-18
Use shared tracker in network breadcrumbs package

gingerbenw:
## Goal

This PR refactors the network-breadcrumbs plugin to use a shared request tracker implementation instead of maintaining its own monkey-patching logic.

## Testing

Covered by existing unit and end to end tests


hamudi-ahmed (approved):
LGTM

PR #2647, opened 2026-01-05
[Integration] HTTP Errors

gingerbenw:
## Goal

This PR implements HTTP error tracking functionality by introducing a new `@bugsnag/plugin-network-instrumentation` plugin and a shared `@bugsnag/request-tracker` package. The plugin automatically captures and reports HTTP errors (4xx/5xx status codes) from both fetch and XMLHttpRequest APIs, with configurable error code ranges, request/response body size limits, and customizable filtering via callbacks.

## Changeset

- New request tracker package providing singleton-based instrumentation for fetch and XHR
- HTTP errors plugin that leverages the request tracker to detect and report error status codes
- Integration with existing network breadcrumbs plugin to use shared request tracking infrastructure


packages/plugin-network-instrumentation/lib/should-capture-status-code.js
```diff
@@ -0,0 +1,17 @@
+/**
+   * Check if a status code should be captured
+   * @param {number[]} codes - Array of HTTP status codes
+   * @param {number} statusCode - HTTP status code
+   * @returns {boolean} True if should be captured
+   */
```
lemnik: nit: inconsistent doc comment formatting

```suggestion
 * Check if a status code should be captured
 * @param {number[]} codes - Array of HTTP status codes
 * @param {number} statusCode - HTTP status code
 * @returns {boolean} True if should be captured
 */
```

packages/plugin-network-instrumentation/lib/truncate.js
```diff
@@ -0,0 +1,10 @@
+/**
+   * Truncate string to max length
+   * @param {string} str - String to truncate
+   * @param {number} maxLength - Maximum length
+   * @returns {string} Truncated string
+   */
```
lemnik: nit: inconsistent formatting

```suggestion
/**
 * Truncate string to max length
 * @param {string} str - String to truncate
 * @param {number} maxLength - Maximum length
 * @returns {string} Truncated string
 */
```

packages/plugin-network-instrumentation/types/bugsnag-plugin-network-instrumentation.d.ts
```diff
@@ -0,0 +1,43 @@
+import { Plugin, Request, Response } from '@bugsnag/core'
+
+export interface HttpErrorRange {
+  min: number
+  max: number
+}
```
lemnik, line 10: Is it worth documenting the exact behaviour of these values here as well? most specifically that `max` is inclusive?

packages/request-tracker/lib/headers-to-object.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = function (headers) {
+  if (!headers) return {}
+
+  const obj = {}
+  if (typeof headers.entries === 'function') {
+    var iterator = headers.entries()
+    var entry = iterator.next()
+    while (!entry.done) {
+      var key = entry.value[0]
+      var value = entry.value[1]
```
lemnik:
```suggestion
    const iterator = headers.entries()
    let entry = iterator.next()
    while (!entry.done) {
      const [key, value] = entry.value
```

packages/plugin-network-instrumentation/types/bugsnag-plugin-network-instrumentation.d.ts
gingerbenw: updated, and I'll ensure the docs pages include this information

hamudi-ahmed (approved):

PR #2648, opened 2026-01-05
Http errors suggestions and fixes

gingerbenw:
## Goal

This PR applies suggestions and fixes related to HTTP error handling, primarily removing deprecated IE8 compatibility options from webpack configurations and improving URL handling for modern browsers.

- Removes `ie8: true` option from UglifyJsPlugin configurations across multiple webpack config files
- Improves URL parsing logic to handle both absolute and relative URLs without relying on a placeholder base URL
- Refactors iterator usage to be ES5-compatible by avoiding destructuring syntax

### DIFF
--- a/packages/plugin-http-errors/lib/redact-query-parameters.js
+++ b/packages/plugin-http-errors/lib/redact-query-parameters.js
@@ -2,9 +2,10 @@ const redactValues = require('./redact-values')
 
 function isAbsoluteURL (url) {
   try {
-    URL.parse ? URL.parse(url) : new URL(url)
+    // eslint-disable-next-line no-new
+    new URL(url)
     return true
-  } catch {
+  } catch (e) {
     return false
   }
 }
--- a/packages/plugin-http-errors/lib/redact-query-parameters.js
+++ b/packages/plugin-http-errors/lib/redact-query-parameters.js
@@ -1,8 +1,20 @@
 const redactValues = require('./redact-values')
-const defaultBaseURL = 'http://invalid-base.com'
+
+function isAbsoluteURL (url) {
+  try {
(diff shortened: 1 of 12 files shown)


packages/plugin-http-errors/lib/redact-query-parameters.js
```diff
@@ -13,8 +26,13 @@ module.exports = function (url, redactedKeys) {
...
   urlObj.search = new URLSearchParams(redactedParams).toString()
-  const urlString = decodeURI(urlObj.toString())
-  return urlString.startsWith(defaultBaseURL)
-    ? urlString.slice(defaultBaseURL.length)
-    : urlString
+
+  // Return appropriate format based on original URL type
+  if (isAbsolute) {
+    return decodeURI(urlObj.toString())
+  }
```
hamudi-ahmed, line 33: `decodeURI` and `new URL` both can throw an exception - Is there any chance of an invalid URL being passed to this? If so is it worth wrapping all of this in a try/catch?

gingerbenw: that's a good shout - what would you propose the behaviour is if it errors? log an error and don't notify the request?

gingerbenw: Update: looks like we already handle this - the call to `redactQueryParameters` is already wrapped in a try/catch which logs `Failed to process HTTP error: err.message`

hamudi-ahmed: ah okay 👍🏾

hamudi-ahmed (approved):

PR #2650, opened 2026-01-05
Fix React Navigation tests and dependencies

hamudi-ahmed:
## Goal

Previously we were only running React Navigation tests on old architecture, but the dependencies were still being pulled in because we still include the plugin and scenarios in the new arch fixture builds. This has started causing CI failures as new versions of these dependencies are incompatible with older React Native versions.

## Design

Simply removing the plugin and scenarios for new arch builds isn't a great idea in the long term as new React Native versions are new arch only, so this PR re-works how we handle react navigation testing by:

- Explicitly installing compatible versions of React Navigation and associated dependencies in the test fixture based on the React Native version
- Introducing a `REACT_NAVIGATION` environment variable - this controls whether the plugin, scenarios and dependencies are included during fixture builds, and also whether or not to skip the scenarios during a test run.
- Re-enables the react navigation tests for new architecture, so that we continue to test as we add new react native versions
- Disables the tests for 0.72 New Arch (using the new environment variable) as this is the only problematic version/arch combination 

## Testing

Covered by a full CI run


gingerbenw (approved):
Much nicer 👌🏻 thanks!

PR #2667, opened 2026-01-30
Handle request and response in delivery-react-native

gingerbenw:
## Goal

Handle `request` and `response` in `@bugsnag/delivery-react-native` and ensure the values are correct on both iOS and Android

This PR also refactors the URL parsing in `plugin-network-instrumentation` to be more performant by separating out the domain, url and query string in a single pass rather than processing the string in multiple stages

## Testing

Update E2E tests to check `bodyLength` and `statusCode` are integers
Update unit tests for url parsing functionality


hamudi-ahmed (approved):
LGTM

PR #2684, opened 2026-02-19
fix(plugin-network-instrumentation): omit stacktrace from HTTP Error events

hamudi-ahmed:
## Goal

As per the design, HTTP Error events should not include a stacktrace

## Design
`plugin-network-instrumentation` now creates an event with an empty stacktrace

## Testing

Added unit and e2e test assertions to check the event stacktrace is empty

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,41 +4,42 @@
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


lemnik (approved):
LGTM

PR #2686, opened 2026-02-27
Report error cause with native stacktrace for Turbo Module exceptions

hamudi-ahmed:
## Goal

In The New Architecture, when an unhandled exception occurs in a Turbo Module method, React Native throws a JS error with a `cause` property containing the native exception class, error message and stacktrace.

Although the JS SDK supports error causes, these native causes are not parsed correctly in events as they don't contain valid JS stacktraces (the cause stack is reported as `stackElements` on Android and `stackSymbols`/`stackReturnAddresses` on iOS) - so we end up reporting events with an empty stacktrace in the 'Caused by' error.

This PR fixes the error reporting for these cases, so that native 'Caused by' errors are reported correctly with a full native stacktrace while maintaining backwards compatibility with native stack handling for Old Architecture.

(Note: on iOS, exceptions in async Turbo Module methods result in a native crash that is detected by the Cocoa SDK, so this does not apply in those cases)

## Changeset

- Updated the native stack handling in `delivery-react-native` to extract the `nativeStack` property for native error causes.

- Updated `BugsnagEventDeserializer.m` in the iOS native module to correctly handle events containing multiple errors (the equivalent changes for Android were released in v6.22.0 of `bugsnag-plugin-react-native`)

## Testing

Added new unit test and e2e test cases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,16 +5,17 @@
 ### Changed
 
 - (plugin-network-instrumentation) Manually parse URLs to improve React Native compatibility [#2674](https://github.com/bugsnag/bugsnag-js/pull/2674)
-- Update bugsnag-android to [v6.22.0](https//github.com/bugsnag/bugsnag-android/releases/tag/v6.22.0) [#2656](https://github.com/bugsnag/bugsnag-js/pull/2656)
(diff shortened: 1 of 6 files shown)


lemnik (approved):
LGTM

PR #2690, opened 2026-03-03
Update verdaccio config and React Native development guide

hamudi-ahmed:
## Goal

- Updates the verdaccio config to ensure that the external `@bugsnag/cli` package is proxied correctly
- Moves the verdaccio config file from `test/electron` up to the `test` directory as it's useful generally and not just for electron testing
- Updates the development guide for React Native to use the repo's verdaccio setup

## Testing

Tested manually


packages/react-native/CONTRIBUTING.md
```diff
@@ -37,46 +37,43 @@ To solve this problem we publish to a local npm clone, which proxies requests fo
...
-#### Installing the development notifier on a React Native project
+#### Installing the development notifier in a React Native project
 
 1. Make changes.
-2. Run the following command to publish to the local registry:
+2. In a new terminal window, from the repo root, run the `local-npm:publish-all` npm script to publish to the local registry:
 
     ```
-    npx lerna publish v99.99.99-canary.`git rev-parse HEAD` --no-push --exact --no-git-tag-version --registry http://localhost:4873/
+    VERSION_IDENTIFIER=8.99.99 npm run local-npm:publish-all
```
gingerbenw, line 64: Assuming we're still using a specific version identifier here that's always higher than the actual published version? is it worth noting that, or do you think it's safe to assume that's understood?

gingerbenw (approved):
Much cleaner 👌🏻

packages/react-native/CONTRIBUTING.md
hamudi-ahmed: Yeah, I think that should (hopefully) be obvious!

PR #2729, opened 2026-04-15
fix axios vulnerability issue

SB-priyankap:
## Goal

fix axios vulnerability issue

## Changeset

update package.json

## Testing

perform unit testing


SB-gulshans (approved):
Changes look good to me

PR #2743, opened 2026-05-04
bug fix unbreak build on glibc >= 2.28

Juraci:
## Goal

`@bugsnag/plugin-electron-client-state-persistence` fails to compile on any system with glibc ≥ 2.28 (Ubuntu 18.10+, Debian 10+, Fedora 29+, Arch, etc.). The vendored `tinycthread` redefines `once_flag` and `call_once` against names that glibc now exposes through `<stdlib.h>`, producing `conflicting types for 'pthread_once_t'` and `conflicting types for 'pthread_once'` errors during `node-gyp rebuild`. This breaks `npm install` for any consumer of `@bugsnag/electron` on a current Linux distro that doesn't have a prebuilt binding cached.

## Design

The plugin uses **only** `mtx_t`, `mtx_init`, `mtx_lock`, `mtx_unlock`, and `mtx_plain` from tinycthread (see `src/bugsnag_electron_client_state_persistence.c`). `once_flag`, `ONCE_FLAG_INIT`, and `call_once` are never referenced on POSIX — the only `call_once` definition in `tinycthread.c` is gated behind `#if defined(_TTHREAD_WIN32_)`.

Since the colliding code is dead on POSIX, the smallest and safest fix is to drop the POSIX `#else` branches entirely. The Win32 `typedef struct once_flag` and `void call_once(...)` declarations are untouched, preserving Windows behaviour.

A more durable fix would be to drop the vendored tinycthread on platforms that ship C11 `<threads.h>` and use the system header. That's a larger change; happy to do it as a follow-up if reviewers prefer.

## Changeset

`packages/plugin-electron-client-state-persistence/src/deps/tinycthread/tinycthread.h`: deletes the two POSIX `#else` branches (5 lines total) that redefine `once_flag` / `ONCE_FLAG_INIT` / `call_once`. No other files modified.

```diff
@@ -456,9 +456,6 @@ int tss_set(tss_t key, void *val);
     CRITICAL_SECTION lock;
   } once_flag;
   #define ONCE_FLAG_INIT {0,}
-#else
-  #define once_flag pthread_once_t
-  #define ONCE_FLAG_INIT PTHREAD_ONCE_INIT
 #endif

 /** Invoke a callback exactly once
@@ -468,8 +465,6 @@ int tss_set(tss_t key, void *val);
  */
 #if defined(_TTHREAD_WIN32_)
   void call_once(once_flag *flag, void (*func)(void));
-#else
-  #define call_once(flag,func) pthread_once(flag,func)
 #endif
```

## Testing

Manual:
- [x] `npm install && npx node-gyp rebuild` in `packages/plugin-electron-client-state-persistence` succeeds on Arch Linux (glibc 2.43); produces `build/Release/bugsnag_pecsp_bindings.node`. Without this patch, the same command fails with `conflicting types for 'pthread_once_t'`.
- [x] CI to confirm Linux / macOS / Windows builds remain green (Win32 branch is untouched; POSIX-side macros were dead code, so no behavioural change is expected).

Automated:
- No new tests added — the change deletes dead-code macro definitions on POSIX. Existing CI build steps for the plugin exercise the regression surface.

| Platform | Before | After |
|---|---|---|
| Linux glibc ≥ 2.28 | build fails | builds |
| Linux glibc < 2.28 | builds | builds (macros removed but unused) |
| macOS | builds | builds (macros removed but unused) |
| Windows | builds | builds (Win32 branch untouched) |

### DIFF
--- a/packages/plugin-electron-client-state-persistence/src/deps/tinycthread/tinycthread.h
+++ b/packages/plugin-electron-client-state-persistence/src/deps/tinycthread/tinycthread.h
@@ -456,9 +456,6 @@ int tss_set(tss_t key, void *val);
     CRITICAL_SECTION lock;
   } once_flag;
   #define ONCE_FLAG_INIT {0,}
-#else
-  #define once_flag pthread_once_t
-  #define ONCE_FLAG_INIT PTHREAD_ONCE_INIT
 #endif
 
 /** Invoke a callback exactly once
@@ -468,8 +465,6 @@ int tss_set(tss_t key, void *val);
  */
 #if defined(_TTHREAD_WIN32_)
   void call_once(once_flag *flag, void (*func)(void));
-#else
-  #define call_once(flag,func) pthread_once(flag,func)
 #endif
 
 #ifdef __cplusplus


hannah-smartbear:
Hi @Juraci,

Thanks for raising this PR. Our engineers will take a look, and any comments or updates will be posted here accordingly.

SB-shivamgarg:
Hi @Juraci
We are considering your changes in this [PR #2769](https://github.com/bugsnag/bugsnag-js/pull/2769)
Thanks!

PR #2753, opened 2026-05-25
plat-16462 js to typescript conversion for node plugin

SB-jigneshR:
## Goal

PNIP JS to TS conversion

## Changeset

PNIP JS to TS conversion and did relevant changes

## Testing

Did UT and end to end testing


SB-gulshans (changes requested):
Update script section of package.json
Update tsconfig for outdir and other parameters

SB-gulshans (approved):
LGTM

PR #2764, opened 2026-06-08
Convert plugin server session js to ts

SB-priyankap:
## Goal

Convert plugin server session js to ts

## Changeset

Convert plugin server session js to ts and relevant changes

## Testing

Did UT and end to end testing


packages/plugin-server-session/src/session.ts
```diff
@@ -0,0 +1,128 @@
...
+      client._logger.debug(String(sessionCount) + ' session(s) reported')
+      return
+    }
+
+    if (backoff.attempts === MAX_ATTEMPTS) {
+      client._logger.error('Session delivery failed, max retries exceeded', err)
+      return
+    }
+
+    client._logger.error(
```
SB-rohandhiman, line 95: Previously, we used client._logger.debug in the JS file. Why was this changed?
client._logger.debug('Session delivery failed, retry #' + (backoff.attempts + 1) + '/' + maxAttempts, err)

packages/plugin-server-session/src/session.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  if (!sessionCounts.length) return
+
+  const backoff = new Backoff({ min: 1000, max: 10000 })
+  req(handleRes)
+
+  function handleRes (err?: Error | null): void {
+    if (!err) {
+      const sessionCount = sessionCounts.reduce((accum, s) => accum + s.sessionsStarted, 0)
+      client._logger.debug(String(sessionCount) + ' session(s) reported')
```
SB-rohandhiman: We were already using template literals in the JS code; please use the same approach here.

client._logger.debug(`${sessionCount} session(s) reported`)

packages/plugin-server-session/src/session.ts
```diff
@@ -0,0 +1,128 @@
...
+    sessionTracker.on('summary', sendSessionSummary(client))
+    sessionTracker.start()
+
+    client._sessionDelegate = {
+      startSession: (sessionClient, session) => {
+        sessionClient._session = session
+        sessionClient._pausedSession = null
+        sessionTracker.track(sessionClient._session)
+        return sessionClient
+      },
```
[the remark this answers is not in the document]
SB-rohandhiman: The Copilot suggestion is correct. The interval now starts even when startSession() is never called. Previously, it started only when needed, avoiding unnecessary background work when sessions weren't being used. This changes the original behavior.

packages/plugin-server-session/src/session.ts
```diff
@@ -0,0 +1,128 @@
...
+        releaseStage: client._config.releaseStage,
+        version: client._config.appVersion,
+        type: client._config.appType
+      },
+      sessionCounts
+    }
+
+    const ignore = runSyncCallbacks(
+      client._cbs.sp,
+      payload as unknown as Session,
```
SB-rohandhiman, line 118: This suggests that the runSyncCallbacks type in @bugsnag/core doesn't match what's being passed here. We should either add a comment explaining why, or update the core types to support SessionPayload properly. or // <-- workaround

packages/plugin-server-session/src/type/backo.d.ts
```diff
@@ -0,0 +1,19 @@
+declare module 'backo' {
```
SB-rohandhiman, line 1: Since these are types, can we create a /types folder and move this file there?

packages/plugin-server-session/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_SUMMARY_INTERVAL = 10000;
```
SB-rohandhiman, line 1: Since constants.ts has been removed, can we either move DEFAULT_SUMMARY_INTERVAL to tracker.ts and MAX_ATTEMPTS to session.ts where they are used, or create a utils folder and keep the shared constants there

packages/plugin-server-session/src/tracker.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (this._interval !== null) {
+      clearInterval(this._interval)
+    }
+    this._interval = null
+  }
+
+  track (session: Session): Session {
+    const key = dateToMsKey(session.startedAt)
+    const cur = this._sessions.get(key)
+    this._sessions.set(key, cur === null || cur === undefined ? 1 : cur + 1)
```
SB-rohandhiman: nullish coalescing is cleaner
this._sessions.set(key, (cur ?? 0) + 1)

SB-gulshans (approved):

PR #2767, opened 2026-06-10
Pri/plat 16526 plugin restify js to ts

SB-priyankap:
## Goal

Convert plugin testify JS to TS

## Changeset

Convert plugin testify JS to TS

## Testing

Did UT and end to end testing


SB-gulshans (approved):
LGTM

PR #2768, opened 2026-06-10
Plugin stack frame path normaliser js to ts conversion

SB-priyankap:
## Goal

Plugin stack frame path normaliser js to ts conversion

## Changeset

Plugin stack frame path normaliser js to ts conversion and relevant changes

## Testing

Did UT and end to end testing


SB-gulshans (approved):
LGTM

PR #2781, opened 2026-06-21
sca fixes for hono and others

SB-gulshans:
## Goal

sca fixes for hono and other plugins

## Changeset

sca fixes for hono and other plugins

## Testing

Unit test cases


SB-priyankap:
Fix this in PR #2787

PR #2788, opened 2026-07-02
sca findings react native fix

SB-gulshans:
## Goal

sca findings react native fix

## Changeset

sca findings react native fix

## Testing

unit test cases


SB-priyankap:
Fix in other PR #2787

PR #2741, opened 2026-04-28
Update react-native-navigation version to v8

SB-priyankap:
## Goal

Update react-native-navigation version to v8

## Changeset

changes in yml file and script file

## Testing

did end to end testing and run unit test cases

PR #2735, opened 2026-04-22
await next in hono error handler middleware

gingerbenw:
## Goal

Updates the Hono plugin’s error-handling middleware to properly await next() and adds/adjusts integration tests and fixtures to cover async behavior, non-Error throws, and request-body parsing.

## Changeset

- Update `@bugsnag/plugin-hono` error handler to await next() and handle thrown non-Error values.

## Testing

- Extend Node maze-runner steps/features to allow POSTing with an explicit Content-Type (including JSON).
- Adjust AWS Lambda Hono feature coverage and upload maze output artifacts on workflow failure.

PR #2730, opened 2026-04-16
fix deprication warning issue for parse

SB-priyankap:
## Goal

Fix deprecation warning for parse(url);

## Changeset

update with new URL(url) inside request.js

## Testing

did unit testing

PR #2789, opened 2026-07-03
Release/v8.10.0

SB-priyankap:
**Added**

(react-native-navigation) Add support for react-native-navigation v8 https://github.com/bugsnag/bugsnag-js/pull/2741
Requires React Native 0.78+ (New Architecture only)
Peer dependency range updated: react-native-navigation 2.0.0 - 8.x

**Fixed**

(delivery-node) Replace deprecated `url.parse()` with `URL` constructor https://github.com/bugsnag/bugsnag-js/pull/2730
(plugin-hono) Fix issue where error handler middleware did not `await next()` https://github.com/bugsnag/bugsnag-js/pull/2735
(react-native) Fix issue related to uploading source map https://github.com/bugsnag/bugsnag-js/pull/2773)


packages/delivery-node/request.js
```diff
@@ -11,14 +9,14 @@ module.exports = ({ url, headers, body, agent }, cb) => {
     cb(err)
   }
 
-  const parsedUrl = parse(url)
+  const parsedUrl = new URL(url)
```
SB-jigneshR, line 12: Can we have check for parsedUrl

packages/delivery-node/request.js
```diff
@@ -11,14 +9,14 @@ module.exports = ({ url, headers, body, agent }, cb) => {
...
   const secure = parsedUrl.protocol === 'https:'
   const transport = secure ? https : http
   const req = transport.request({
     method: 'POST',
     hostname: parsedUrl.hostname,
     port: parsedUrl.port,
```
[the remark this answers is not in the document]
SB-priyankap: [parsedUrl.port || undefined](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is acceptable but unnecessary. When no port is present, [URL.port](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is '', and Node’s request handling treats that effectively the same as undefined here. So it does not fix a bug; it just normalizes the value

the best choice is going with the original code
port: parsedUrl.port

packages/delivery-node/request.js
SB-priyankap: parsedUrl.port || undefined is acceptable but unnecessary. When no port is present, URL.port is '', and Node’s request handling treats that effectively the same as undefined here. So it does not fix a bug; it just normalizes the value

the best choice is going with the original code
port: parsedUrl.port

.buildkite/basic/react-native-ios-full-pipeline.yml
```diff
@@ -87,30 +87,6 @@ steps:
             - exit_status: "*"
               limit: 1
 
-      - label: ':mac: Build react-native-navigation {{matrix}} test fixture ipa (Old Arch)'
```
SB-jigneshR, line 90: @SB-priyankap : As we are introducing support for the new react-native-navigation plugin, could you please confirm that removing these test cases will not impact the functionality of the existing architecture?

SB-priyankap: It will not affect the existing functionality of a JS error SDK,
We have removed the old architecture feature test case as the v8 version of React-Native-Navigation only supports new architecture.

SB-gulshans (approved):

SB-jigneshR (approved):

PR #2800, opened 2026-07-17
done with SCA finding fixed

SB-priyankap:
## Goal

SCA finding fixes

## Changeset

Done changes in package.json file and also Restore TypeScript strictness (strict, noImplicitAny, strictNullChecks)

## Testing

Did UT

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "bugsnag-js",
+  "name": "app",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "app",
+  "name": "bugsnag-js",
```
SB-jigneshR: Specific reason to change name ??

SB-priyankap: Automatically changes this name I have revert back to original one

SB-jigneshR (approved):

